PR #2, opened 2019-10-24
Fix lens code generation and add basic test

jereksel:


compiler-plugin/src/main/kotlin/arrow/meta/plugins/optics/LensPlugin.kt
```diff
@@ -34,7 +34,7 @@ val Meta.lenses: Plugin
             newDeclaration =
             if (c.companionObjects.isEmpty())
               """|
-                 |$modality $visibility data $kind $name($`(valueParameters)`) {
```
nomisRev: Was this causing the stack overflow?

jereksel: No this is another error - it causes class to have signature `class Name(constructor (parameters))` - constructor cannot be in parenthesis.

SO was caused by https://github.com/arrow-kt/arrow-meta/pull/2/files#diff-b8d55b69b6b0f14dba8cd95119be2bc4R23

nomisRev: This code will be updated later because it's a prototype rather than an actual implementation.
This implementation will get a more user-friendly API which will be build while I work on documenting the quote system and further develop that public API.

jereksel:
I don't have write rights to this repository.

rachelcarmena:
Hi @jereksel! Thanks for this PR :tada:  Sorry, we changed the testing DSL and `build` check is failing now because of the test which is added here. Please, let us know if we can help you to adapt the test with the new DSL.

nomisRev:
@jereksel oh right. Sorry, I mixed up with Arrow repo 😅

Anyone with write access can merge once the build is green!

Thanks for the PR 👏

PR #3, opened 2019-10-25
Testing: new DSL

rachelcarmena:
* New DSL
* Fix: to be able to find the package descriptor when `datatype` is in a `.class` file.
* Small format changes


compiler-plugin/src/test/kotlin/arrow/meta/plugins/typeclasses/TypeClassesTest.kt
```diff
@@ -0,0 +1,83 @@
...
+    | import arrow.extension
+    | import arrow.core.ForOption
+    | import arrow.core.fix
+    | import arrow.core.None
+    |
+    | //metadebug
+    |
+    | @extension
+    | object OptionMappable : Mappable<ForOption> {
+    |   override fun <A, B> Kind<ForOption, A>.map(f: (A) -> B): Kind<ForOption, B> =
```
nomisRev: @raulraja do we currently support `Option<A>` here instead already with the type checker?

If not should we continue using `OptionOf` instead? I've found this to be annoying to fix and maintain throughout the Arrow codebase.

compiler-plugin/src/test/kotlin/arrow/meta/plugins/typeclasses/TypeClassesTest.kt
```diff
@@ -0,0 +1,83 @@
...
+    |       is Some -> Some(f(o.t))
+    |       None -> None
+    |     }
+    | } 
+    | 
+    | interface Mappable<F> {
+    |   fun <A, B> Kind<F, A>.map(f: (A) -> B): Kind<F, B>
+    | }
+    |
+    | object Test {
```
nomisRev: Do we need to use a wrapping `object` class for testing reasons?

rachelcarmena:
> Love the testing DSL!! heart_eyes
> Can I start writing more failing tests now? smiling_imp

Thanks @nomisRev !! I hope it's better now. New version! :wink:

compiler-plugin/src/test/kotlin/arrow/meta/plugins/typeclasses/TypeClassesTest.kt
rachelcarmena: No, that `Test` is part of the code snippet under test. Maybe the use of `Test` is confusing if it's inside the code snippet... Good point, thanks! :+1:

testing-plugin/src/test/kotlin/arrow/meta/plugin/testing/ExampleTest.kt
```diff
@@ -2,47 +2,118 @@ package arrow.meta.plugin.testing
...
+        | fun hello(): String =
+        |   "Hello world!"
+        | 
+        """.source
+      },
+      assert = {
+        listOf(compiles)
+      }
+    )
+    compilerTest.run(interpreter)
```
raulraja: We should hide running from the user since all these tests are expected to produce Unit

testing-plugin/src/test/kotlin/arrow/meta/plugin/testing/ExampleTest.kt
```diff
@@ -2,47 +2,118 @@ package arrow.meta.plugin.testing
...
+  fun `check that emits an error diagnostic when compilation fails`() {
+    val compilerTest = CompilerTest(
+      code = {
+        """
+        | classsss Error
+        | 
+        """.source
+      },
+      assert = {
+        listOf(
```
raulraja: Is there a case with more than one assert?. Most assertion are simple so there should be an additional function for `CompilerTest { }` that does not require returning a list of asserts

testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/CompilationAssertions.kt
```diff
@@ -16,35 +13,65 @@ private data class ExpressionParts(
...
+    is Assert.EvalsTo -> assertEvalsTo(compilationResult, assert.source, assert.output)
   }
-  return Result(classesDirectory = compilationResult.classesDirectory)
+
+  val configuration = it.config(CompilerTest)
+  val assertions = it.assert(CompilerTest)
+
+  configuration.map(::runConfig)
+
+  val compilerPlugins = configuration.filterIsInstance<Config.Many>().flatMap { it.configs }.filterIsInstance<Config.AddCompilerPlugins>().flatMap { it.plugins }.flatMap { it.dependencies }.map { it.mavenCoordinates }
```
raulraja: Let's pair on this part, this should be easier than this by folding or recursing over the cases.

testing-plugin/src/test/kotlin/arrow/meta/plugin/testing/ExampleTest.kt
rachelcarmena: Now there is just one case: `quoteOutputMatches` + `evalsTo`. 

Maybe we could create 2 tests in that case. I'm going to try it.

testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/CompilationAssertions.kt
rachelcarmena: Thanks @raulraja ! :+1:

testing-plugin/src/test/kotlin/arrow/meta/plugin/testing/ExampleTest.kt
rachelcarmena: Done! I recovered the previous `assertThis`. I think it makes sense now:

* There is an assert in the test.
* The user could be able to change the default interpreter.

Thanks for the review @raulraja !

testing-plugin/src/test/kotlin/arrow/meta/plugin/testing/ExampleTest.kt
rachelcarmena: Changed!

PR #4, opened 2019-10-25
`buildMeta` & `publishAndRunIde` build tasks

raulraja:
Ports the build tasks from the old location and rewrites them to refer to the local module tasks.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,3 +22,18 @@ allprojects {
     jcenter()
   }
 }
+
+task buildMeta {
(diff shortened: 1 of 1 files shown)


rachelcarmena:
Thanks @raulraja ! I'm going to update CI with those tasks!

raulraja:
these are mostly for local dev building with fresh starts since the CI task already runs all tests which is what we want there.

rachelcarmena:
Ok! :+1:

PR #6, opened 2019-10-25
Add base test and simple example how to use it

jansorg:
- Adds abstract class `KotlinIdeBaseTest`, to be extended later on. The check for the resolved is disabled for now to make the test pass. This can be fixed as soon as we have heavy tests.
- Adds `IconProviderTest`, which servers as a first example how to use it
- Configures logging for tests only. `DEBUG`, `INFO`, `WARN`, `ERROR` messages of channel `#arrow` will be logged to stdout when running locally or on CI


idea-plugin/src/test/resources/log4j.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE log4j:configuration SYSTEM "log4j.dtd">
+
+<log4j:configuration xmlns:log4j="http://jakarta.apache.org/log4j/">
+    <appender name="console" class="org.apache.log4j.ConsoleAppender">
+        <param name="Target" value="System.out"/>
+        <layout class="org.apache.log4j.PatternLayout">
+            <param name="ConversionPattern" value="%-5p %c{1} - %m%n"/>
+        </layout>
```
i-walker: What does this append Log files to CI?

jansorg: it still logs to stdout. In my experience (with Travis) `stdout` is available for each job afterwards, but I'm not sure which CI provider you're using

i-walker: Yes perfect we're using Travis

PR #7, opened 2019-10-25
Test Arrow LineMarkerProviders

jansorg:
This adds a test to test Arrow's line marker providers.

Changes:
- Adds test for `@higherKind`
- Fix LineMarkerProvider of `@higherKind` to provide markers in the way, which is recommended by JetBrains (see notes in the code)

To do:
- atm only the `@higherKind` provider is tested, add testing of the other implementations, too
- because of https://github.com/arrow-kt/arrow-meta/issues/10 multiple markers will be displayed when you open several project windows. Each window triggers registration of another, identical set of line marker providers, which are then used in all projects.


jansorg:
CI is run, but there's not output from gradle for lot of tasks, including `test` of `idea-plugin`.
https://github.com/arrow-kt/arrow-meta/runs/274859948#step:4:276

idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/plugins/higherkinds/HigherKindsIdePlugin.kt
```diff
@@ -5,16 +5,20 @@ import arrow.meta.invoke
...
 val IdeMetaPlugin.higherKindsIdePlugin: Plugin
   get() = "HigherKindsIdePlugin" {
     meta(
-      addLineMarkerProvider(
+      addFastLineMarkerProvider(
         icon = ArrowIcons.HKT,
         message = "HKT",
-        matchOn = { it.safeAs<KtClass>()?.let(::isHigherKindedType) == true }
+        matchOn = {
+          KtTokens.IDENTIFIER == it.node?.elementType
```
i-walker: @jansorg Why was the parent-child check necessary?

rachelcarmena:
> CI is run, but there's not output from gradle for lot of tasks, including `test` of `idea-plugin`.
> https://github.com/arrow-kt/arrow-meta/runs/274859948#step:4:276

@jansorg , thanks for the info because we didn't realize about it. And it also happens with `compiler-plugin:test`. It would be necessary to add this configuration in their `build.gradle` files:

```
test {
  testLogging.showStandardStreams = true
}
```

We're going to add it. Thanks again!

i-walker:
Thank you @rachelcarmena, much appreciated.

idea-plugin/src/main/kotlin/arrow/meta/dsl/ide/editor/lineMarker/LineMarkerSyntax.kt
```diff
@@ -13,6 +13,22 @@ import com.intellij.psi.PsiElement
 import javax.swing.Icon
 
 interface LineMarkerSyntax {
+  /**
+   * Adds a fast line provider
+   * See [com.intellij.codeInsight.daemon.LineMarkerProvider] for notes how to implement a fast provider.
+   */
+  fun IdeMetaPlugin.addFastLineMarkerProvider(
```
i-walker: This doesn't add any LineMarkers to the Editor.

jansorg: It does, at least here. It's not displaying icons during indexing because our providers are not dumb-aware yet. Implementing the interface `PossiblyDumbAware` is probably best here as it allows to control the value.

jansorg: https://github.com/JetBrains/intellij-community/blob/192/platform/lang-impl/src/com/intellij/codeInsight/daemon/impl/LineMarkersPass.java#L93 is where the collectors are collected, `getMarkerProviders` filters by dumb-awareness.

idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/plugins/higherkinds/HigherKindsIdePlugin.kt
jansorg: see https://github.com/JetBrains/intellij-community/blob/192/platform/lang-api/src/com/intellij/codeInsight/daemon/LineMarkerProvider.java#L22. 
We're making sure that we're only adding the marker for the class name identifier, not the whole `KtClass`.
A `KtClass` covers all of this `@higherkind class Name {...}`, the identifier leaf element is just `Name` here. See the javadoc above for a few more details on this.

idea-plugin/src/main/kotlin/arrow/meta/dsl/ide/editor/lineMarker/LineMarkerSyntax.kt
i-walker: I see your point now. We should have the "fast" option global for everyone. Let me redefine that.

Thanks for looking into it :)

raulraja:
@jansorg we can test all other plugins in further Prs if you want to get this in so other people can also help test with the DSL for comprehensions etc. It's up to you.

idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/plugins/higherkinds/HigherKindsIdePlugin.kt
nomisRev: What does difference imply for the user? I understand your point but I am curious to understand the use-case being an IDEA user.

i-walker: AFAIK they'll be situations, where the Icons blink annoyingly.
I am sure we can find workarounds to prevent that, but we don't have any User data.
In this PR we discuss it more in-depth https://github.com/arrow-kt/arrow-meta/pull/12

jansorg:
@i-walker cc @raulraja In my opinion it's best to merge this now and to add tests for the other line marker providers in separate pull requests. 
@i-walker It's probably best if you merge this PR when you think it fits with the other changes you're working on.

i-walker:
I will merge this PR after that one https://github.com/arrow-kt/arrow-meta/pull/12 today

i-walker:
Thank you very much @jansorg. Now let's move forward with the IdeTest and HeavyTests next time :)

PR #12, opened 2019-10-26
LinemarkerSyntax update

i-walker:
This aligns the former `match` functions to the Quote system.
Furthermore, it subscribes LineMarkers to one PsiElement leave. 
The recommendation from IntelliJ is to use the PsiIdentifier, but not all `Targets` have that property and we don't need to traverse AST nodes, because of their isomorphism to PsiElements.

The Goal is to abstract finding the next smallest unit, beginning with a composite PsiElement, which the user passes.

Thanks to @jrgonzalezg here https://github.com/arrow-kt/arrow-meta/pull/7/files#r339240595

- [x] Add HTML support


idea-plugin/src/main/kotlin/arrow/meta/dsl/ide/editor/lineMarker/LineMarkerSyntax.kt
```diff
@@ -17,38 +18,42 @@ interface LineMarkerSyntax {
...
     message: (element: A) -> String = Noop.string1(),
     placed: GutterIconRenderer.Alignment = GutterIconRenderer.Alignment.RIGHT
   ): ExtensionPhase =
     addLineMarkerProvider(
-      matchOn,
-      { a ->
-        lineMarkerInfo(icon, a, message as (PsiElement) -> String, placed)
+      transform,
+      {
+        it.children.firstOrDefault(it).let { p -> lineMarkerInfo(icon, p, message as (PsiElement) -> String, placed) }
```
jansorg: Why is `firstOrDefault()` called here? If this is to get a leaf element, then it's not correct. 
A `PsiElement` is either a leaf (i.e. no child elements) or a composite element with `PsiElements` as child elements. Each of these children can now be leads or composite elements.

Here's the subtree of a `KtClass` (Tools->View Psi structure of current file...)
Screenshot_20191026_172031

The identifier is a leaf, but not necessarily the first. And there will be other provider implementations which don't want to limit the marker for the first child.

And, in general, a line marker provider should always add the marker to the `PsiElement` passed in by IntelliJ and not to a child element. IntelliJ keeps a cache of markers for `PsiElements` and you'll probably see unexpected behaviour with this implementation.
I'll add a general note, too.

idea-plugin/src/main/kotlin/arrow/meta/dsl/ide/editor/lineMarker/LineMarkerSyntax.kt
```diff
@@ -17,38 +18,42 @@ interface LineMarkerSyntax {
...
     extensionProvider(
       LineMarkerProviders.INSTANCE,
       object : LineMarkerProvider {
         override fun getLineMarkerInfo(element: PsiElement): LineMarkerInfo<PsiElement>? =
-          matchOn(element)?.let(lineMarkerInfo)
+          transform(element)?.let(lineMarkerInfo)
 
         override fun collectSlowLineMarkers(elements: MutableList<PsiElement>, result: MutableCollection<LineMarkerInfo<PsiElement>>) {
-          for (element: PsiElement in elements.filter { IdeUtils.isNotNull(matchOn(it)) }) {
+          for (element: PsiElement in elements.filter { IdeUtils.isNotNull(transform(it)) }) {
```
jansorg: `transform` is called multiple times, once in `filter` and once in the body of the `for` loop. If the passed function takes a long time, then it'll be slower than necessary.

idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/plugins/comprehensions/comprehensionsIdePlugin.kt
```diff
@@ -15,7 +15,7 @@ val IdeMetaPlugin.comprehensionsIdePlugin: Plugin
     meta(
       addLineMarkerProvider(
         icon = ArrowIcons.BIND,
-        matchOn = { it.safeAs<KtExpression>()?.takeIf(KtExpression::isBinding) },
+        transform = { it.safeAs<KtExpression>()?.takeIf(KtExpression::isBinding) },
```
jansorg: With the current implementation this adds a line marker for the first child of a class.
- this child is not necessarily a leaf
- this child is not necessarily on the same line as the class name identifier (e.g. with annotations)

idea-plugin/src/main/kotlin/arrow/meta/dsl/ide/editor/lineMarker/LineMarkerSyntax.kt
```diff
@@ -17,38 +18,42 @@ interface LineMarkerSyntax {
...
-      { a ->
-        lineMarkerInfo(icon, a, message as (PsiElement) -> String, placed)
+      transform,
+      {
+        it.children.firstOrDefault(it).let { p -> lineMarkerInfo(icon, p, message as (PsiElement) -> String, placed) }
       }
     )
 
+  /**
+   * It is advised to create LineMarkerInfo for leave elements and not entire ParentElements
```
jansorg: nit: leave -> leaf

jansorg: perhaps use `composite PsiElements` instead of `ParentElements`?

idea-plugin/src/main/kotlin/arrow/meta/dsl/ide/editor/lineMarker/LineMarkerSyntax.kt
i-walker: It makes sense to filter them before passing them to the `slowLineMarker`, but that's just a minor performance improvement. It doesn't really affect the overall performance.

idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/plugins/higherkinds/HigherKindsIdePlugin.kt
```diff
@@ -20,7 +20,7 @@ val IdeMetaPlugin.higherKindsIdePlugin: Plugin
     meta(
       addLineMarkerProvider(
         icon = ArrowIcons.HKT,
-        matchOn = { it.safeAs<KtClass>()?.takeIf(::isHigherKindedType) },
+        transform = { it.safeAs<KtClass>()?.takeIf(::isHigherKindedType) },
```
jansorg: With the current implementation this adds a line marker for the first child of a class.
- this child is not necessarily a leaf
- this child is not necessarily on the same line as the class name identifier (e.g. with annotations)

idea-plugin/src/main/kotlin/arrow/meta/dsl/ide/editor/lineMarker/LineMarkerSyntax.kt
i-walker: I agree with a few points of you, but the emphasis for this implementation is not to get the PsiIdentifier, because not all PsiElements, users will match on have that Property.

Thus we can simplify IntelliJ's suggestion to any leave element that a user passes. 
BTW the behavior stays the same. But we can verify that later with tests.
I am open for suggestions on how to get the deepest leave. I am trying out different things

jansorg: Could you explain what you mean by `the emphasis for this implementation is not to get the PsiIdentifier, because not all PsiElements, users will match on have that Property.`? My understanding is that the PsiIdentifier will be used for higherkind line marker, for example. Other line marker providers would annotate other leaf element.

What I mean is that the caller of `addLineMarkerProvider` has to traverse the PSI tree [here](https://github.com/arrow-kt/arrow-meta/blob/7bb2d0ccdb91928312c1d7f7b2f3bd9e53810048/idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/plugins/higherkinds/HigherKindsIdePlugin.kt#L23). My other PR has an example [here](https://github.com/arrow-kt/arrow-meta/blob/7cf6a6f22913853dd889f79f3da4de1e498bdeb7/idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/plugins/higherkinds/HigherKindsIdePlugin.kt#L19).

With the current implementation you're getting this:
Screenshot_20191026_185325
Here `@kotlin.Deprecated` is annotated. That's probably not correct (I assume) and `@kotlin.Deprecated` isn't a leaf element, too.

My understanding is that you'd like to put the line marker next to the class name.

To get the leaf element of a PsiElement you could use `PsiTreeUtil.getDeepestFirst()`, but that will only work as expected with a Psi structure where each parent contains only one child.
For the first example in the screenshot this would return the token `@` and the token `/**` for the 2nd example.

i-walker: Yes I am aware of that I thought you meant with behavior, that it doesn't subscribe to the right component. 
Makes sense. But the overall goal is not to pass a PsiIdentifier in `transform`. 
As you do in your PR. 

I am currently implementing something which looks for the deepest leave.

jansorg: (Rephrased my comments above, apologies if it sounded too direct or unfriendly.)
`PsiTreeUtil.getDeepestFirst` will return the leaf at the leftmost position in a PsiElement, `getDeepestLast` will return the rightmost element (referring to the position in the editor).
`getDeepestFirst` might be what's needed here for the annotation.

i-walker: Thank you I will try both of them out. 
No worries. It happens to me accidentally, too. May it's whenever we think in German and speak in English  :slightly_smiling_face:

i-walker: Maybe we could do somthing like: 
Give me the PsiIdentifier if that doesnt exist subscribe to the `Annotation`.
Would that satisfy the IntelliJ suggestion @jansorg ?

i-walker: The result is in the latest push.

idea-plugin/src/main/kotlin/arrow/meta/dsl/ide/editor/lineMarker/LineMarkerSyntax.kt
```diff
@@ -11,44 +11,51 @@ import com.intellij.openapi.actionSystem.AnAction
...
     message: (element: A) -> String = Noop.string1(),
     placed: GutterIconRenderer.Alignment = GutterIconRenderer.Alignment.RIGHT
   ): ExtensionPhase =
     addLineMarkerProvider(
-      matchOn,
-      { a ->
-        lineMarkerInfo(icon, a, message as (PsiElement) -> String, placed)
+      transform,
+      {
+        lineMarkerInfo(icon, (it as? PsiNameIdentifierOwner)?.identifyingElement
```
jansorg: Great idea to use the identifying element! That's much more generic.

Perhaps add a note in the comment or rename the method to tell a developer that it's automatically choosing a leaf for the result of `transform()`?

I took another look at the IntelliJ SDK. It seems to be assuming that a line marker is returned for the `PsiElement` which was passed to the line marker provider. But the relation `PsiElement <-> LineMarker` isn't used outside of the marker itself, as far as I can tell. 
So this implementation should work fine. LineMarkers created here might do unexpected things in the future when JetBrains changes the internals. 
2019.3 EAP is still following the same logic here as 2019.2.

jansorg:
@i-walker Sorry for all the noise.
If I hover over the line marker icon added for a `@higherkind` class, then an exception is raised.

```
java.lang.ClassCastException: class com.intellij.psi.impl.source.tree.LeafPsiElement cannot be cast to class org.jetbrains.kotlin.psi.KtClass (com.intellij.psi.impl.source.tree.LeafPsiElement is in unnamed module of loader com.intellij.util.lang.UrlClassLoader @6f1fba17; org.jetbrains.kotlin.psi.KtClass is in unnamed module of loader com.intellij.ide.plugins.cl.PluginClassLoader @4371c935)
	at arrow.meta.plugin.idea.plugins.higherkinds.HigherKindsIdePluginKt$higherKindsIdePlugin$1$2.invoke(HigherKindsIdePlugin.kt)
	at arrow.meta.dsl.ide.editor.lineMarker.LineMarkerSyntax$sam$com_intellij_util_Function$0.fun(LineMarkerSyntax.kt)
	at com.intellij.codeInsight.daemon.LineMarkerInfo.getLineMarkerTooltip(LineMarkerInfo.java:146)
	at com.intellij.codeInsight.daemon.LineMarkerInfo$LineMarkerGutterIconRenderer.getTooltipText(LineMarkerInfo.java:198)
	at com.intellij.openapi.editor.impl.EditorGutterComponentImpl$2.lambda$run$0(EditorGutterComponentImpl.java:1588)
	at com.intellij.openapi.application.impl.ApplicationImpl.runReadAction(ApplicationImpl.java:936)
	at com.intellij.openapi.application.ReadAction.compute(ReadAction.java:57)
	at com.intellij.openapi.editor.impl.EditorGutterComponentImpl$2.run(EditorGutterComponentImpl.java:1588)
	at com.intellij.openapi.progress.impl.CoreProgressManager$TaskRunnable.run(CoreProgressManager.java:894)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$runProcess$2(CoreProgressManager.java:169)
	at com.intellij.openapi.progress.impl.CoreProgressManager.registerIndicatorAndRun(CoreProgressManager.java:591)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeProcessUnderProgress(CoreProgressManager.java:537)
	at com.intellij.openapi.progress.impl.ProgressManagerImpl.executeProcessUnderProgress(ProgressManagerImpl.java:59)
	at com.intellij.openapi.progress.impl.CoreProgressManager.runProcess(CoreProgressManager.java:156)
	at com.intellij.openapi.progress.impl.CoreProgressManager$4.run(CoreProgressManager.java:408)
	at com.intellij.openapi.application.impl.ApplicationImpl$1.run(ApplicationImpl.java:294)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

idea-plugin/src/main/kotlin/arrow/meta/dsl/ide/editor/lineMarker/LineMarkerSyntax.kt
i-walker: Yes, the User shouldn't bother with these implementation details. I am still refining the DSL so we can merge your tests without any changes on User code. I saw the Errors, too. I think it tries to associate the passed element to the composite. Which generally leads to `ClassCastExceptions`. 
KtTokens work, but constructors seem to bail.

Thank you for checking in with me. This is all a step by step improvement.

i-walker:
I agree with your points @jansorg . Nonetheless, we should keep the algebra flexible for both styles. 
We will encourage users to align with IntilliJ's suggestions, but in some cases, the user needs the composite. As you see here https://github.com/arrow-kt/arrow-meta/blob/5c83567bdd0a621341e2cfc1fdcb7c862258a59b/idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/plugins/higherkinds/HigherKindsIdePlugin.kt#L34

With the recent changes, we can monitor, which Subtype of `PsiIdentifyingOwner` works in all cases. For now, I found 2, but once our `LightTests` are up and running we can add more.

idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/plugins/higherkinds/HigherKindsIdePlugin.kt
```diff
@@ -20,32 +30,28 @@ val IdeMetaPlugin.higherKindsIdePlugin: Plugin
...
-                   |${validExtensionLocations()}
-                   |
-                   |<a href="https://arrow-kt.io/docs/patterns/polymorphic_programs/">Learn more about type class and ad-hoc polymorphism</a>
-                   |""".trimMargin()
+            val compose: List<Node> = h1(name) + kotlin(this@run) +
+              code(name) + text(" is a polymorphic function that can be applied over $target.") +
+              code(name) + text(" requires @extension proofs for ") + code(givenParameters()) + text("Valid ") +
+              code("@extension") + text("locations:\n${validExtensionLocations()}") +
+              a(href = "https://arrow-kt.io/docs/patterns/polymorphic_programs/") { "Learn more about type class and ad-hoc polymorphism" }
+            html(compose).render()
```
i-walker: @raulraja I added Html support. This renders like this for :
`compose.render()`
Screenshot 2019-10-27 at 12 34 17

or as in line `54`
Screenshot 2019-10-27 at 12 36 48

We should be able to add GoTo Actions to the Web, so you dont have to add the link at the end. And the User simply clicks on the Icon.

i-walker: Please specify, which style works for you rn :)

raulraja: Either style works and user should be able to manually generate html or use the jetbrains library equally. I for one prefer string and htnl over dsls but I think many would appreciate the HTML based DSL

i-walker:
Thank you @jansorg I will change it appropriately

PR #13, opened 2019-10-26
Add development quick start instructions

jereksel:
Makes forking from https://github.com/arrow-kt/arrow/issues/1742 absolete.


README.md
```diff
@@ -52,18 +52,36 @@
...
+- compiler-plugin: Kotlin compiler plugin
+- gradle-plugin: Plugin for Gradle for activating Kotlin compiler plugin
+- idea-plugin: Plugin for Intellij IDEA for IDE integration (icons, autocomplete etc.)
+- testing-plugin: Tools for testing compiler and idea plugin
 
-## Credits
+Quick start development environment:
+
+- Run `publishAndRunIde` Gradle task
+- Import sample project (https://github.com/jereksel/arrow-meta-example) in new IDE instance
```
jereksel: Feel free to fork it

jereksel: Feel free to fork it

jereksel: https://github.com/arrow-kt/arrow-meta/pull/2

raulraja: @jereksel you have been invited :) feel free to port the sample project over to a module. Welcome to Meta!

raulraja:
@jereksel ready to go once conflicts are resolved, thanks!

build.gradle
```diff
@@ -21,15 +20,25 @@ allprojects {
...
+  dependsOn ':compiler-plugin:publishToMavenLocal'
   dependsOn ':gradle-plugin:publishArrowPluginMarkerMavenPublicationToMavenLocal'
   dependsOn ':gradle-plugin:publishPluginMavenPublicationToMavenLocal'
-  tasks.findByPath(':gradle-plugin:publishArrowPluginMarkerMavenPublicationToMavenLocal').mustRunAfter 'build'
-  tasks.findByPath(':gradle-plugin:publishPluginMavenPublicationToMavenLocal').mustRunAfter 'build'
+  dependsOn ':gradle-plugin:publishToMavenLocal'
+}
+
+task buildMeta {
+  dependsOn 'build'
```
i-walker: Can we add `clean`, too?

README.md
```diff
@@ -52,18 +52,36 @@
 
 ## Contributing
 
-## License
+Arrow-meta project is split into 4 packages:
+- compiler-plugin: Kotlin compiler plugin
+- gradle-plugin: Plugin for Gradle for activating Kotlin compiler plugin
+- idea-plugin: Plugin for Intellij IDEA for IDE integration (icons, autocomplete etc.)
+- testing-plugin: Tools for testing compiler and idea plugin
```
i-walker: Currently, the dsl to create idea plugin tests is in `idea-plugin`.
But is also not merged into master, yet 😄. 
We heavily rely dependencies which we already get from the existing set up in the `idea-plugin`.

compiler-plugin/build.gradle
```diff
@@ -41,6 +44,9 @@ shadowJar {
     configurations = [project.configurations.compile]
     relocate 'org.jetbrains.kotlin.com.intellij', 'com.intellij'
     relocate 'org.jetbrains.kotlin.load', 'kotlin.reflect.jvm.internal.impl.load'
+
+    //Required for idea-plugin tests
+    relocate 'org.apache.log4j', 'shadow.org.apache.log4j'
```
i-walker: Thank you 😄 🎉

i-walker: @jansorg Is that the correct dependency? For running it in CI

gradle-plugin/src/main/kotlin/arrow/meta/plugin/gradle/ArrowGradlePlugin.kt
```diff
@@ -20,10 +20,5 @@ class ArrowGradlePlugin : Plugin<Project> {
 
   override fun apply(project: Project): Unit {
     project.extensions.create("arrow", ArrowExtension::class.java)
```
i-walker: You should not delete that. Otherways we can not add the compiler-plugin to the repo. Once we deploy the compiler plugin we can change that to add the dependency. But as for now, we would keep it.

gradle-plugin/src/main/kotlin/arrow/meta/plugin/gradle/ArrowKotlinGradleSubplugin.kt
```diff
@@ -13,7 +13,7 @@ import org.jetbrains.kotlin.gradle.plugin.SubpluginOption
  */
 class ArrowKotlinGradleSubplugin : KotlinGradleSubplugin<AbstractCompile> {
   companion object {
-    private const val GRADLE_ARTIFACT_NAME = "gradle-plugin"
+    private const val KOTLIN_ARTIFACT_NAME = "compiler-plugin"
```
i-walker: You should not change the names here

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.0-bin.zip
-distributionPath=wrapper/dists
+#Sat Oct 26 17:44:20 CEST 2019
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.3-all.zip
```
i-walker: Can you revert those changes here?

gradle-plugin/build.gradle
```diff
@@ -6,16 +6,13 @@ plugins {
...
-project.group = 'io.arrow-kt'
-project.version = '0.0.1'
 def pluginId = "arrow.meta.plugin.gradle"
 
 dependencies {
   implementation gradleApi()
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
   implementation "org.jetbrains.kotlin:kotlin-gradle-plugin-api:$kotlin_version"
   implementation "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-  compileOnly project(":compiler-plugin") { "embeddable" }
```
i-walker: Why did you delete this? 😄

compiler-plugin/build.gradle
```diff
@@ -1,5 +1,8 @@
+plugins {
+    id 'java-gradle-plugin'
```
i-walker: Why do we need those plugins? :)

gradle-plugin/src/main/kotlin/arrow/meta/plugin/gradle/ArrowKotlinGradleSubplugin.kt
```diff
@@ -28,11 +28,19 @@ class ArrowKotlinGradleSubplugin : KotlinGradleSubplugin<AbstractCompile> {
     variantData: Any?,
     androidProjectHandler: Any?,
     kotlinCompilation: KotlinCompilation<KotlinCommonOptions>?
-  ): List<SubpluginOption> = emptyList()
+  ): List<SubpluginOption> {
+
+    @Suppress("UNUSED_VARIABLE")
+    val extension = project.extensions.findByType(ArrowExtension::class.java)
```
i-walker: I can't really tell from the code for what this is used. Can you help me, please ? @jereksel

idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/gradle/MetaClasspathContributor.kt
```diff
@@ -9,6 +9,11 @@ import java.io.File
 import java.io.FileInputStream
 import java.util.jar.JarInputStream
 
+/**
+ * Source:
+ * https://github.com/fabianishere/kotlin-plugin-generated/commit/19fbbe9d222126b33cfe90b604473ba7e7ea371b#diff-2f347766958584e8e50469cc6bab68c9
+ * https://github.com/JetBrains/kotlin/blob/822b455/plugins/kotlin-serialization/kotlin-serialization-ide/src/org/jetbrains/kotlinx/serialization/idea/KotlinSerializationImportHandler.kt
```
i-walker: This link seems to be dead?

build.gradle
jereksel: I don't like `clean`, because it makes build much longer. I understand using it in specific times like changing branch, or updating gradle - but not during every build.

gradle-plugin/build.gradle
jereksel: Gradle plugin does not need compiler-plugin embedded - it's downloaded from repo

gradle-plugin/src/main/kotlin/arrow/meta/plugin/gradle/ArrowKotlinGradleSubplugin.kt
jereksel: For me `GRADLE_ARTIFACT_NAME` means gradle plugin, not kotlin compiler plugin that it's supposed to be

gradle-plugin/src/main/kotlin/arrow/meta/plugin/gradle/ArrowKotlinGradleSubplugin.kt
jereksel: When we will be adding options for gradle plugin we must pass them to compiler plugin. It's done in this method

idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/gradle/MetaClasspathContributor.kt
jereksel: Both works fine here

compiler-plugin/build.gradle
jereksel: Yeah, java-gradle-plugin is not needed, but maven-plugin is required for local maven deployment

gradle-plugin/src/main/kotlin/arrow/meta/plugin/gradle/ArrowGradlePlugin.kt
jereksel: Because of that to use plugin you must clone arrow-meta in specific directory. Gradle plugin has mechanism of automatically downloading and applying kotlin compiler plugin from maven repository.

gradle-plugin/src/main/kotlin/arrow/meta/plugin/gradle/ArrowKotlinGradleSubplugin.kt
i-walker: But there are no Options here now. So why add a variable?

gradle-plugin/src/main/kotlin/arrow/meta/plugin/gradle/ArrowKotlinGradleSubplugin.kt
i-walker: That is not the intent what you might think.
We supply tooling in the idea plugin to get the gradle-plugin, which does the registration automatically, based on rules we define in `ArrowGradlePlugin`. Because you deleted it it makes sense just to retrieve it directly. 
With the changes, you introduce your naming the gradle-plugin `compiler-plugin`, which doesn't make sense to me.

build.gradle
i-walker: CI needs it. I don't mind having it on your settings, but that introduces changes to our process.

gradle-plugin/src/main/kotlin/arrow/meta/plugin/gradle/ArrowGradlePlugin.kt
i-walker: Can you explain how that works. We did not have that experience that Gradle does that?

compiler-plugin/build.gradle
i-walker: Nop it is not.

compiler-plugin/build.gradle
jansorg: @i-walker The plugin is using the logging facility of the IntelliJ SDK.  IntelliJ is using log4j internally but the plugin itself isn't using log4j and log4j is in the classpath because of the dependency on the SDK.
Do you know why this is needed atm? Is there another log4j dependency which needs to be moved to work with the SDK?

gradle-plugin/src/main/kotlin/arrow/meta/plugin/gradle/ArrowGradlePlugin.kt
i-walker: @jereksel Thanks for the explanation in Slack. The problem was that we were pointing to the gradle-plugin itself, which lead to the configurations we had back then to add the compiler-plugin manually.

compiler-plugin/build.gradle
jereksel: I've removed dependency between Gradle and Kotlin compiler plugin. Maybe log4j was some transitive dependency.

In any case log4j that is used in tests is binary incompatible with one that is here.

gradle-plugin/src/main/kotlin/arrow/meta/plugin/gradle/ArrowKotlinGradleSubplugin.kt
jereksel: So it'll be easier to add options in the future. I don't want to remove something that will surely be used in the future.

raulraja: We need settings soon for all plugins and meta itself to enable and disable plugins via config without code changes so if this provides an entry path to it let's keep it.

README.md
```diff
@@ -258,18 +258,36 @@ This is intentional so the API is all about intercepting desired elements and tr
...
 
-## Credits
+Quick start development environment:
 
-**Build and run tests**
+- Run `publishAndRunIde` Gradle task
+- Import sample project in new IDE instance
```
jereksel: I'll add it in another PR

gradle-plugin/build.gradle
```diff
@@ -5,16 +5,13 @@ plugins {
 }
 apply plugin: 'kotlin'
 
-project.group = 'io.arrow-kt'
-project.version = '0.0.1'
```
rachelcarmena: Sorry @jereksel , I'm looking forward to merging this pull request because it has a lot of good things, thanks!, but I've seen these lines right now. I think it's necessary [a way to manage the group and version for Gradle plugin](https://plugins.gradle.org/docs/publish-plugin). As far as I know it cannot be the same process we are following so far. A Gradle plugin can only be published once with a specific version. However, snapshots can be published without limit. Thanks!

build.gradle
```diff
@@ -87,13 +87,20 @@ subprojects { project ->
   }
 }
 
+task deployMeta {
+  dependsOn ':compiler-plugin:publishToMavenLocal'
+  dependsOn ':gradle-plugin:publishArrowPluginMarkerMavenPublicationToMavenLocal'
+  dependsOn ':gradle-plugin:publishPluginMavenPublicationToMavenLocal'
+  dependsOn ':gradle-plugin:publishToMavenLocal'
+}
```
rachelcarmena: It seems this task could be `installMeta` because it's publishing to the local repository.

gradle-plugin/src/main/kotlin/arrow/meta/plugin/gradle/ArrowKotlinGradleSubplugin.kt
```diff
@@ -13,9 +13,9 @@ import org.jetbrains.kotlin.gradle.plugin.SubpluginOption
  */
 class ArrowKotlinGradleSubplugin : KotlinGradleSubplugin<AbstractCompile> {
   companion object {
-    private const val GRADLE_ARTIFACT_NAME = "gradle-plugin"
+    private const val META_ARTIFACT_NAME = "compiler-plugin"
```
rachelcarmena: I checked this change when preparing #69. However, it seems that both `gradle-plugin` and `compiler-plugin` don't provide an additional feature for including the `compiler-plugin` with `-Xplugin`.

PR #35, opened 2019-10-28
Flattens CompilationResult ADT

i-walker:
It seems that `FailsWith` is written as a Leaf of Assert rather than CompilationResult. 

Tell me if I am wrong :)


i-walker:
Your welcome :)

PR #37, opened 2019-10-28
Low level API docs

raulraja:
Most docs that will go in classes once @nomisRev setups the doc site with ank and some placeholders to insert videos and animations. @israelperezglez we need all videos and animations in gif format too for the markdown docs.


JorgeCastilloPrz:
Will those carry code snippets also? It reads well but at the same time feels a bit abstract and hard to get mental mapping with just text.

raulraja:
@JorgeCastilloPrz yes, we are just getting the structure in first and then split into apidocs, ilustrations etc.

JorgeCastilloPrz:
Yep @nomisRev, but even after moving those it'd make sense to add snippets or other more graphic examples.

PR #40, opened 2019-10-29
Configuration: publish artifacts

rachelcarmena:
## Changes
* Add **Github Actions** to publish artifacts:
  * `io/arrow-kt/arrow-meta-testing-plugin`
  * `io/arrow-kt/arrow-meta-idea-plugin`
  * `io/arrow-kt/arrow-meta-compiler-plugin`
  * `io/arrow-kt/arrow-meta-docs`
* **NOTE**. Old shell scripts are not used (`deploy-scripts`  directory). Just an additional command in Github Actions.
* Add Bintray configuration
* Use `gradle-mvn-push` to upload archives in the same way as Arrow
* Add `gradle.properties` for every artifact except for `gradle-plugin` because it has its own way to be published: #41 

## Configuration
* [Add secrets](https://help.github.com/en/github/automating-your-workflow-with-github-actions/virtual-environments-for-github-actions#creating-and-using-secrets-encrypted-variables): `BINTRAY_USER`, `BINTRAY_API_KEY`

### DIFF
--- a/idea-plugin/gradle.properties
+++ b/idea-plugin/gradle.properties
@@ -1,0 +1,4 @@
+# Maven publishing configuration
+POM_NAME=Arrow IDEA Plugin
+POM_ARTIFACT_ID=arrow-meta-idea-plugin
+POM_PACKAGING=jar
--- a/.github/workflows/publish_artifacts.yml
+++ b/.github/workflows/publish_artifacts.yml
@@ -1,0 +1,31 @@
+name: Publish artifacts
+
+on:
+  push:
+    branches:
+    - master
+
+jobs:
+  publish_artifacts:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+        architecture: x64
+    - name: Build with Gradle
+      env:
+        JAVA_OPTS: -Xms512m -Xmx1024m
+      run: |
+        ./gradlew clean build
+    - name: Publish artifacts
+      env:
+        BINTRAY_USER: ${{ secrets.BINTRAY_USER }}
+        BINTRAY_API_KEY: ${{ secrets.BINTRAY_API_KEY }}
+      run: |
+        ./gradlew uploadArchives
+        echo "$(cat gradle.properties | grep VERSION_NAME | cut -d'=' -f2) deployed!"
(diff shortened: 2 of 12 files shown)


gradle/gradle-mvn-push.gradle
```diff
@@ -0,0 +1,142 @@
+/*
```
raulraja: Is there a way in which we can have this and other files that are the same as the ones in arrow in a single place? This is gonna be an increasing issue with all other libs splitting out of the monorepo

idea-plugin/gradle.properties
```diff
@@ -0,0 +1,4 @@
+# Maven publishing configuration
+POM_NAME=Arrow Meta Idea Plugin
```
raulraja:
```suggestion
POM_NAME=Arrow IDEA Plugin
```

gradle/gradle-mvn-push.gradle
rachelcarmena: Thanks @raulraja for the idea, done! 
We cannot use the original repository because we adapted it for Arrow, so I changed it to use the file from Arrow repository. 
In case we have more repositories, we could move it to a single repository.

idea-plugin/gradle.properties
rachelcarmena: Done!

rachelcarmena:
Everything is ready to merge this pull request :tada: 
When merging in master, `publish_artifacts` from GitHub Actions will run and these artifacts will be published (timestamp will change in the name of these artifacts):

* https://oss.jfrog.org/artifactory/oss-snapshot-local/io/arrow-kt/arrow-meta-testing-plugin/0.10.3-SNAPSHOT/arrow-meta-testing-plugin-0.10.3-20191030.110905-1.jar

* https://oss.jfrog.org/artifactory/oss-snapshot-local/io/arrow-kt/arrow-meta-docs/0.10.3-SNAPSHOT/arrow-meta-docs-0.10.3-20191030.110905-1.jar

* https://oss.jfrog.org/artifactory/oss-snapshot-local/io/arrow-kt/arrow-meta-compiler-plugin/0.10.3-SNAPSHOT/arrow-meta-compiler-plugin-0.10.3-20191030.110908-1.jar

* https://oss.jfrog.org/artifactory/oss-snapshot-local/io/arrow-kt/arrow-meta-idea-plugin/0.10.3-SNAPSHOT/arrow-meta-idea-plugin-0.10.3-20191030.110905-1.jar

PR #43, opened 2019-10-30
Relocated docs + new typeChecker config function

raulraja:
Relocates docs in the README to the KDoc


compiler-plugin/src/main/kotlin/arrow/meta/quotes/Quote.kt
```diff
@@ -205,12 +231,6 @@ inline fun <reified K : KtElement, P : KtElement, S> processKtFile(
       ).process(element as K)
       transformation?.let { mutations.add(it) }
     }
-//    file.accept(object : MetaTreeVisitor() {
-//      override fun visitKtElement(element: KtElement) {
-//
```
i-walker: @raulraja is that the old technique or did we abstract that into another Function I could find it?

PR #53, opened 2019-11-01
Minor improv. on `testing-plugin` & public CompilationResult

i-walker:
In order for us to have `HeavyTests` in  the `ide`, which verify that `MetaSyntheticPackageFragementProvider` does its job and all PsiElements result in with valid references to their typeclass, we need the `outputdirectory` in `CompilationResult` `classesfiles`. I renamed it for coherence. Concluding that the `CompilationResult` DT is public, but our `interpreter` is still internal and not discoverable for users. 

I am open to Feedback, separating the compilation and the actual interpretation of that result seems also to be a good pattern. 
Contrary, if we don't make `CompilationResult` public we can't establish `HeavyTests` in testing. As there is no way for us to generate the metadata in the editor.


compiler-plugin/src/test/kotlin/arrow/meta/plugins/comprehensions/ComprehensionsTest.kt
```diff
@@ -49,7 +49,7 @@ class ComprehensionsTest {
         listOf(addCompilerPlugins(compilerPlugin))
       },
       code = {
-        codeSnippet.source
```
rachelcarmena: I'd keep it because of: avoiding the use of primitives types, better types and it's planned to have several sources.

testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/TestingLibraryWrapper.kt
```diff
@@ -10,11 +10,11 @@ import java.nio.file.Paths
 
 private const val DEFAULT_FILENAME = "Example.kt"
 
-internal data class CompilationResult(
+data class CompilationResult(
   val actualStatus: CompilationStatus,
   val log: String,
   val actualGeneratedFilePath: Path,
-  val classesDirectory: File
```
rachelcarmena: There are different directories and we only use classes directory. I think outputDirectory could be confusing with the outputDirectory of the testing library.

testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/TestingLibraryWrapper.kt
```diff
@@ -10,11 +10,11 @@ import java.nio.file.Paths
 
 private const val DEFAULT_FILENAME = "Example.kt"
 
-internal data class CompilationResult(
+data class CompilationResult(
```
rachelcarmena: Instead of doing CompilationResult public, I'd think about the desired assertion. The purpose of this library is to provide a simple testing framework. If we publish the compilationResult, it seems this library doesn't make sense and we might use the kotlin compiler testing library directly.

rachelcarmena:
@i-walker , despite of the comment, thanks for this pull request!! I think it could be useful to think about the code you'd like to be able to write. What's the desired assertion you would like to write? And then, we could think about how to change the library.

testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/TestingLibraryWrapper.kt
i-walker: where is the outputDirectory of the testing library.  I couldnt find it.

testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/TestingLibraryWrapper.kt
i-walker: I am not writing an assertion

i-walker: I need the `testing-library` with the wrappers we added in `testing-library`. For the `IDE` it's a utility.

compiler-plugin/src/test/kotlin/arrow/meta/plugins/comprehensions/ComprehensionsTest.kt
i-walker: May you reiterate on what you mean with multiple sources?

rachelcarmena:
Another idea, @i-walker: use directly kotlin compiler testing library and then we'll think about a way to move your need to this library. I hope it's useful for you :pray:

i-walker:
We can also focus on the more pressing issue, that other libraries won't be able to pull `testing-library`
```
{
  "errors" : [ {
    "status" : 409,
    "message" : "The repository 'oss-snapshot-local' rejected the resolution of an artifact 'oss-snapshot-local:com/github/arrow-kt/kotlin-compile-testing/1.3.0/kotlin-compile-testing-1.3.0.pom' due to conflict in the snapshot release handling policy."
  } ]
}
```

i-walker:
This is a blocker for us because we already tried workarounds and pulling the underlying library directory and copy-pasting the code from `testing-plugin` doesn't seem to be an elegant solution.

compiler-plugin/src/test/kotlin/arrow/meta/plugins/comprehensions/ComprehensionsTest.kt
rachelcarmena: I mean about different snippets.

testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/TestingLibraryWrapper.kt
rachelcarmena: line 33

rachelcarmena:
> We can also focus on the more pressing issue, that other libraries won't be able to pull `testing-library`
> 
> ```
> {
>   "errors" : [ {
>     "status" : 409,
>     "message" : "The repository 'oss-snapshot-local' rejected the resolution of an artifact 'oss-snapshot-local:com/github/arrow-kt/kotlin-compile-testing/1.3.0/kotlin-compile-testing-1.3.0.pom' due to conflict in the snapshot release handling policy."
>   } ]
> }
> ```

Sorry, I don't understand this error. We are not deploying that library in oss.jfrog.org. We're using JitPack to get that dependency.

testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/TestingLibraryWrapper.kt
rachelcarmena: Please, could you give more details?

i-walker:
Ok. How can we add testing-plugin to the dependencies in idea-plugin?

compiler-plugin/src/test/kotlin/arrow/meta/plugins/comprehensions/ComprehensionsTest.kt
i-walker: that should be isomorphic to `List<Source>`. What am indicating here is that your not doing anything to your DataType which is domain-specific other than renaming `String` to `Source`

testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/TestingLibraryWrapper.kt
i-walker: About what specifically?

testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/TestingLibraryWrapper.kt
i-walker: So, you're saying that `classesDirectory` is the same as `outputDirectory`?

rachelcarmena:
> Ok. How can we add testing-plugin to the dependencies in idea-plugin?

It's already included in `compiler-plugin` for compiling/executing tests: 
* `testImplementation project(":testing-plugin")`
* `testRuntimeOnly` dependencies for `compiler-plugin` and the rest of the libraries which are necessary to compile snippets in tests
* JitPack repository:
```
repositories {
    maven { url "https://jitpack.io" }
}
``` 
because testing-plugin is using the dependency of `kotlin-compiler-testing` which is only provided by JitPack right now.

cc: @jansorg

testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/TestingLibraryWrapper.kt
rachelcarmena: They are different. That's the reason why I chose `classesDirectory`...

testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/TestingLibraryWrapper.kt
rachelcarmena: About the use of the library in IDE...

i-walker:
Closing this

PR #55, opened 2019-11-01
Improve dependency resolution time

Octogonapus:
This PR improves dependency resolution time by reordering the declaration order of the repositories and adding a resolution filter for the arrow-kt snapshot dependencies. This reduced the dependency resolution time on my machine from 11 minutes and 28.960 seconds to 29.781 seconds (measured using Gradle build scans).

The wrapper changes are because I had to update the wrapper version to use the recent resolution filter (the `content` method).


Octogonapus:
The proof is in the pudding: this build took 4 minutes. Other builds seem to take around 16 minutes.

i-walker:
I love the pudding.

PR #57, opened 2019-11-01
[WIP] TypeClassResolutionTest for code snippets

jansorg:
@i-walker cc @raulraja 

This PR adds a very early version of a "heavy" test to test type resolution. A heavy test is a test which configures a full IntelliJ runtime environment and a project on disk before the test methods are run. This is more heavyweight than the light tests we're usually running in `idea-plugin`. 

**This is not ready to merge**

`arrow.meta.ide.plugins.typeclasses.TypeResolutionTest` is a wip test to demonstrate how the heavy testing is done. 
This needs to happen as soon as it's working:
- The code to setup the module content root, build folder and src folder will most probably be shared between different tests
- The method to compile a snippet should be [shared too](https://github.com/arrow-kt/arrow-meta/pull/53)

## What's broken
- running tests with IntelliJ is broken. The reason is that IntelliJ's gradle integration can't handle the shadow plugin. https://youtrack.jetbrains.com/issue/IDEA-163411. 
- Running test `TypeClassResolutionTest` currently produces an exception about a mismatch of class `org.jetbrains.kotlin.idea.KotlinFileType`.  Both the Kotlin IDE plugin and dependency `kotlin-compiler-embedded` contain this class.


jansorg:
@i-walker The heavy test base class `HeavyIdeTestSetUp` is inheriting from `CodeInsightFixtureTestCase`. It's also sharing a fixture and this might allow some reuse between light and heavy test logic.
`CodeInsightFixtureTestCase` is similar to class `PlatformTestCase`, which I mentioned earlier. It's setting up a fixture to simplify PsiFile creation, etc. Without a fixture setup we'd have to reimplement this in `PlatformTestCase`.

testing-plugin/build.gradle
```diff
@@ -15,11 +16,12 @@ dependencies {
 
     compileOnly "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
     compileOnly "io.github.classgraph:classgraph:$class_graph_version"
-    implementation "org.assertj:assertj-core:$assertj_version"
+    compile "org.assertj:assertj-core:$assertj_version"
```
rachelcarmena: Hi @jansorg , this dependency is used in runtime as well.

testing-plugin/build.gradle
```diff
@@ -15,11 +16,12 @@ dependencies {
...
 
     // TODO: temporary dependency until PR is approved and published
     // implementation "com.github.tschuchortdev:kotlin-compile-testing:$kotlin_compile_testing"
-    implementation "com.github.arrow-kt:kotlin-compile-testing:$kotlin_compile_testing"
+    // implementation "com.github.arrow-kt:kotlin-compile-testing:$kotlin_compile_testing"
+    compile "com.github.tschuchortdev:kotlin-compile-testing:1.2.4-SNAPSHOT"
```
rachelcarmena: This version doesn't exists: https://github.com/tschuchortdev/kotlin-compile-testing/releases
I think the only option would be updating `$kotlin_compile_testing` to `1.3.0` with the previous dependency: https://github.com/arrow-kt/kotlin-compile-testing/releases

And it's also used in runtime, so `compile` will raise errors.

i-walker:
Thanks to both of you I will do a complete cleanUp and close the PR https://github.com/arrow-kt/arrow-meta/pull/53

testing-plugin/build.gradle
tschuchortdev: It's a snapshot release so it's not listed under the regular releases. I was going to wait with the official 1.2.4 release until the PR for easier compiler plugin support is merged.

docs/build.gradle
```diff
@@ -2,6 +2,10 @@ apply plugin: 'ank-gradle-plugin'
 apply plugin: 'kotlin-kapt'
 //apply plugin: 'org.ajoberstar.git-publish'
 
+repositories {
+    maven { url "https://jitpack.io" }
+}
+
```
rachelcarmena: I think it's not necessary here...

testing-plugin/build.gradle
```diff
@@ -2,64 +2,63 @@ apply plugin: 'kotlin'
...
+    // TODO: temporary version until PR is approved and published
+    kotlin_compile_testing = '1.3.0'
+  }
 }
 
 dependencies {
 
-    compileOnly "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
-    compileOnly "io.github.classgraph:classgraph:$class_graph_version"
-    compileOnly project(":compiler-plugin")
```
rachelcarmena: This is necessary now because of the last commits

testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/TestingLibraryWrapper.kt
```diff
@@ -10,14 +10,14 @@ import java.nio.file.Paths
 
 private const val DEFAULT_FILENAME = "Example.kt"
 
-internal data class CompilationResult(
+data class CompilationResult(
   val actualStatus: CompilationStatus,
   val log: String,
   val actualGeneratedFilePath: Path,
-  val classesDirectory: File
```
rachelcarmena: There are other directories and this is only the classes directory. For instance:

```Paths.get(internalResult.outputDirectory.parent, "sources", "$DEFAULT_FILENAME.meta")```

is jumping to the parent in order to access to sources. I think if we talk about classes directory, the source code communicates better.

File: testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/CompilationAssertions.kt [hunk 6]
```diff
@@ -48,6 +32,21 @@ private val interpreter: (CompilerTest) -> Unit = {
...
+      when (config) {
+        is Config.AddCompilerPlugins -> remaining.compilationData(acc.addCompilerPlugins(config))
+        is Config.AddDependencies -> remaining.compilationData(acc.addDependencies(config))
+        is Config.Many -> (config.configs + remaining).compilationData(acc)
+        Config.Empty -> remaining.compilationData(acc)
+        is Config.AddMetaPlugins -> TODO("How do we bootstrap the Meta Plugin Quote system?")
+      }
+    }
+  }
+
```
rachelcarmena: Please, what's changing here?

idea-plugin/src/test/kotlin/arrow/meta/ide/plugins/typeclasses/TypeResolutionTest.kt
```diff
@@ -0,0 +1,96 @@
...
+
+        // TODO: make sure that the source file is copied into a source directory and the right dir structure
+        myFixture.configureFromExistingVirtualFile(sourceFile)
+        val psiFile = myFixture.file
+        assertEquals(psiFile.fileType, KotlinFileType.INSTANCE)
+
+        // TODO: fix compileCodeSnippet and copy data properly into build folder
+        val classOutputDir = compileCodeSnippet(codeSnippet)!!
+        val target = myFixture.copyDirectoryToProject(classOutputDir.path, buildFolder.path)
+        assertNotNull(target)
```
rachelcarmena: This assert could be in the testing library.

testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/TestingLibraryWrapper.kt
```diff
@@ -10,14 +10,14 @@ import java.nio.file.Paths
 
 private const val DEFAULT_FILENAME = "Example.kt"
 
-internal data class CompilationResult(
```
rachelcarmena: I think this change breaks the purpose of our testing library. I'm still trying to figure out what's the need but I'm thinking about 2 options:
* Thinking about an assertion and improving the testing library with that assertion.
* If it's only necessary to get the classes after the compilation, it seems that's not a goal of our testing library. We could add a function for compiling sources with the use of the compiler plugin in this module.

Does it make sense for you, @jansorg , @i-walker ?

testing-plugin/build.gradle
i-walker: added

docs/build.gradle
i-walker: yes it is try building the module without. That was the problem me an @jansorg  had

idea-plugin/src/test/kotlin/arrow/meta/ide/plugins/typeclasses/TypeResolutionTest.kt
i-walker: We're not building assertions in the testing dsl

i-walker:
This looks good. We can now work on it 😄

testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/TestingLibraryWrapper.kt
i-walker: I propose that we focus on implementing this deliverable and negotiate with @raulraja what the best approach is. I do understand that you value the set of principles in `testing-plugin` and I am certain there is a solution, which works for everyone.
Though for now, we need the time to iterate and fulfill our deadlines.

idea-plugin/src/main/kotlin/arrow/meta/ide/testing/env/types/HeavyTestSyntax.kt
```diff
@@ -1,3 +1,100 @@
...
+    myFixture: CodeInsightTestFixture,
+    srcDirName: String = "src",
+    buildDirName: String = "build",
+    srcFileName: String = "Source.kt",
+    config: List<Config>
+  ): HeavyTestSetUp? =
+    module.root?.let { root: VirtualFile ->
+      root.addDir(buildDirName, module)?.let { buildDir: VirtualFile ->
+        root.addDir(srcDirName, module)?.let { srcDir: VirtualFile ->
+          project.buildFolders().takeIf { it.isNotEmpty() }?.run {
```
jansorg: buildFolders is currently collecting build folders of all modules, I'll add `Module.buildFolders` and call it here to avoid confusion

idea-plugin/src/main/kotlin/arrow/meta/ide/testing/env/types/HeavyTestSyntax.kt
```diff
@@ -1,3 +1,100 @@
...
+  fun Source.copyToDir(name: String = "Source.kt", dir: VirtualFile, myFixture: CodeInsightTestFixture): KtFile? =
+    toFile(name, dir)?.let { file: VirtualFile ->
+      myFixture.configureFromExistingVirtualFile(file)
+      myFixture.file.fileType.safeAs()
+    }
+
+  /**
+   * @param receiver is the root file
+   */
+  fun VirtualFile.addDir(name: String, module: Module): VirtualFile? =
```
jansorg: At the moment this is always adding an excluded folder (see 2 lines below). I'll rename this to `addExcludedDir` and add method `addSourceDir`

idea-plugin/src/main/kotlin/arrow/meta/ide/testing/env/types/HeavyTestSyntax.kt
```diff
@@ -1,3 +1,101 @@
...
+    WriteAction.computeAndWait<VirtualFile, Throwable> {
+      val sourceFile = dir.createChildData(this, name)
+      sourceFile.setBinaryContent(toByteArray())
+      sourceFile
+    }
+
+  fun Source.copyToDir(name: String = "Source.kt", dir: VirtualFile, myFixture: CodeInsightTestFixture): KtFile? =
+    toFile(name, dir)?.let { file: VirtualFile ->
+      myFixture.configureFromExistingVirtualFile(file)
+      myFixture.file.fileType.safeAs()
```
jansorg: `fileType` is never a `KtFile`, but `file` is. I'll change this

testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/TestingLibraryWrapper.kt
raulraja: @i-walker CompilationResult should remain internal unless you have a valid reason to make it public as the compiler tests are hiddng that behind asserts. Why do you need this method public? We are attempting to make the DSL cohesive with the rest of the APIs and both compiler and ide should share the design but this model was not intended to be part of the public API

i-walker: In order to provide typeclass resolution tests we need the metadata, that the compiler-plugin produces. Within an IdeEnvironment we can only reference and create files, but without the engine of a compiler. Naturally, we would lean on something which we already use in `Meta` We are in need of a function in `testing-plugin`, where we can utilize a function `(Source) -> (List<Config>) -> CompilationResult` or something isomorphic.

We're only proposing a collaboration between `testing-plugin` and `idea-plugin`, unless we're dropping tests for `SyntheticResolution`. Consequently, we can not guarantee issues that the current `scala-idea-plugin` has e.g.: redlines on Syntheticmembers and other unresolved references we add to the language.

raulraja: What kind of metadata is that and how is it used? We can just add a new assert for that that gives you that metadata or analyzes.it in the interpreter of the test DSL.

i-walker: We need `.class` files and everything generated in the build folder.
We also need the actual files and directories, no interpretation or assertion.

i-walker: The underlying libraries has the property `outputDirectory, which includes some of the files we need. But we're not able to compile it with the current settings.

i-walker: We grab the metadata and copy it into the buildFolder of each IdeTestEnviroment. After that, we can initiate indexing as it would happen if a user would start a gradle build task in Idea.

rachelcarmena:
See proposal in #65

jansorg:
@i-walker I fixed a few of the classloader issues. The problem is, that the JetBrains plugin already comes with `kotlin-stdlib`, `kotlin-compiler-embedded` and a lot of the other kotlin libs used in idea-plugin's dependencies. 
Because `testing-plugin` was transitively adding a dependency on `kotlin-compiler-embedded` this library was included twice in the classpath. This confused the classloader which now had multiple separate .class files for KotlinFileType. I've not investigated much further, but this fixed the class cast exceptions I was seeing.

As a side-effect, this brings down the side of `idea-plugin.zip` from 44mb to 6.4mb.

There's one (last?) exception which I can't solve:
IntelliJ complains about `project` in `metaPlugin.registerMetaComponents(project, configuration)` used in `MetaPluginRegistrarComponent`. `project` of `MetaPluginRegistrarComponent` is from the IntelliJ SDK, but the expected type of `project` is from `kotlin-compiler`, afaik.
I don't know why this compiles (shadowed?), but still raises an exception when executed.
Do you know how to fix this?

I've made sure that the Kotlin plugin is bundling 1.3.50 by checking `META-INF/MANIFEST.MF` of `IDEA-U/ch-1/193.4932.9/plugins/Kotlin/lib/kotlin-stdlib.jar`.
Here's the full listing of what's included with the Kotlin plugin:
```
-rw-r--r-- 1 jansorg users    20448 19. Sep 20:04 allopen-ide-plugin.jar
-rw-r--r-- 1 jansorg users   347398 19. Sep 20:04 android-extensions-compiler.jar
-rw-r--r-- 1 jansorg users      339 19. Sep 20:04 android-extensions-ide.jar
-rw-r--r-- 1 jansorg users    17536 11. Sep 06:23 annotations-13.0.jar
-rw-r--r-- 1 jansorg users   611568 11. Sep 06:23 javaslang-2.0.6.jar
-rw-r--r-- 1 jansorg users    26418 11. Sep 06:23 javaslang-match-2.0.6.jar
-rw-r--r-- 1 jansorg users     2497 11. Sep 06:23 javax.inject-1.jar
drwxr-xr-x 2 jansorg users        3 19. Sep 20:04 jps
-rw-r--r-- 1 jansorg users    24351 19. Sep 20:04 kapt3-idea.jar
-rw-r--r-- 1 jansorg users    13442 19. Sep 20:04 kotlin-allopen-compiler-plugin.jar
-rw-r--r-- 1 jansorg users  1466574 19. Sep 20:04 kotlin-compiler-client-embeddable.jar
-rw-r--r-- 1 jansorg users   679560 19. Sep 20:04 kotlin-daemon-client.jar
-rw-r--r-- 1 jansorg users   793532 19. Sep 20:04 kotlin-daemon-client-new.jar
-rw-r--r-- 1 jansorg users  1182434 19. Sep 20:04 kotlin-daemon.jar
-rw-r--r-- 1 jansorg users   134181 19. Sep 20:04 kotlin-gradle-tooling.jar
-rw-r--r-- 1 jansorg users    24367 19. Sep 20:04 kotlin-noarg-compiler-plugin.jar
-rw-r--r-- 1 jansorg users 47508125 19. Okt 21:04 kotlin-plugin.jar
-rw-r--r-- 1 jansorg users  2793141 19. Sep 20:04 kotlin-reflect.jar
-rw-r--r-- 1 jansorg users    11773 19. Sep 20:04 kotlin-sam-with-receiver-compiler-plugin.jar
-rw-r--r-- 1 jansorg users   167775 19. Sep 20:04 kotlin-scripting-common.jar
-rw-r--r-- 1 jansorg users   292936 19. Sep 20:04 kotlin-scripting-compiler-impl.jar
-rw-r--r-- 1 jansorg users     2992 19. Sep 20:04 kotlin-scripting-intellij.jar
-rw-r--r-- 1 jansorg users   157602 19. Sep 20:04 kotlin-scripting-jvm.jar
-rw-r--r-- 1 jansorg users    42123 19. Sep 20:04 kotlin-script-runtime.jar
-rw-r--r-- 1 jansorg users    73668 19. Sep 20:04 kotlin-script-util.jar
-rw-r--r-- 1 jansorg users   170949 19. Sep 20:04 kotlin-stdlib-common.jar
-rw-r--r-- 1 jansorg users  1326290 19. Sep 20:04 kotlin-stdlib.jar
-rw-r--r-- 1 jansorg users     3125 19. Sep 20:04 kotlin-stdlib-jdk7.jar
-rw-r--r-- 1 jansorg users    15472 19. Sep 20:04 kotlin-stdlib-jdk8.jar
-rw-r--r-- 1 jansorg users    43214 19. Sep 20:04 kotlin-util-io.jar
-rw-r--r-- 1 jansorg users   123558 19. Sep 20:04 kotlin-util-klib.jar
-rw-r--r-- 1 jansorg users  1115113 19. Sep 20:04 kotlinx-coroutines-core-1.2.1.jar
-rw-r--r-- 1 jansorg users    15183 19. Sep 20:04 kotlinx-coroutines-jdk8-1.2.1.jar
-rw-r--r-- 1 jansorg users   453068 19. Sep 20:04 kotlinx-serialization-compiler-plugin.jar
-rw-r--r-- 1 jansorg users   393002 11. Sep 06:23 markdown-0.1.25.jar
-rw-r--r-- 1 jansorg users    22333 19. Sep 20:04 noarg-ide-plugin.jar
-rw-r--r-- 1 jansorg users    20223 19. Sep 20:04 sam-with-receiver-ide-plugin.jar

```

rachelcarmena:
@jansorg , maybe it's solved with #65 because it replaces `testing-plugin` by `kotlin-compile-testing`. 

#65 is not a pull request on `master`branch but on this branch: `ja-heavy-testing`.

jansorg:
@rachelcarmena Thanks! Probably only partially. Most of the dependencies excluded from testing-plugin are inherited from kotlin-compile-testing. I‘d be glad to replace testing-plugin, though!
My last commit also added more necessary excludes.

jansorg:
I've merged #65. Now there are a few more remaining problems.
1. Compilation of a Kotlin code snippet starts, but breaks with an exception. I've update the description of this PR. The cause is that `KotlinFileType` is available twice on the classpath, once in the Kotlin IDE plugin and once in the dependency `kotlin-compiler-embeddable.jar`. The dependency is required and can't be removed, though.

Exception:
```
 java.lang.ClassCastException: org.jetbrains.kotlin.idea.KotlinFileType cannot be cast to org.jetbrains.kotlin.com.intellij.openapi.fileTypes.FileType
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment$Companion.registerApplicationServices(KotlinCoreEnvironment.kt:657)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment$Companion.createApplicationEnvironment(KotlinCoreEnvironment.kt:523)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment$Companion.getOrCreateApplicationEnvironmentForProduction(KotlinCoreEnvironment.kt:476)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment$Companion.createForProduction(KotlinCoreEnvironment.kt:430)
	at org.jetbrains.kotlin.cli.jvm.K2JVMCompiler.createCoreEnvironment(K2JVMCompiler.kt:253)
	at org.jetbrains.kotlin.cli.jvm.K2JVMCompiler.doExecute(K2JVMCompiler.kt:151)
	at org.jetbrains.kotlin.cli.jvm.K2JVMCompiler.doExecute(K2JVMCompiler.kt:54)
	at org.jetbrains.kotlin.cli.common.CLICompiler.execImpl(CLICompiler.kt:84)
	at org.jetbrains.kotlin.cli.common.CLICompiler.execImpl(CLICompiler.kt:42)
	at org.jetbrains.kotlin.cli.common.CLITool.exec(CLITool.kt:104)
	at com.tschuchort.compiletesting.KotlinCompilation.compileKotlin(KotlinCompilation.kt:545)
	at com.tschuchort.compiletesting.KotlinCompilation.compile(KotlinCompilation.kt:713)
```

rachelcarmena:
@jansorg, we've just merged a fix on master. There was an error when shadowing `compiler-plugin`.

jansorg:
@rachelcarmena Thanks! I've merge master. It's still broken.
The `kotlin-compiler-embedded` dependency comes from `com.github.arrow-kt:kotlin-compile-testing:1.3.0`.

jansorg:
as far as I can tell, this will be superseded by https://github.com/arrow-kt/arrow-meta/pull/252 . Heavy tests won't be needed anymore when that wip PR is finished and merged.

rachelcarmena:
> as far as I can tell, this will be superseded by #252 . Heavy tests won't be needed anymore when that wip PR is finished and merged.

Thanks @jansorg !! :tada: So, let's close it :+1:

PR #11, opened 2019-10-25
WIP: Replace lang operators with arrow operators

ahinchman1:
Linked Issue: [#1633](https://github.com/arrow-kt/arrow/issues/1633)

First attempt at building out an example for building an binary operator psi locator and writing out an example Eq plugin.

This feature may sit for a while until after KotlinConf, but the main purpose at the moment is to get more familiar with the Kotlin compiler!

PR #60, opened 2019-11-01
WIP: EqOperator Plugin

ahinchman1:
Linked Issue: #1633

Old Merge Request [here](https://github.com/arrow-kt/arrow-meta/pull/11)

First attempt at building out an example for building an binary operator psi locator and writing out an example Eq plugin.

This feature may sit for a while until after KotlinConf, but the main purpose at the moment is to get more familiar with the Kotlin compiler!

### DIFF
--- a/compiler-plugin/src/main/kotlin/arrow/meta/MetaPlugin.kt
+++ b/compiler-plugin/src/main/kotlin/arrow/meta/MetaPlugin.kt
@@ -2,6 +2,7 @@ package arrow.meta
 
 import arrow.meta.phases.CompilerContext
 import arrow.meta.plugins.comprehensions.comprehensions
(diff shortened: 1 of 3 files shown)


raulraja:
Closing this as it's part of the support for type proofs extensions on operator overloading.
Some type-classes like `Eq` and `Hash` will take precedence over `Object.equals` and `Object.hashcode`

```kotlin
interface Eq<A> {
  val value: A
  fun eqv(other: A): A 
}

inline class UserEq(val value: User) : Eq<User> {
  override fun eqv(other: User): User = TODO()
}

@Proof(TypeProof.Extension)
fun User.eq(): Eq<User> = UserEq(this)

user1 == user2 // replaces call to `EQEQ(user1, user2)` with `user1.eq().eqv(user2)` proof.
```

Issue #10, opened 2019-10-25
Ide Blockers for Alpha Release


i-walker:
We currently facing a few issues during registration:
- [x] **_Multiple Registrations in MetaPluginRegistrarComponent_**
```
metaPlugin.registerMetaComponents(project, configuration)
```
As of today we're applying for each Project window the `IdeMetaPlugin`, thus `Extensions`, which are applied on `Application` level will have duplicates. 

- [x] **_Undisposed Extensions_**
The `Disposable` in `ExtensionProvider` needs to be disposed, same applies for `MetaPluginRegistrarComponent#disposeComponent`. If not, all registered Extensions are not removed from memory, which leads to memory leaks and higher RAM consumptions. Possible solutions encompase to lift `MetaPluginRegistrarComponent` to an `ApplicationComponent` and define a `Disposable` strategy without the need of the `PicoContainer`, but this is an open discussion.

- [x] **_Make Extensions dumbAware_**
We'll need to indicate `Extensions`, which are available during `Indexing` and supply that in the `IdeSyntax`. `DumbAware` and `DumbService` should facilitate us with sufficient tools.

These need to be dealt with after we have a sufficient test environment and before we release the Alpha. The first 2 have High Priority, where as the last is a minor fix we can do after the release. Feel free to correct me @jansorg.
We are open for suggestions.

jansorg:
@i-walker cc @raulraja  here are some notes how the issues above could be solved 

PR # is changing `idea-plugin` to simplify the necessary changes, which are outlined below.

# Multiple registrations
The main point is that there are application extensions and project extensions.

There's only one instance of [Application](https://github.com/JetBrains/intellij-community/blob/192/platform/core-api/src/com/intellij/openapi/application/Application.java) for an IntelliJ process. Application extensions are only registered once.

There's an arbitrary number of instances of a [Project](https://github.com/JetBrains/intellij-community/blob/192/platform/core-api/src/com/intellij/openapi/project/Project.java) for an IntelliJ process (0 for no open project, 2 for two open projects, etc.)
Project level extensions have to be registered once for each of the projects. They are unregistered and disposed when the project is closed. This happens, for example, when you close a project window.

Application extensions use [ExtensionPointName](https://github.com/JetBrains/intellij-community/blob/192/platform/extensions/src/com/intellij/openapi/extensions/ExtensionPointName.java)

[...]

PR #62, opened 2019-11-02
Single entry point to register application and project extensions

jansorg:
This PR uses an extension so that only one class is managing application and project extensions.

It's related to https://github.com/arrow-kt/arrow-meta/issues/10#issuecomment-549043757


idea-plugin/src/main/kotlin/arrow/meta/ide/MetaExtensionsRegistrar.kt
```diff
@@ -0,0 +1,44 @@
...
+
+class MetaExtensionsRegistrar : ApplicationInitializedListener {
+  companion object {
+    private val LOG = Logger.getInstance("#arrow.metaRegistrar")
+  }
+
+  override fun componentsInitialized() {
+    val app = ApplicationManager.getApplication()
+
+    // register application-level extensions
```
i-walker: How can do we register application-level extensions
Most of our ide dsl is based on Application-level extensions

jansorg: @i-walker The current dsl is already setting up application-level extensions, e.g. using [this code](https://github.com/arrow-kt/arrow-meta/blob/725d3754b4ca0899c5bbd517e28c072da058fd77/idea-plugin/src/main/kotlin/arrow/meta/ide/internal/registry/IdeInternalRegistry.kt#L130). That's fine for the current set of supported extensions.
The issue is that this code is called too often.

Part of the problem is that the registration above is triggered by `InternalRegistry.registerMetaComponents(project, configuration)`. This takes a project and therefore can't be executed in application scope. 
This way IDE application extensions, IDE project extensions (none atm) and compiler plugins are all registered within a project scope. 

As far as I can tell the lifecycle of IDE extensions is not related to the lifecycle of compiler plugins. At least application-level IDE extensions should not be forced into the lifecycle of compiler plugins, which requires a project.

Currently IDE extensions are kept here:
```
open class IdeMetaPlugin : MetaPlugin(), IdeInternalRegistry, IdeSyntax {
  @ExperimentalContracts
  override fun intercept(ctx: CompilerContext): List<Plugin> =
    super.intercept(ctx) +
      initialIdeSetUp +
      higherKindsIdePlugin +
      typeclassesIdePlugin +
      comprehensionsIdePlugin +
      opticsIdePlugin +
      nothingIdePlugin
}
```

It might be possible to store the IDE plugins in a separate list and offer a method `IdeMetaPlugin.registerApplicationExtensions()` or similar which just iterates on that list. `MetaPluginRegistrar` should then call this method in `componentsInitialized()`.

i-walker: Yes, this sounds legitimate. We might want to add @raulraja to this, because that would mean we either have to abstract away which services are instantiated at what time or we might have to add another cycle to the `intercept` methode

ahinchman1: I think it sounds like a great idea. There would be a bit of a refactor involved but if @jansorg is able to propose a composition to satisfy the same amount of flexibility, it could be worth a PR in itself

idea-plugin/src/main/kotlin/arrow/meta/ide/MetaExtensionsRegistrar.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    // register a project lifecycle listener to register project components when necessary
+    // the listener registers project-level extensions
+    app.messageBus.connect(app).subscribe(ProjectLifecycleListener.TOPIC, object : ProjectLifecycleListener {
+      override fun beforeProjectLoaded(project: Project) {
+        LOG.info("beforeProjectLoaded(${project.name})")
+
+        val start = System.currentTimeMillis()
+        val configuration = CompilerConfiguration()
+        // TODO: only register project extensions here
```
ahinchman1: From the comments I see, it looks like there's an attempt for a proposal on how we can register the project extensions here - if we created an abstraction that shares the same kind of control with ide dsl, would the logic be similar or would it be shared?

i-walker: May you rephrase your question. I don't really understand your question?

i-walker:
It turns out that we have to consider a few major changes with lifting the entrypoint to the application. The last commit is just a WIP because we need to discuss the ergonomics of writing plugins. 
1. We should consider a way where we abstract away whether the user needs to pass in ProjectLevel Extensions or ApplicationLevel Extensions. We're not anticipating any development in ProjectExtensions in the Ide, due to the lack of ProjectlevelExtensions and functionality. Furthermore, we're able to define Service Extensions, which are more lightweight and are created on-demand rather than by Project start.
This is one possible solution
```
- create one intercept Function where Users can pass in both ExtensionTypes
- filter in `MetaPluginRegistrar` accordingly
```
This would mean, that we're traversing the list of Plugins twice. But have a unified Interface.
Or
```
- create 2 functions one for ProjectLevel extensions and another for ApplicationLevel
- this is showcased by the latest changes
- This also introduces a slight learning curve, because we did not explicitly Type a distinction between ProjectLevel and ApplicationLevel Extensions, this can be added with something as simple as adding another toplevel hierarchy in form of an ADT. The learning curve would then be diminished because Users could only add the right set of Extensions to the respective function. 
```

Another problem, which arises is that in `MetaPluginRegistrar`, we applying a `project` from kotlin. Now, this is fine and has worked in the past. Because we shadow it. Though, for this to work with idea successfully. We need to restructure `registrarMetaComponent`, so that the compiler logic, won't affect the editor. It works now because we're not utilizing anything which is created in that function and solely works with the output. 

In other words, we're looking for a way to separate the List of Extensions from the CompilerContext, even though both are dependent on each other in `Meta`.

@raulraja what do you think is best for our Design Goals.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/nothing/NothingIdePlugin.kt
```diff
@@ -1,18 +1,20 @@
...
 
-val IdeMetaPlugin.nothingIdePlugin: Plugin
+val IdeMetaPlugin.nothingIdePlugin: IdePlugin
   get() = "NothingIdePlugin" {
     meta(
       addLineMarkerProvider(
         icon = ArrowIcons.NOTHING,
-        transform = { it.safeAs<KtUserType>()?.takeIf { type -> type.referencedName == "Nothing" } },
+        transform = { it.safeAs<KtUserType>()?.takeIf { type -> type.referencedName == "Nothing" }
+          ?.referenceExpression?.getReferencedNameElement()},
```
i-walker: This change is because, we're not subscribing to the TypeIdentifier

idea-plugin/src/main/kotlin/arrow/meta/ide/internal/registry/IdeInternalRegistry.kt
```diff
@@ -50,19 +55,33 @@ internal interface IdeInternalRegistry : InternalRegistry {
...
-      is IntentionExtensionProvider -> registerIntentionExtensionProvider(currentPhase)
-      is SyntaxHighlighterExtensionProvider -> registerSyntaxHighlighterExtensionProvider(currentPhase)
-      else -> LOG.error("Unsupported ide extension phase: $currentPhase")
+  fun registerMetaApplication(application: Application, compilerConfiguration: CompilerConfiguration): Unit =
+    intercept().forEach { plugin ->
+      println("Registering plugin: $plugin extensions: ${plugin.meta}")
+      plugin.meta.invoke().forEach(::registerIdeExclusiveApplicationPhase)
     }
 
+  fun registerIdeExclusiveApplicationPhase(currentPhase: ExtensionPhase): Unit = when (currentPhase) {
```
i-walker: There is also the option to create one unified `registerIdeExclusivePhase` we're we add State explicitly in the FunctionParameter. Though this poor design makes only sense if we want to keep the editor open for ProjectExtensions, even though their fairly used in the ide.

idea-plugin/src/main/kotlin/arrow/meta/ide/internal/registry/IdeInternalRegistry.kt
```diff
@@ -29,14 +30,18 @@ import com.intellij.codeInsight.intention.IntentionManager
...
 import com.intellij.openapi.extensions.Extensions
 import com.intellij.openapi.fileTypes.SyntaxHighlighterFactory
+import org.jetbrains.kotlin.config.CompilerConfiguration
 import org.jetbrains.kotlin.container.useImpl
 import org.jetbrains.kotlin.idea.KotlinFileType
 import org.jetbrains.kotlin.idea.KotlinLanguage
 
 internal interface IdeInternalRegistry : InternalRegistry {
 
+  fun intercept(): List<IdePlugin>
```
i-walker: This is the function we would utilize for application-level Extensions, whereas `project-level` Extensions would be added in the original `intercept(ctx: CompilerContext)`. I am open for suggestions.

i-walker:
Thanks a lot @jansorg 🎉 👍 
This proposal was instrumental for #519. Can we close this PR

jansorg:
@i-walker Yes, sure. Feel free to close it. It's very old and not necessary anymore, I assume.

PR #64, opened 2019-11-03
Union Types Draft

raulraja:
This PR introduces a draft for Union types with a test that compiles and runs:

The following code shows how Either is assimilated by the environment and values are converted
automatically if they are part of the Union or have a compatible target nullable type in which to unfold the Union.

Some examples:
```diff
- sealed class Error {
-   object NotANumber : Error()
- }
-
- fun parse(s: String): Either<Error, Int> =
-   if (s.matches(Regex("-?[0-9]+"))) Either.Right(s.toInt())
-   else Either.Left(Error.NotANumber)
+ object NotANumber // no need for hierarchy
+ typealias `NotANumber | Int` = Union2<NotANumber, Int> // arity up to 22
+
+ fun parse(s: String): `NotANumber | Int` = //perhaps we can allow somehow `|`
+  if (s.matches(Regex("-?[0-9]+"))) s.toInt() // Implicit cast, No need to Right(Int)
+  else NotANumber // Implicit conversion, No need to Left(NotANumber)
```

```kotlin
fun f1(): Union2<String, Int> = "a" //Implicit conversion to Union("a")
fun f12(): Union2<String, Int> { 
  val s = "x"
  val z: Union2<String, Int> = s //Implicit conversion to Union("x")
  return z
}
fun f2(): Union2<String, Int> = 1 
fun f3(): String? = f1() //Implicit conversion to String? from Union
fun f4(): Int? = f2()//Implicit conversion to Int? from Union
```


i-walker:
Wouldnt the cast here be:
```
fun f12(): Union2<String, Int> { 
  val s = "x"
  val z: Union2<String, Int> = s //Implicit conversion to Union("x")
  return z
}
```
instead of Union(x)

raulraja:
@i-walker yes, feel free to suggest it as suggestion change in a comment on the file diff and we'll mere it in

pakoito:
How do you fold over the types? If you're using `is NotANumber` and `isInt`, the `inline` class may fail due to boxing.

clesaec:
`Union2<A, B>` should be strictly equals to `Union2<B, A>`
As well as `Union2<A, Union2<B,C>>` and `Union2<B, Union2<A, C>>`
(Not so simple)

The ceylon project https://github.com/eclipse/ceylon treated union type as a sequence of type (language/runtime/org/eclipse/ceylon/compiler/java/runtime/metamodel/meta/UnionTypeImpl.java)

AdrianRaFo:
Could we keep an alias Union2<A, B> = Either<A, B>?

1Jajen1:
> Could we keep an alias Union2<A, B> = Either<A, B>?

That would not be a good idea imo, since while the structure is equivalent the semantics may not be, especially typeclasses, but also order (if Union<A, B> = Union<B, A>). Also how would you do higher kinded types if all you have is a union? I think inline classes over unions or another newtype like approach is a decent solution. Inline classes just suck because they are boxed way too much if you are not careful :/

compiler-plugin/src/test/kotlin/arrow/meta/plugins/lens/LensTest.kt
```diff
@@ -27,12 +25,8 @@ class LensTest {
       """
 
     assertThis(CompilerTest(
-        config = {
-          addCompilerPlugins(compilerPlugin) + addDependencies(arrowAnnotations, arrowOptics)
-        },
-        code = {
-          codeSnippet.source
-        },
+        config = { metaDependencies + addDependencies(arrowOptics) },
```
ahinchman1: @raulraja I really like what you did with the dependencies bit here. I think I have to use that idea in my own branch!

compiler-plugin/src/test/kotlin/arrow/meta/plugins/union/UnionTest.kt
```diff
@@ -0,0 +1,89 @@
...
+        |interface Union3<out A, out B, out C> : Union2<A, B>
+        |interface Union4<out A, out B, out C, out D> : Union3<A, B, C>
+        |
+        |inline class Union(val value: Any?) :
+        |  Union2<Nothing, Nothing>,
+        |  Union3<Nothing, Nothing, Nothing>,
+        |  Union4<Nothing, Nothing, Nothing, Nothing> {
+        |
+        |   inline operator fun <reified A> invoke(): A? =
+        |     value as? A
```
ahinchman1: I actually don't know if the `@extensions` operator is being used at all here in the source code - is it necessary to work with the annotation dependency if we're not using some kind of `@`? Perhaps I'm not understanding how the annotations feature works

1Jajen1:
Sadly inline class constructors have to be public otherwise you could scope them and implicitly call them to get implicit conversion on assignments. The more I try inline classes the more limited they feel... Would love to see something similar to haskell newtypes, that would solve quite a bit of problems with using Unions.

compiler-plugin/src/test/kotlin/arrow/meta/plugins/union/UnionTest.kt
raulraja: I think we need annotations here because even though in this case it's not needed this runs the full suite of meta plugins and we are gonna test them togwther in further tests I'm adding like unions I'm type classes etc

raulraja:
@AdrianRaFo Either will remain but we will slowly recommend people to use Unions and those would be used in arrow as well in place of either. Their runtime is much leaner

compiler-plugin/src/main/kotlin/arrow/meta/plugins/union/UnionsPlugin.kt
```diff
@@ -0,0 +1,263 @@
...
+         */
+        typeChecker(::UnionTypeChecker),
+        /**
+         * ```kotlin
+         * fun proof(): Union2<String, Int> = 1
+         * val willBeNullButOk: String? = proof() //ok
+         * val willBeInt: Int? = proof() //ok
+         * ```
+         */
+        suppressDiagnostic(Diagnostic::suppressTypeMismatchOnNullableReceivers),
```
i-walker: We need the same pice in the ide.
May you add that or can I do it?

raulraja: Please go ahead, id love others to understand the plugin

PR #63, opened 2019-11-03
`testing-plugin` proposal

i-walker:
@raulraja cc @rachelcarmena 
This PR tries to turn the return-type of `assertThis` into an GADT. 

Furthemore, extends `CompilationResult` with the property `generatedFiles`, which we need for this PR including `classesDirectory`. 
In the following commit, we'll attempt to extend the algebra to provide these Properties. 

What is the best approach for this?
 Should we create a new `assertThis` like function, who does the resource acquisition.
Or something else. 

docs need the additional Gradle settings, in order to execute `./gradlew clean :idea-plugin:test`
@jansorg though the Test passes, the compilation returns an `InternalError`, due to ClassCastException.
Here is the Stacktrace:
```
java.lang.ClassCastException: org.jetbrains.kotlin.idea.KotlinFileType cannot be cast to org.jetbrains.kotlin.com.intellij.openapi.fileTypes.FileType
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment$Companion.registerApplicationServices(KotlinCoreEnvironment.kt:657)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment$Companion.createApplicationEnvironment(KotlinCoreEnvironment.kt:523)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment$Companion.getOrCreateApplicationEnvironmentForProduction(KotlinCoreEnvironment.kt:476)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment$Companion.createForProduction(KotlinCoreEnvironment.kt:430)
	at org.jetbrains.kotlin.cli.jvm.K2JVMCompiler.createCoreEnvironment(K2JVMCompiler.kt:253)
	at org.jetbrains.kotlin.cli.jvm.K2JVMCompiler.doExecute(K2JVMCompiler.kt:151)
	at org.jetbrains.kotlin.cli.jvm.K2JVMCompiler.doExecute(K2JVMCompiler.kt:54)
	at org.jetbrains.kotlin.cli.common.CLICompiler.execImpl(CLICompiler.kt:84)
	at org.jetbrains.kotlin.cli.common.CLICompiler.execImpl(CLICompiler.kt:42)
	at org.jetbrains.kotlin.cli.common.CLITool.exec(CLITool.kt:104)
[...]

PR #65, opened 2019-11-04
Proposal on `ja-heavy-testing`

rachelcarmena:
## Goal

This is a pull request on `ja-heavy-testing`:
* To use `kotlin-compile-testing` in `idea-plugin`
* To avoid #63 

## Details

I think `idea-plugin` doesn't need to use `testing-plugin` but `kotlin-compile-testing`:

details-testing-library

### DIFF
--- a/idea-plugin/build.gradle
+++ b/idea-plugin/build.gradle
@@ -20,7 +20,8 @@ dependencies {
     exclude group: "org.jetbrains.kotlin", module: "kotlin-stdlib-jdk8"
   }
   compile project(path: ':compiler-plugin', configuration: 'shadow')
-  implementation(project(":testing-plugin")) {
+  compileOnly "io.github.classgraph:classgraph:4.8.47"
+  implementation("com.github.arrow-kt:kotlin-compile-testing:1.3.0") {
     // already provided by the kotlin-plugin
     exclude group: "org.jetbrains.kotlin", module: "kotlin-compiler-embeddable"
     exclude group: "org.jetbrains.kotlin", module: "kotlin-stdlib"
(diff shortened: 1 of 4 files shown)


i-walker:
We definitely can use `kotlin-compile-testing` this PR has the same problem as #63. 

Do you know how we can fix the Internal Error that we're getting?
if you run `clean :idea-plugin:test` you can see it.

i-walker:
Thank you @rachelcarmena.
We can keep this as a draft until we find a solution for the `internal_error`. This doesnt have to happen this week.

jansorg:
As long as it‘s possible to generate .class files from a code snippet, I‘m all for replacing testing-plugin.

rachelcarmena:
> As long as it‘s possible to generate .class files from a code snippet, I‘m all for replacing testing-plugin.

Great @jansorg! Yes, it's possible to generate .class files from a code snippet.

`testing-plugin` is using `kotlin-compiler-testing` to get them.

ahinchman1:
Hey @rachelcarmena! I was hoping for some clarification - would it be necessary to use the `testing-plugin` if we wanted to write tests for the IDE or is the plan to separate the testing setup into their respective projects?

rachelcarmena:
> Hey @rachelcarmena! I was hoping for some clarification - would it be necessary to use the `testing-plugin` if we wanted to write tests for the IDE or is the plan to separate the testing setup into their respective projects?

It seems the tests for the IDE just need the compiled files and that's already provided by `kotlin-compile-testing`.

`testing-plugin` is not ready for that need and adds more things that tests for the IDE wouldn't use.

That's the reason why it make sense to use  `kotlin-compile-testing` for IDE tests.

jansorg:
@rachelcarmena I've merged your commits into `ja-heavy-testing` (#57). Thank you for the proposal, this definitely simplifies the compiler integration!

rachelcarmena:
Thanks to you @jansorg !! Good to know it :+1:

PR #68, opened 2019-11-04
testing-plugin: multiple asserts

rachelcarmena:
* Add multiple asserts
* Improve the message error when evaluating expressions


compiler-plugin/src/test/kotlin/arrow/meta/plugins/comprehensions/ComprehensionsTest.kt
```diff
@@ -310,8 +254,8 @@ class ComprehensionsTest {
         |
         |""".source
       },
-      assert = {
-        failsWith { it.contains("Unresolved reference: a") }
+      asserts = {
+        listOf(failsWith { it.contains("Unresolved reference: a") })
```
JorgeCastilloPrz: 💡 Something used in some Android testing frameworks when it comes to multiple matches / assertions is `allOf()` which can be just a forwarding to listOf. Just a random idea. lisOf is also fine!

rachelcarmena: Thanks @JorgeCastilloPrz! It seems like `assertAll` from JUnit5. Thanks for the idea, added!

testing-plugin/src/test/kotlin/arrow/meta/plugin/testing/ExampleTest.kt
```diff
@@ -14,8 +14,8 @@ class ExampleTest {
         | 
         """.source
       },
-      assert = {
```
raulraja: are we still preserving this? there are cases where we may want a single assert

rachelcarmena: Good idea. thanks!! I'll change to have both options :+1:

rachelcarmena: For instance, `assert` for a single assert and `assertAll` for multiple asserts.

rachelcarmena: I recovered `assert` instead of `asserts` and I'll add the commented improvement into another PR :+1:

PR #86, opened 2019-11-06
WIP: Document classOrObject in Quotes

ahinchman1:
Add more context to `classOrObject` function:

Screen Shot 2019-11-06 at 10 35 11 AM

There's a lot in here to explain these two parameters - I worried the length of the comments might make the file look clunky, but if this is the look we wish to go for I will continue on!

### DIFF
--- a/compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
+++ b/compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
@@ -13,6 +13,55 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
 import org.jetbrains.kotlin.psi.psiUtil.modalityModifierType
 import org.jetbrains.kotlin.psi.psiUtil.visibilityModifierType
 
+/**
+ * An extension function of [Meta] and inheriting from [ExtensionPhase], [classOrObject] was designed to to feed in
+ * virtually any kind of [KtClass] predicate, followed by mapping a function that takes the desired [Scope] of our
+ * [KtClass] to change whatever PSI elements desired.
+ *
+ * @param [match] designed to to feed in any kind of [KtClass] predicate returning a [Boolean]
+ *
+ * For example, the [LensPlugin] and the [HigherKindPlugin] favors enabling ad-hoc polymorphism and more for the
+ * compiler by taking advantage of typeclass checks.
+ *
+ * In the [LensPlugin], we check to see if a type is a product type.  To add context, the cardinality of a product
+ * type is the product of their cardinalities:
+ *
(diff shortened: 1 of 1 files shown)


compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,6 +13,55 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
 import org.jetbrains.kotlin.psi.psiUtil.modalityModifierType
 import org.jetbrains.kotlin.psi.psiUtil.visibilityModifierType
 
+/**
+ * An extension function of [Meta] and inheriting from [ExtensionPhase], [classOrObject] was designed to to feed in
+ * virtually any kind of [KtClass] predicate, followed by mapping a function that takes the desired [Scope] of our
+ * [KtClass] to change whatever PSI elements desired.
+ *
+ * @param [match] designed to to feed in any kind of [KtClass] predicate returning a [Boolean]
```
ahinchman1: Can't seem to figure out why the param keeps shoving down the description to the next line!

nomisRev: `@param`, `@return` etc should be at the of the KDoc as convention, no?

Not sure if this is related to "shoving down the description", I am not sure I understand what you mean  by that :D

ahinchman1: Ah yes, it's happening definitely in the screenshot, regardless of whether I'm doing `@param match ...` or `@param [match] ....`

It should be part of the KDocs yeah!

ahinchman1:
Sorry, I have to close this one and open up a new one because I accidentally worked over Raul's branch. @nomisRev I will link you the new one momentarily!

ahinchman1:
@nomisRev Here it is! Sorry for the mix-up: https://github.com/arrow-kt/arrow-meta/pull/87

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
nomisRev: I'm sorry I meant:
Shouldn't be at the end/bottom of the KDoc.
[Example in Arrow Fx Concurrent](https://github.com/arrow-kt/arrow/blob/c3fa9ec8567da5bfbd9ecaac2bc055c908dfac00/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/Concurrent.kt#L138)

nomisRev:
All good no problem :)

PR #87, opened 2019-11-06
Update documentation for classOrObject(...)

ahinchman1:
Add more context to classOrObject function:
Screen Shot 2019-11-06 at 10 35 11 AM
There's a lot in here to explain these two parameters - I worried the length of the comments might make the file look clunky, but if this is the look we wish to go for I will continue on!


compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,97 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
 import org.jetbrains.kotlin.psi.psiUtil.modalityModifierType
 import org.jetbrains.kotlin.psi.psiUtil.visibilityModifierType
 
+/**
+ * An extension function of [Meta] and inheriting from [ExtensionPhase], [classOrObject] was designed to to feed in
+ * virtually any kind of [KtClass] predicate, followed by mapping a function that takes the desired [Scope] of our
+ * [KtClass] to change whatever PSI elements desired.
+ *
+ * @param [match] designed to to feed in any kind of [KtClass] predicate returning a [Boolean]
```
nomisRev: All these block tags `@param`, `@return` etc should all be grouped at the bottom of the KDoc, that is done by convention.

Example: [Arrow Fx Concurrent](https://github.com/arrow-kt/arrow/blob/c3fa9ec8567da5bfbd9ecaac2bc055c908dfac00/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/Concurrent.kt#L138).

I think that might also fix the newline issue

ahinchman1: ahh - thank you so much Simon! This is helpful.

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,111 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
   quote(match, map) { ClassScope(it) }
 
+/**
+ * The [ClassScope] is projected over the template to allow destructuring of the different parts of the
+ * [KtClass]. The scope enables template syntax where the user may add new members or modify the class structure
+ * before it's compiled.
+ *
+ * @param value the PSI element being fed in for the compiler plugin.
+ * @param `@annotationEntries`
+ * @param modality
```
ahinchman1: KDocs doesn't appear to generate `quotes` as parameter names here smh

Screen Shot 2019-11-06 at 4 29 35 PM

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,111 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ * fun isProductType(ktClass: KtClass): Boolean =
+ *      ktClass.isData() &&
+ *        ktClass.primaryConstructorParameters.isNotEmpty() &&
+ *        ktClass.primaryConstructorParameters.all { !it.isMutable } &&
+ *        ktClass.typeParameters.isEmpty()
+ * ```
+ *
+ * For a [KtClass] to be a product type, we check to see if the [KtClass] is an ADT, and that the cardinality of
+ * the constructor parameters is greater than one and non-mutable. For the [LensPlugin], we've just demonstrated how we
+ * can check for the product type of a [KtClass] by access the properties made available by the Kotlin PSI change.
```
ahinchman1: @nomisRev I thought it might be cool to include an example from arrow-meta to showcase FP capabilities we're enabling in the compiler on top of the basic example I give below. Sometimes not including more complicated examples limits the imagination of those reading documentation

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
nomisRev: Do you mean "param `@annotationEntries\`"?
I am not sure if Dokka supports that properly.

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
nomisRev: Yes, that's great especially since it'll also generate the website documentation!

We also include playground example snippets in Arrow, and hopefully with Arrow Meta IDE plugins we can give them support for playgrounds in the IDE :D

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
ahinchman1: I can just go with annotationEntries perhaps to keep the same formatting if that makes more sense?

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
ahinchman1: Oh yeah, that was another thing too - so I have some executable code for a smaller example, the question I meant to ask is "should we have executable playground code for this example?" I wanted to reduce the code to focus on the bits that mattered most, but it's not totally consistent with the `runAnk` task allowing Kotlin to compile and execute in the simple example plugin.

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
ahinchman1: Also, what is the significance that we want this parameter to be read as a string? I thought it was curious but it might be worth commenting on in the documentation, regardless whether it's significant or nah

ahinchman1:
Issue on the ank build explained by @raulraja 👍 

```The ank error about not compiling the PsiElement was not working because there were both versions of the compiler plugin in the classpath - the shadowed version pulled transitively by the idea plugin and the regular version was in the build file, so there were two representations in the class-paths as to what PsiElement when there showed have just been the shadowed one```

File: compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt [hunk 4]
```diff
@@ -13,12 +13,150 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
 
+/**
+ * [classOrObject] is a function that intercepts all [KtClass] elements that [match]
+ * allowing a [Transform] to change the intercepted AST tree before compilation.
+ *
+ * An extension function of [Meta] and inheriting from [ExtensionPhase], [classOrObject] was designed to to feed in
+ * virtually any kind of [KtClass] predicate, followed by mapping a function that takes the desired [Scope] of our
+ * [KtClass] to change whatever PSI elements desired.
+ *
+ * For example, the [LensPlugin] and the [HigherKindPlugin] favors enabling ad-hoc polymorphism and more for the
```
raulraja:
```suggestion
 * For example, the [LensPlugin] and the [HigherKindPlugin] favor easy update of immutable data structures and ad-hoc polymorphism that does not require inheritance.
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,150 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ * compiler by taking advantage of typeclass checks.
```
raulraja:
```suggestion
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,150 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ *
+ * In the [LensPlugin], we check to see if a type is a product type.  To add context, the cardinality of a product
+ * type is the product of their cardinalities:
```
raulraja:
```suggestion
 * type is the product of its contained properties:
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,150 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ * fun isProductType(ktClass: KtClass): Boolean =
+ *      ktClass.isData() &&
+ *        ktClass.primaryConstructorParameters.isNotEmpty() &&
+ *        ktClass.primaryConstructorParameters.all { !it.isMutable } &&
+ *        ktClass.typeParameters.isEmpty()
+ * ```
+ *
+ * For a [KtClass] to be a product type, we check to see if the [KtClass] is an ADT, and that the cardinality of
+ * the constructor parameters is greater than one and non-mutable. For the [LensPlugin], we've just demonstrated how we
+ * can check for the product type of a [KtClass] by access the properties made available by the Kotlin PSI change.
```
raulraja:
```suggestion
 * can check for the product type of a [KtClass] by accessing the properties made available by the Kotlin PSI tree.
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,150 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ *
+ * While it is not necessary to build [KtClass] predicates for [Meta.classOrObject] does not necessarily have to serve
```
raulraja: I don't understand the meaning of this paragraph. @ahinchman1 Can you please elaborate on what we are trying to illustrate here?, thanks :)

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,150 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ * as a check for functional proofs in your compiler plugin.  It is relevant to highlight that the `match` predicate
+ * written gives access to the Kotlin PSI when analyzing written code via the written compiler plugin.
+ *
+ * The second parameter `map` is a function that allows the resulting action from matching on the transformation at
```
raulraja:
```suggestion
 * The second parameter [map] is a function that allows the resulting action from matching on the transformation at
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,150 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ * the PSI level. Let us example another example plugin:
```
raulraja:
```suggestion
 * the PSI level. The following example plugin illustrates how a Class is intercepted and transformed given `name == "Test" filter. Once matched it's then transformed by [Transform.replace], a transformation that will replace the intercepted class by a new user-declared synthetic replacement. In this example we can observe how the [ClassScope] is available to destructure the class template in full and allows us to reconstruct it back into a [KtClass] by using the [ElementScope.`class`] function.
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,150 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ *                  |    println("Implemented by ΛRROW Meta!")
+ *                  |}
+ *                  |""".`class`.synthetic
+ *           )
+ *         }
+ *       )
+ *     }
+ * ```
+ *
+ * After analyzing the PSI elements available, we pass a resulting [KtClass] matching the predicate (in our case,
```
raulraja: I guess what I wrote above in https://github.com/arrow-kt/arrow-meta/pull/87/files#r344923142 is similar to this so feel free to discard or incorporate whatever you think is worth it

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
ahinchman1: thanks for this catch!

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
ahinchman1: I had a rather messy rebase on this particular one since it's been sitting for a couple days - I definitely have to go back to double check stuff like this!

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
ahinchman1: Good to know - I think I did this originally and then second guessed myself haha

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
ahinchman1: ah yeah, I wrote this horribly.  

```
While it is not necessary to build [KtClass] predicates for [Meta.classOrObject] does not necessarily have to serveas a check for functional proofs in your compiler plugin.
```

What I mean here is that in the example I've provided, we favor and talk about writing compiler plugins for functional stuff, which is great - I wanted to emphasize that folks really don't need to worry about writing predicates (`match`) that are functional proofs. They can be for anything! I think, to me, it is a priority to have an example that we have a compiler plugin that is not functional related as well

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,112 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
 import org.jetbrains.kotlin.psi.psiUtil.modalityModifierType
 import org.jetbrains.kotlin.psi.psiUtil.visibilityModifierType
 
+/**
+ * [classOrObject] is a function that intercepts all [KtClass] elements that [match]
```
JorgeCastilloPrz: nit: Unintended newline?

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,112 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ * allowing a [Transform] to change the intercepted AST tree before compilation.
```
JorgeCastilloPrz:
```suggestion
* then uses a [Transform] to change the intercepted AST tree before compilation.
```
💡 nitpick, but it seems easier to understand for me like this, when you're not aware of what's a `Transform` in the first place.

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,112 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ *
+ * An extension function of [Meta] and inheriting from [ExtensionPhase], [classOrObject] was designed to to feed in
```
JorgeCastilloPrz:
```suggestion
 * An extension function of [Meta] and inheriting from [ExtensionPhase], [classOrObject] was designed to feed in
```
typo!

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,112 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ * virtually any kind of [KtClass] predicate, followed by mapping a function that takes the desired [Scope] of our
```
JorgeCastilloPrz:
```suggestion
 * virtually any kind of [KtClass] predicate, followed by a mapping function that takes the desired [Scope] of our
```
Is this accurate as is? I'm not 100% sure about this one since I don't have full understanding of the codebase yet, but from my point of view it feels like it's a mapping function what follows, instead of "mapping a function". It might also be because of my not very high English level :D

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,112 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ * [KtClass] to change whatever PSI elements desired.
+ *
+ * For example, the [LensPlugin] and the [HigherKindPlugin] favor easy update of immutable data structures and ad-hoc polymorphism that does not require inheritance.
+ *
+ * In the [LensPlugin], we check to see if a type is a product type.  To add context, the cardinality of a product
+ * type is the product of its contained properties:
```
JorgeCastilloPrz: nit: Is this newline intended? to be honest I get a bit lost on where the newlines within paragraphs should be added or whether they shouldn't at all, given they'll break the text in random places depending on how it's rendered on the site. Do we have an strategy for that? Should we go with a one liner for complete paragraphs so the site breaks it at will when it reaches the end? cc @raulraja @pakoito

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,112 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ *
+ * ```
+ * fun isProductType(ktClass: KtClass): Boolean =
+ *      ktClass.isData() &&
+ *        ktClass.primaryConstructorParameters.isNotEmpty() &&
+ *        ktClass.primaryConstructorParameters.all { !it.isMutable } &&
+ *        ktClass.typeParameters.isEmpty()
+ * ```
+ *
+ * For a [KtClass] to be a product type, we check to see if the [KtClass] is an ADT, and that the cardinality of
```
JorgeCastilloPrz: nit: "is an ADT" -> is that accurate? as in: does `isData()` check that or just if it's a `data class` ? i.e: sealed classes and enums are considered sum types hence ADTs if I'm not mistaken. In case `isData()` does not check for those I think I'd update the wording here. Otherwise please ignore me.

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,112 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ * the constructor parameters is greater than one and non-mutable. For the [LensPlugin], we've just demonstrated how we
+ * can check for the product type of a [KtClass] by accessing the properties made available by the Kotlin PSI tree.
+ *
+ * In the example provided, we favor and talk about writing compiler plugins for functional stuff,
```
JorgeCastilloPrz: Nit: We've got many random line breaks in all these paragraphs also.

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,112 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ * which is great - I wanted to emphasize that folks really don't need to worry about writing predicates
```
JorgeCastilloPrz:
```suggestion
 * which is great - We'd want to emphasize that folks really don't need to worry about writing predicates
```
nit: To me it's nice if we talk about "us" as a team in the docs, so there's an overall language coherence and they don't look as if they were written by tons of different devs, even if that's true.

JorgeCastilloPrz:
Added some nitpicks here and there, feel free to address / ignore them at will @ahinchman1. Thanks for the PR!

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,112 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ * (the [match] parameter) that are functional proofs. They can be for any boolean predicate!
+ *
+ * The second parameter [map] is a function that allows the resulting action from matching on the transformation at
+ * the PSI level. The following example plugin illustrates how a Class is intercepted and transformed given `name == "Test" filter. Once matched it's then transformed by [Transform.replace], a transformation that will replace the intercepted class by a new user-declared synthetic replacement. In this example we can observe how the [ClassScope] is available to destructure the class template in full and allows us to reconstruct it back into a [KtClass] by using the [ElementScope.`class`] function.
```
JorgeCastilloPrz:
```suggestion
 * the PSI level. The following example plugin illustrates how a Class is intercepted and transformed given `name == "Test"` filter. Once matched it's then transformed by [Transform.replace], a transformation that will replace the intercepted class by a new user-declared synthetic replacement. In this example we can observe how the [ClassScope] is available to destructure the class template in full and allows us to reconstruct it back into a [KtClass] by using the [ElementScope.`class`] function.
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
ahinchman1: Thanks Jorge! Much appreciated.

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
ahinchman1: @JorgeCastilloPrz should I be letting it go as far as I want? I just press enter when I hit 80 characters out of habit. I don't normally work on docs so I'm not sure

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
ahinchman1: I think you make a really good point - at the very least, this suggests we need to clarify what we mean.

I want to say yes though - I think I talk about it more below with this section of code:

``` ```
 *  val Meta.lenses: Plugin
 *    get() =
 *      "lenses" {
 *        meta {
 *          classOrObject(::isProductOrType) { c -> ... }
 *        }
 *      }
 * ```

You can see the pattern where you match and you do something with the resulting match. 
Since the second parameter of `classOrObject` is 
```map: ClassScope.(KtClass) -> Transform<KtClass>```

We always know we're applying a function to the resulting predicate coming out of the analysis phase

File: compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt (hunk 17 above)
Comment (ahinchman1) [reply compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt]: Let's double check with @raulraja but I believe that the proof behind a product type lays in algebraic data types - the proof doesn't line up exactly with the code, but I think it's close with this idea:

The cardinality of a product type is the product of their cardinalities. 

    |(a, b)| = |a| x |b|

Comment (ahinchman1) [reply compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt]: Maybe I'm being too didactic, but I'm very new about thinking with types! So this could easily be wrong

File: compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt (hunk 19 above)
Comment (ahinchman1) [reply compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt]: Thank you for that catch!!

File: compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt (hunk 20 above)
Comment (raulraja) [reply compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt]: yes, because `class` is a kotlin reserved keyword but that property is called `class`

Comment (ahinchman1) [reply compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt]: @JorgeCastilloPrz No, I don't think they are - but I don't know if Ank files can reference extension functions well.  Would @nomisRev know anything about that?

File: compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt (hunk 12 above)
Comment (JorgeCastilloPrz) [reply compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt]: I'd suggest running the site locally then double check formatting.

File: compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt (hunk 15 above)
Comment (JorgeCastilloPrz) [reply compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt]: Yeah I'd put this as a "mapping function"

File: compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt (hunk 16 above)
Comment (JorgeCastilloPrz) [reply compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt]: I guess just making it look good when running the site.

File: compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt (hunk 17 above)
Comment (JorgeCastilloPrz) [reply compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt]: Yep that's right but I think the wording is wrong, and maybe the implementation of the plugin too. Given its the implementation of `isProductType()`, we need to check for data classes and also any other aggregation types like all the types of tuples. All those are product types. I think the implementation is just checking for data classes here, so it'd be incomplete. Also, the wording in the description shouldn't refer to ADTs overall since that also implies checking for sum types which is not correct for this one.

File: compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt (hunk 16 above)
Comment (ahinchman1) [reply compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt]: I'll be sure to check locally for the next push!

File: compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt (hunk 17 above)
Comment (ahinchman1) [reply compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt]: Thank you so much for the feedback, that's incredibly helpful! but also lol... that's a good point. This is @nomisRev's plugin that we're looking at I think with LensPlugin - perhaps the productType check is modified for a specific purpose for LensPlugin?

File: compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt (hunk 12 above)
Comment (ahinchman1) [reply compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt]: I will do so tonight after Kotlin User Group!

File: compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt (hunk 16 above)
Comment (ahinchman1) [reply compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt]: I have them! Thanks!

File: compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt (hunk 12 above)
Comment (ahinchman1) [reply compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt]: Unable to run it locally - it looks like there are some compatibility issues in the gemfile for jekyll

```
Bundler could not find compatible versions for gem "rubygems":
  In Gemfile:
    rubygems (= 2.5.2.3)

    jekyll (~> 4.0.0) was resolved to 4.0.0, which depends on
      rubygems (>= 2.7.0)
```

File: compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt [hunk 21]
```diff
@@ -13,12 +13,112 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ *
+ * ```
+ * fun isProductType(ktClass: KtClass): Boolean =
+ *      ktClass.isData() &&
+ *        ktClass.primaryConstructorParameters.isNotEmpty() &&
+ *        ktClass.primaryConstructorParameters.all { !it.isMutable } &&
+ *        ktClass.typeParameters.isEmpty()
+ * ```
+ *
+ * For a [KtClass] to be a product type, we check to see if the [KtClass] is an ADT, and that the cardinality of
```
raulraja:
```suggestion
 * For a [KtClass] to be a product type, we check to see if the [KtClass] is an ADT, and that the arity of
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,112 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
+ *     }
+ * ```
+ *
+ * After analyzing the PSI elements available, we pass a resulting [KtClass] matching the predicate (in our case,
+ * we pass the resulting [KtClass] whose name is "Test") and replace the entire object with the string block, which is
+ * then wrapped as a [ClassScope] and wrapped so the `newDeclaration` is of the type [Scope]<[ClassScope]> to match
+ * match compatibility of the intercepted classes wrapped in some kind of [Scope]. To see more, see
+ * [ClassScope].
+ *
+ * @param match designed to to feed in any kind of [KtClass] predicate returning a [Boolean]
```
raulraja:
```suggestion
 * @param match filters [KtClass] elements based on a [Boolean] predicate
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/ClassOrObject.kt
```diff
@@ -13,12 +13,112 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
   map: ClassScope.(KtClass) -> Transform<KtClass>
 ): ExtensionPhase =
   quote(match, map) { ClassScope(it) }
 
+/**
+ * The [ClassScope] is projected over the template to allow destructuring of the different parts of the
+ * [KtClass]. The scope enables template syntax where the user may add new members or modify the class structure
+ * before it's compiled.
+ *
+ * @param value the PSI element being fed in for the compiler plugin.
```
raulraja:
```suggestion
 * @param The scoped [KtElement] being destructured in the template
```

raulraja:
thanks @ahinchman1 !

Issue #89, opened 2019-11-06
Scopes for AST elements for the Quote System


ahinchman1:
We need to come up with a list of the entire set of AST elements and declarations that we want the user to be able to destructure with templates. Right now, we only have scopes for templates for `classOrObject`, `func`, `property` and `type aliases`.

Every element of the `KtElement` hierarchy is worth restructuring with a template, which needs a `Scope` type.  We want to make it easy to build nodes for the AST as quotes. Ideally, we should have one file per quote and that includes the public function and its scope.

ahinchman1:
Asides from the `binary oper` #90 one I'm working on, would it make sense to work with AST elements top-down? By that, I mean we start with `block`, `call expression` and so on

ahinchman1:
@raulraja made an excellent point about the system we create and thought it would make sense to add it to this ticket, from a discussion on the [BinaryOper PR](https://github.com/arrow-kt/arrow-meta/pull/90#discussion_r344511834)

```
I think the name for this quote and all other quotes should be 
whatever the Psi element name is.
I think we should replace:

func -> namedFunction
bynaryOper -> binaryExpression

there are other function matches we can do like lambdas 
and there are other binary expressions that are not operators for example any function flagged as infix
```

PR #90, opened 2019-11-06
`binaryOper` Quote and Template Scope

ahinchman1:
See #89 for referencing, meant to be a place for discussion about AST elements with Scope templating

### DIFF
--- a/compiler-plugin/src/main/kotlin/arrow/meta/quotes/BinaryOper.kt
+++ b/compiler-plugin/src/main/kotlin/arrow/meta/quotes/BinaryOper.kt
@@ -1,0 +1,49 @@
+package arrow.meta.quotes
+
+import arrow.meta.Meta
+import arrow.meta.phases.ExtensionPhase
+import org.jetbrains.kotlin.psi.KtBinaryExpression
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtOperationReferenceExpression
+
+/**
+ * A binary expression with an operator can occur in the following instances:
+ *  - from an expression
+ *  - from an element
+ *  - from an argument
+ *
+ *  Since there are so many ways an operator can be detected, it may be beneficial to drill down
+ *  using a more generalized visitor as the top Meta operation....
+ *
+ *  Below is starting in the wrong place so this will need to start in a more broad area
+ */
+fun Meta.binaryOper(
+  match: KtBinaryExpression.() -> Boolean,
+  map: BinaryExpressionScope.(KtBinaryExpression) -> Transform<KtBinaryExpression>
+): ExtensionPhase =
+  quote(match, map) { BinaryExpressionScope(it) } // How can I drill down into the right scope here?
+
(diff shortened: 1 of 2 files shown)


compiler-plugin/src/main/kotlin/arrow/meta/quotes/BinaryOper.kt
```diff
@@ -0,0 +1,49 @@
...
+
+// TODO this binary expression needs to be drilled down to primaries or objects, so this will have to be filtered
+class BinaryExpressionScope(
+  override val value: KtBinaryExpression,
+  val left: ExpressionScope? = value.left?.let { ExpressionScope(it) },
+  val right: ExpressionScope? = value.right?.let { ExpressionScope(it) },
+  val operationReference: OperationReferenceExpressionScope? = value.operationReference.let { OperationReferenceExpressionScope(it) }
+): Scope<KtBinaryExpression>(value)
+
+class ExpressionScope(
```
raulraja: In this cases you can just use Scope instead of crating this intermediate classes.
If there was something useful to destructure inside then it would have its own quote file.
Scope and ScopedList are there for these cases and when you have lists of scopes like value arguments lists

ahinchman1: It took me a while to answer this one because the first few times I didn't know what this meant - why can I just use the `Scope`? I would need a separate expression scope for each side, right? Are we saying that we want to have `ExpressionScope` as a separate file and use `Scope` in the `BinaryExpressionScope` where the left and right takes place?

ahinchman1: Sorry - nevermind, I found the updated work from recent and I think I have a better idea what to do here!

compiler-plugin/src/main/kotlin/arrow/meta/quotes/BinaryOper.kt
```diff
@@ -0,0 +1,26 @@
...
+import org.jetbrains.kotlin.psi.KtBinaryExpression
+import org.jetbrains.kotlin.psi.KtCallExpression
+import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.KtOperationReferenceExpression
+
+/**
+ * A template destructuring [Scope] for a [KtBinaryExpression]
+ */
+fun Meta.binaryOper(
```
raulraja: I think the name for this quote and all other quotes should be whatever the Psi element name is.
I think we should replace:
```
func -> namedFunction
bynaryOper -> binaryExpression
```
there are other function matches we can do like lambdas and there are other binary expressions that are not operators for example any function flagged as `infix`

raulraja:
```suggestion
fun Meta.binaryExpression(
```

ahinchman1: great! I'll have this fixed, and also add to the AST parsing issue as a note!

PR #92, opened 2019-11-07
Set baseurl param when building the site 

calvellido:
I'm leaving it as a comment for now, but when we start publishing versioned sites of the docs we need to set a specific parameter `baseurl` when building the site.

This value will be the same as the directory where the site will be hosted/published, in our case the name/version of the release: `$MAYOR_MINOR_VERSION`

This will correctly set the paths on the menu entries, assets, and such.


rachelcarmena:
Thanks @calvellido :tada:

PR #93, opened 2019-11-07
Updates Jekyll 4.0.0 and adds error.html

### LINKED ISSUE #91: Set Jekyll 4.0.0
We should change the Jekyll version up to 4.0.0 since we can get the benefit of its improvements, also we need to change sidebar links to relative_url in order to navigate in the proper different versions of Arrow meta.

AntonioMateoGomez:
This PR:
- Updates Jekyll version 4.0.0
- Set `relative_url` sidebar links
- Implements error.html

Fixes #91


calvellido:
@franciscodr on the versioned sites structure, we might be needing to copy/host this `error.html` page on the bucket root, or redirect accordingly.

PR #95, opened 2019-11-07
Fix: use of shadowJar for compiler plugin

rachelcarmena:
## Changes

- Update checks for the use of shadowJar
- `configuration: 'shadow'` for `compiler-plugin` dependencies
- Recover `compile` for `"org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"` to be included in shadowJar
- Update `metaDependencies`
- Remove doc artifact from publication in artifact repository
- Remove duplicated test in testing-plugin.

### DIFF
--- a/compiler-plugin/build.gradle
+++ b/compiler-plugin/build.gradle
@@ -3,13 +3,14 @@ apply plugin: 'kotlin-kapt'
 apply plugin: 'com.github.johnrengelman.shadow'
 
 dependencies {
-    compileOnly "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    // WARNING: don't change the configuration name because it's used by shadowJar
+    compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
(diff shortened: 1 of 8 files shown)


compiler-plugin/build.gradle
```diff
@@ -3,13 +3,13 @@ apply plugin: 'kotlin-kapt'
 apply plugin: 'com.github.johnrengelman.shadow'
 
 dependencies {
-    compileOnly "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
```
i-walker: May I ask, why we had to change it to `comple`?

rachelcarmena: Of course @i-walker , questions are welcome!! Because I changed it to `compileOnly` without looking at `shadowJar` configuration where `compile` dependencies are included:

```
configurations = [project.configurations.compile]
```

i-walker: Meaning the compiler-jar now includes the std lib?

PR #97, opened 2019-11-08
[WIP] TypeClassResolutionTest for code snippets (2)

rachelcarmena:
It comes from #57

New branch from `master` with the following commits:

*  TypeClassResolutionTest for code snippets: #57 without changing `docs` and `testing-plugin`. I changed the author by jansorg.
* Refactor: move testing code to src/test
* Refactor: move idea-plugin to idea directory
* Refactor: idea-plugin: avoid the direct use of `kotlin-compiler-testing` to manage dependencies

**Pending task**: solve errors with tests.


i-walker:
May I ask why the testing dsl is in the test directory. This will lead to complications with the IdeaFramework. The dsl should stay within the main module. If not, JUnit will spin up the IntelliJ Testing Framework for the DSL, even though it is intended as tooling and not as an instance.

i-walker:
May we merge this into the testing branch of the ide not in master?
Thank you @rachelcarmena for helping us.

rachelcarmena:
Thanks for the review @i-walker !! 

This branch is not ready to merge. It was just a test in order to know if there was a difference between using `testing-plugin` or `kotlin-compiler-testing` library. Result: the same problem remains.

I moved testing dsl from `src/main` to `src/test` because it's not provided to another consumer, just for tests.

There is a new commit in this branch so I'm going to close the pull request to open another one with a different content. 

Thanks @i-walker !!

i-walker:
Ok :)
The testing DSL provides the testing environment and tooling for users and that will be a public API.

rachelcarmena:
> The testing DSL provides the testing environment and tooling for users and that will be a public API.

Oh, in that case, it makes sense its location, thanks!!

PR #98, opened 2019-11-08
compiler-plugin: publish twice (normal and shadow jar)

rachelcarmena:
* Add configuration to publish  `compiler-plugin` with `uploadShadow`.
* `compiler-plugin` will be published twice:
   * `io:arrow-kt:arrow-meta-compiler-plugin:<version>`
   * `io:arrow-kt:arrow-meta-compiler-plugin-all:<version>`
* Recover the simple command: `./gradlew clean build`.


.github/workflows/build_artifacts.yml
```diff
@@ -18,17 +18,4 @@ jobs:
...
-        ./gradlew :compile-plugin:shadowJar
-        #
-        # Generate testing-plugin artifact when running compiler-plugin tests
-        ./gradlew :compile-plugin:test
-        #
-        # Run tests from testing-plugin
-        ./gradlew :testing-plugin:test
-        #
-        # Generate idea-plugin artifact
-        ./gradlew :idea-plugin:build
```
ahinchman1: Hi! What is the motivation behind have the other plugin tasks running removed?

.github/workflows/publish_artifacts.yml
```diff
@@ -21,23 +21,15 @@ jobs:
...
     - name: Publish artifacts
       env:
         BINTRAY_USER: ${{ secrets.BINTRAY_USER }}
         BINTRAY_API_KEY: ${{ secrets.BINTRAY_API_KEY }}
       run: |
+        echo "NOTE: gradle-plugin and docs have their own publication workflows" 
+        echo "Uploading idea-plugin and testing-plugin..."
         ./gradlew uploadArchives
+        echo "Uploading compiler-plugin..."
+        ./gradlew :compiler-plugin:uploadShadow
```
ahinchman1: I think it's great we're having a solid logging system being created! <3

.github/workflows/build_artifacts.yml
rachelcarmena: Hi! It was an unnecessary change this morning... :woman_facepalming:

gradle/upload-shadow.gradle
```diff
@@ -0,0 +1,23 @@
...
+apply plugin: 'maven'
+
+uploadShadow {
+    repositories {
+        mavenDeployer {
+            pom.groupId = GROUP
+            pom.artifactId = POM_ARTIFACT_ID
+            pom.version = VERSION_NAME
+
+            if ("$VERSION_NAME".contains("SNAPSHOT"))
```
i-walker: Yeees 🎉 that is an elegant way to distribute our dependencies. Awesome @rachelcarmena

Issue #28, opened 2019-10-01
Purity plugin IDE integration


raulraja:
The original version of the purity plugin checks when a function returns Unit or contains calls in its body to other functions that it return Unit.

This plugin should show visually in the editor this is happening and provide intentions to refactor the function to become `suspend`. This is the IDE piece of the purity plugin.

i-walker:
We should keep that open because we're still working on traversing and resolving the type of each Call of a Function.
This code would not be inspected by the current purity plugin:
```
object Impure {
    fun sideEffect()=
        println("BOOM!")

    suspend fun other(): Unit {
        println("other")
    }

    val x = { println() }

    fun another2(): String {
        1; println("test");
        {
            { println() }
        }()
        return "another"
    }
}
```

PR #99, opened 2019-11-09
Initial purity plugin

i-walker:
Screenshot 2019-11-09 at 21 56 30
Screenshot 2019-11-09 at 21 56 56

This aims to solve https://github.com/arrow-kt/arrow-meta/issues/28
The `InspectionSyntax` allows us to write costume Inspections based on Tooling, what Kotlin already uses in the editor. This allows plugin developers to not only indicate code elements, which need to be inspected but additionally to manipulate them programmatically. 
As here to add a `suspended` keyword to a function. 

`ApplicableInspections` can traverse any KtElement. Furthermore, we can extend the Scope of Inspection to the whole Project instead of individual files. This allows users to write fast fixes, which are universal for the whole project and deal with subsequent issues. 

I'll add checks for calls in the function body, which return Unit or Nothing.
We should also add Text for an explanation.

### DIFF
--- a/idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/inspection/InspectionUtilitySyntax.kt
+++ b/idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/inspection/InspectionUtilitySyntax.kt
@@ -1,8 +1,5 @@
 package arrow.meta.ide.dsl.editor.inspection
 
-import arrow.meta.phases.ExtensionPhase
-import com.intellij.codeInspection.LocalInspectionTool
-import com.intellij.openapi.extensions.ExtensionPointName
 import org.jetbrains.kotlin.builtins.KotlinBuiltIns
 import org.jetbrains.kotlin.descriptors.FunctionDescriptor
 import org.jetbrains.kotlin.idea.util.actualsForExpected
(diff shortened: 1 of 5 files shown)


idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/inspection/InspectionUtilitySyntax.kt
```diff
@@ -19,11 +16,8 @@ interface InspectionUtilitySyntax {
     return listOf(expect) + actuals
   }
 
-  fun InspectionUtilitySyntax.addLocalInspectionToolToIdeRegistry(): ExtensionPhase = TODO("untested")
-  /*registerExtensionPoint(
-    EP_NAME,
-    LocalInspectionTool::class.java
-  )*/
+  val ArrowPath: Array<String>
+    get() = arrayOf("Kotlin", "Λrrow")
```
raulraja:
```suggestion
    arrayOf("Kotlin", "Λrrow")
```

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/dummy/DummyIdePlugin.kt
```diff
@@ -1,37 +1,7 @@
...
-          f.nameIdentifier != null && !f.hasModifier(KtTokens.SUSPEND_KEYWORD) &&
-            f.resolveToDescriptorIfAny()?.run {
-              !isSuspend && (ReturnsCheck.ReturnsUnit.check(this) || ExtendedReturnsCheck.ReturnsNothing.check(this)
-                || ExtendedReturnsCheck.ReturnsNullableNothing.check(this))
-            } == true
-        }
-      )
-    )
-  }
+  get() = TODO()
```
raulraja: We should probably remove the whole file

PR #100, opened 2019-11-09
Parameter quote and template scope

bloderxd:
This PR implements a KtParameter quote with its template, this is related to https://github.com/arrow-kt/arrow-meta/issues/89


compiler-plugin/src/main/kotlin/arrow/meta/quotes/Parameter.kt
```diff
@@ -0,0 +1,33 @@
...
+fun Meta.parameter(
+  match: KtParameter.() -> Boolean,
+  map: ParameterScope.(KtParameter) -> Transform<KtParameter>
+) : ExtensionPhase =
+  quote(match, map) { ParameterScope(it) }
+
+/**
+ * A template destructuring [Scope] for a [KtParameter]
+ */
+class ParameterScope(
```
raulraja: All looks great but this may be missing the default value part of the structure since KtParameters can point to an initializer expression block.

bloderxd: Yeah you're right, now I've mapped it as a `KtExpression`, do you think it's enough to go or need some kind of custom handling (mapping different properties of the default value)

compiler-plugin/src/main/kotlin/arrow/meta/quotes/Parameter.kt
```diff
@@ -0,0 +1,31 @@
...
+  override val value: KtParameter,
+  val modality: Name? = value.modalityModifierType()?.value?.let(Name::identifier),
+  val visibility: Name? = value.visibilityModifierType()?.value?.let(Name::identifier),
+  val name: Name? = value.nameAsName,
+  val `@annotationEntries`: ScopedList<KtAnnotationEntry> = ScopedList(value.annotationEntries),
```
ahinchman1: @raulraja why is `@annotationEntries` in string format? I've seen this else where in the quote system but I was wondering if there is more context about it

raulraja: When structuring a template you want the member to look as close to the parts:
```
`@annotations`
`(valueArguments)`
`(typeArguments)`
```
For example:
```kotlin
"""|$`@annotation` $kind $name $`(typeParameters)` $`(valueParameters)` : $supertypes {
   |  $body
   |}
   |""".`class`.synthetic
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/Parameter.kt
raulraja: If the initializer is of type `KtExpression` wrapping it in a `Scope` is all you need to do as that is an ADT we don't yet cover. It includes Constants, Calls and many others we don't yet have as quotes. We can revisit this one once we have those. Thanks so much for this great contribution @bloderxd

compiler-plugin/src/main/kotlin/arrow/meta/quotes/Parameter.kt
raulraja: symbols like `@` `(` and `)` are not allowed in Kotlin unless quoted. Alternatively, we can offer the same version of those without the quotes and the special symbols and let people use whichever looks more idiomatic for their use case

compiler-plugin/src/main/kotlin/arrow/meta/quotes/Parameter.kt
```diff
@@ -0,0 +1,31 @@
...
+  val type: Scope<KtTypeReference> = Scope(value.typeReference),
+  val `(typeParameters)`: ScopedList<KtTypeParameter> = ScopedList(prefix = "<", value = value.typeParameters, postfix = ">"),
+  val defaultValue: KtExpression? = value.defaultValue
```
ahinchman1: @bloderxd I think we're missing some items here.  The way to think about AST elements is by the elements themselves. For example, for a parameter AST element, we might be interested in `hasValOrVar` and also `isVarArg`.

@raulraja I'm a little unsure about `KtParameter::getValueParameters` and so on - I would say we should add of lot of these other options as well, but it looks like it's simply not implemented in the class since a lot of them return `null` or `emptyList()`

compiler-plugin/src/main/kotlin/arrow/meta/quotes/Parameter.kt
ahinchman1: @bloderxd I think that we actually need to remove `@annotationEntries` as well. This is not an element that is part of `KtParameter`.  The same goes for `modality` and `visibility`

compiler-plugin/src/main/kotlin/arrow/meta/quotes/Parameter.kt
i-walker: @ahinchman1 functions like `hasValOrVar`, `isVarArg` are already provided by  KtParameter. It would add little as we have `KtParameter` in Scope at `match`.

compiler-plugin/src/main/kotlin/arrow/meta/quotes/Parameter.kt
raulraja: @ahinchman1 `KtParameter` inherits `KtAnnotated` which is in most elements that receive annotations including parameters

PR #104, opened 2019-11-11
Inspection cleanUp

i-walker:
A general cleanUp for the editor fixes unnecessary NPE with `Inspections`, boosts performance for LineMarker Resolution and cleansUp `InspectionSyntax`.

### DIFF
--- a/idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/purity/PurityPlugin.kt
+++ b/idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/purity/PurityPlugin.kt
@@ -6,10 +6,12 @@ import arrow.meta.ide.dsl.editor.inspection.ExtendedReturnsCheck
 import arrow.meta.invoke
 import com.intellij.codeHighlighting.HighlightDisplayLevel
 import com.intellij.codeInspection.ProblemHighlightType
+import org.jetbrains.kotlin.codegen.coroutines.isSuspendLambdaOrLocalFunction
 import org.jetbrains.kotlin.idea.caches.resolve.resolveToDescriptorIfAny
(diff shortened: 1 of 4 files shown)


idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/purity/PurityPlugin.kt
```diff
@@ -27,7 +29,7 @@ val IdeMetaPlugin.purity: Plugin
         isApplicable = { f: KtNamedFunction ->
           f.nameIdentifier != null && !f.hasModifier(KtTokens.SUSPEND_KEYWORD) &&
             f.resolveToDescriptorIfAny()?.run {
-              !isSuspend && (ReturnsCheck.ReturnsUnit.check(this) || ExtendedReturnsCheck.ReturnsNothing.check(this)
+              !isSuspend && !isSynthesized && !isSuspendLambdaOrLocalFunction() && (ReturnsCheck.ReturnsUnit.check(this) || ExtendedReturnsCheck.ReturnsNothing.check(this)
```
i-walker: @raulraja Should we also make a distinction for `get`, `set`, `init` functions?  As we did back then 
https://github.com/arrow-kt/arrow-meta-prototype/blob/32554a5ade4b49737cfe7ed1e5112415ea84be02/compiler-plugin/src/main/java/arrow/plugin/MetaClassBuilderInterceptorExtension.kt#L54

raulraja: I think so

i-walker: I am not sure either, let's revisit this another time.

PR #108, opened 2019-11-11
`nullableType` quote and scope

ahinchman1:
### Checklist for nullableType
 - [x] Added/replaced the inverse element in `ElementScope`

### DIFF
--- a/compiler-plugin/src/main/kotlin/arrow/meta/phases/analysis/DefaultElementScope.kt
+++ b/compiler-plugin/src/main/kotlin/arrow/meta/phases/analysis/DefaultElementScope.kt
@@ -1,6 +1,12 @@
 package arrow.meta.phases.analysis
 
-import arrow.meta.quotes.*
+import arrow.meta.quotes.ClassScope
+import arrow.meta.quotes.ForExpressionScope
+import arrow.meta.quotes.NamedFunctionScope
+import arrow.meta.quotes.NullableTypeScope
+import arrow.meta.quotes.ParameterScope
+import arrow.meta.quotes.Scope
+import arrow.meta.quotes.WhileExpressionScope
 import org.jetbrains.kotlin.com.intellij.openapi.project.Project
 import org.jetbrains.kotlin.com.intellij.psi.PsiComment
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement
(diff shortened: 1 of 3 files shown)


raulraja:
Can we add to the docs examples of how this may be used? thanks!

ahinchman1:
Ah, yeah I definitely can - to be honest, I'm still trying to figure out what exactly this one is - I spent a little time searching but I can pick that back up. 

> Can we add to the docs examples of how this may be used? thanks!

ahinchman1:
After some clarification from @raulraja, we're going to go ahead and merge this in to master, and we'll come back to make a task to add documentation to this items

compiler-plugin/src/main/kotlin/arrow/meta/phases/analysis/DefaultElementScope.kt
```diff
@@ -55,6 +60,9 @@ class DefaultElementScope(project: Project) : ElementScope {
   override val String.typeOrNull: Scope<KtTypeReference>
     get() = Scope(delegate.createTypeIfPossible(trimMargin()))
 
+  override val String.nullableType: NullableTypeScope
+    get() = TODO("not implemented") //To change initializer of created properties use File | Settings | File Templates.
```
ahinchman1: @raulraja @bloderxd There appears to be no 1:1 match in the PsiElementFactory (the delegate) that creates a `KtNullableType`. There are delegate methods that create very close types (see the ones surrounding this bit of code - I wonder if it makes sense to remove this from ElementScope then.

I briefly looked around for uses in the kotlin codebase and I think it's maybe like not a "top-level" element but it may be good to still have this:

Screen Shot 2019-11-13 at 11 47 54 AM

Please, let me know what ya'll think!

raulraja: As far as  I understand a KtNullableType is an instance of a KtTypeElement which is extra info on a KtTypeReference. What you render as a type is the KtTypeReference so I'm not sure we need a template for KtNulllableType but we need one for KtTypeReference and KtElementType which is the reference.

ahinchman1: We can close this one then!

PR #109, opened 2019-11-11
RunLineMarkerSupport

i-walker:
We can draft the `AnkPlugin`based on an Action, which we can wrap into a `RunLineMarker`.
The RunLineMarker will execute the underlying action.


raulraja:
what does this look like and does it come with any docs or example as to how this may be used in a plugin?

i-walker:
With that, we reduced writing a RunLinemarker to writing anAction. 
`AnActionSyntax` has functions to create Actions, but I didn't explore them intensively. Because they are quite extensive on what they can do.
An AnkPlugin draft may look like this:
```
val IdeMetaPlugin.ankPlugin: Plugin
  get() = "AnkPlugin" {
    meta(
      addRunLineMarkerProvider(
        action = addAnAction(
          icon = ArrowIcons.ANK,
          title = "RunAnk",
          description = "Runs Ank Docs",
          actionPerformed = TODO(),
          update = TODO()
        ),
        transform = {it.safeAs<KDocElement>()?.text?.contains("kotlin:ank:")}
      )
    )
  }
```

PR #114, opened 2019-11-12
Purity revisited

i-walker:
This extends the PurityPlugin to Properties and traverses each CallElement of Functions or Properties like a tree.
Currently, the traversal is of depth 1 but can be extended to n.

Screenshot 2019-11-12 at 13 14 59

What can we do to make the Property Suspended? @JorgeCastilloPrz @raulraja @nomisRev


nomisRev:
Nothing, because the parameter is pure. The lambda assigned to it is not, so the lambda should be suspended.

So that it becomes `val: X = suspend {...}`

JorgeCastilloPrz:
Properties are always pure if they reference a lambda. It's a reference to a function still to be run, so in a way it'd be equivalent to IO as is. Still, if we want to follow the line of not allowing to run effects at all in unsafe environments then the lambda itself should be flagged as suspend.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/purity/PurityPlugin.kt
```diff
@@ -29,12 +33,44 @@ val IdeMetaPlugin.purity: Plugin
...
+            modify(body) { b: KtExpression ->
+              b.unwrapBlockOrParenthesis()
+              // createExpressionByPattern("suspend { $0 }", b)
+            }
+          }
+        },
+        isApplicable = { prop: KtProperty ->
+          prop.resolveToDescriptorIfAny()?.run {
+            !returns(
+              f = { t: KotlinType -> t.constructor },
```
i-walker: This defines Type equality when the TypeConstructor is the same

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/inspection/InspectionUtilitySyntax.kt
```diff
@@ -16,15 +30,91 @@ interface InspectionUtilitySyntax {
     return listOf(expect) + actuals
   }
 
+  /**
+   * convenience function where [f] reduces FunctionReturnTypes of subsequent calls in the initializer to their returnType
+   */
+  fun <F : CallableDescriptor> F.returns(prop: KtProperty, types: KotlinBuiltIns.() -> List<KotlinType>): Boolean =
```
raulraja: This probably belongs in the KtUtils in the compiler-plugin module and not here unless these are IDE specific in their types and not accessible from there

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/inspection/InspectionUtilitySyntax.kt
```diff
@@ -16,15 +30,91 @@ interface InspectionUtilitySyntax {
...
+    returns({ if (it.isBuiltinFunctionalType) it.getReturnTypeFromFunctionType() else it }, prop, types)
+
+  /**
+   * convenience function where [f] reduces FunctionReturnTypes of subsequent calls in the body to their returnType
+   */
+  fun <F : CallableDescriptor> F.returns(ktFunction: KtNamedFunction, types: KotlinBuiltIns.() -> List<KotlinType>): Boolean =
```
raulraja: Same as above

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/utils/IdeUtils.kt
```diff
@@ -1,16 +1,45 @@
...
+/**
+ * Modify [element] based on [f]
+ */
+inline fun <reified K : KtElement> KtPsiFactory.modify(element: K, noinline f: KtPsiFactory.(K) -> K?): K? =
+  f(this, element)?.run { element.replaceK(this) }
+
+/**
+ * reified PsiElement replacement
+ */
+inline fun <reified K : PsiElement> K.replaceK(to: K): K? =
```
raulraja: what does this do that astReplace or PsiElement replace doesn't already provide? Is it just the null check?

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/utils/IdeUtils.kt
```diff
@@ -1,16 +1,45 @@
...
+  replace(to).safeAs()
+
+/**
+ * clears null values within List
+ */
 fun <A> List<A?>.toNotNullable(): List<A> = fold(emptyList()) { acc: List<A>, r: A? -> if (r != null) acc + r else acc }
```
raulraja: This already exists in the std lib.
```kotlin
val x: List<Int?> = listOf(1, 2, null)
val y: List<Int> = x.filterNotNull()
```

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/utils/IdeUtils.kt
```diff
@@ -1,16 +1,45 @@
...
 
+/**
+ * General helpers for HTML
+ */
 fun <A> kotlin(a: A): InlineElement = code(other = mapOf("lang" to "kotlin")) { "\t$a\n" }
 fun kotlin(a: String): InlineElement = code(other = mapOf("lang" to "kotlin")) { "\t${text(a).content}\n" }
 fun <A> h1(a: A): BlockElement = org.celtric.kotlin.html.h1("$a")
```
raulraja: Weren't we gonna depend on https://github.com/Kotlin/kotlinx.html for these?

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/purity/PurityPlugin.kt
```diff
@@ -2,39 +2,73 @@ package arrow.meta.ide.plugins.purity
...
 import org.jetbrains.kotlin.resolve.calls.tower.isSynthesized
-import org.jetbrains.kotlin.util.ReturnsCheck
+import org.jetbrains.kotlin.types.KotlinType
 
 val IdeMetaPlugin.purity: Plugin
   get() = "PurityPlugin" {
     meta(
       addApplicableInspection(
-        defaultFixText = "Suspend",
+        defaultFixText = "SuspendFunction",
```
raulraja: We need a table with all plugins and their text so we can iterate and apply it to all of them. A collaborative doc where we can decide what goes on each plugin and how we can best present the info. Pics of what it looks like now on each plugin in the column or with some Lorem ipsum placeholder text would help to asses that. This is an important part of the user experience and we have to nail it not just for this plugin but in Meta in general.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/purity/PurityPlugin.kt
raulraja: I don't understand this API. Do you mind providing more context on how this works?

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/purity/PurityPlugin.kt
```diff
@@ -2,39 +2,73 @@ package arrow.meta.ide.plugins.purity
...
+        },
+        level = HighlightDisplayLevel.ERROR,
+        groupPath = ArrowPath + "PurityPlugin"
       )
     )
-  }
\ No newline at end of file
+  }
+
+private val KotlinBuiltIns.impureTypes: List<KotlinType>
```
raulraja: This is probably extremely slow given these properties are continuously computed and it looks like they could just be a `val` in the plugin that get initialized the moment you can access the `KotlinBuiltIns`

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/utils/IdeUtils.kt
i-walker: It's not from `kotlinx.html` but from https://github.com/celtric/kotlin-html.
Should we clear that? Or change this to another dependency. I thought we could use a library with small binaries and fast string resolution and `kotlinx.html` fitted neither.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/purity/PurityPlugin.kt
i-walker: Should we move this to an actual Doc Part in the plugins directory?
Or should the Text still stay here?
We can curate the table in an issue at first as we're doing it here #131 , but code seems more maintainable. In fact, I am adding the necessary lines to it.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/utils/IdeUtils.kt
i-walker: No, I needed a reified version of that function in order to compose them.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/purity/PurityPlugin.kt
i-walker: How can we improve that ?

i-walker:
I cleanedUp the PR and changed the API, so it uses the Eq typeclass.
Should I remove the checks on KtProperties until we have a refactoring strategy?

i-walker:
It depends on whether one needs the inspection in testing or not.

danimontoya:
Mmm okay good to know because I might need to do the same when I start testing the coercions 👍

PR #122, opened 2019-11-13
Lambda Expression Quote and Template Scope

ahinchman1:
### Checklist for `KtLambdaExpression`


raulraja:
thanks @ahinchman1 !

PR #123, opened 2019-11-13
Function Literal Quote and Template Scope

ahinchman1:
### Checklist for `KtFunctionLiteral`
 - [ ] Destructure and make available all methods related to rendering properties, but no logic


ahinchman1:
PsiFactory doesn't seem to have rendering for function literal - I suppose we can leave this off (the second check in the checklist)

bloderxd:
I believe it makes sense we update the `Scope<KtLambdaExpression>` dependency of the ElementScope to the new lambda expression scope, what do you think?

ahinchman1:
> I believe it makes sense we update the `Scope<KtLambdaExpression>` dependency of the ElementScope to the new lambda expression scope, what do you think?

@bloderxd it's already in there!  From now on I'll be sure to comment in the PR if it's already included in `ElementScope`

```  
fun lambdaExpression(
    parameters: String,
    body: String
  ): Scope<KtLambdaExpression>
```

PR #125, opened 2019-11-13
CI: differentiate Arrow version from Arrow Meta version

rachelcarmena:
## Reason

This is necessary because of the next Arrow release. 

We are using just a single version for both Arrow and Arrow Meta.

## Changes

- Add an additional version in `gradle.properties`: 
```
ARROW_VERSION=0.10.3-SNAPSHOT
```
- Change **Arrow Meta version** to `1.3.70-SNAPSHOT` because of [Kotlin releases](https://github.com/JetBrains/kotlin/releases).
- Dependencies like `arrow-annotations` and `arrow-ank` have the **Arrow version** now
- CI checks cannot do `./gradlew clean build` because it would fail when **release time** (new version): 
    - **Gradle Plugin** version makes reference to **Arrow Meta Compiler Plugin** published version.
- CI checks have a single file to publish artifacts in order to manage the previous dependence.

## Other version changes

- Remove JitPack repository because `kotlin-compile-testing` has been deployed
- Update `kotlin-compile-testing` dependency


gradle.properties
```diff
@@ -1,6 +1,7 @@
 # Package definitions
 GROUP=io.arrow-kt
-VERSION_NAME=0.10.3-SNAPSHOT
+VERSION_NAME=1.3.70-SNAPSHOT
```
i-walker: Shouldn't this be 1.3.60-SNAPSHOT, unless we're jumping versions directly?

rachelcarmena: We start with this version though it will change very soon when sync is ready.

rachelcarmena: Thanks for the review @i-walker!!! :+1:

i-walker: I removed my comment after I read the conv. in Slack. My bad :)

rachelcarmena: Don't worry @i-walker , thanks!! :+1:

PR #127, opened 2019-11-13
Type element quote and template scope

ahinchman1:
### Checklist for `KtTypeElement`

No documentation added for now but will open an issue for folks to contribute.


raulraja:
@ahinchman1 is this ready for merge or does it need the Scope for TypeElement?

ahinchman1:
@raulraja this may be merged in (after fixing the CI and updating the branch) - we have a growing system of "tech debt" where we agree to merge in an element at at a type but when as we create elements, we will go back to update the scope of others

PR #132, opened 2019-11-13
Type reference quote and template scope

ahinchman1:
### Checklist for `KtTypeReference`


compiler-plugin/src/main/kotlin/arrow/meta/quotes/TypeAlias.kt
```diff
@@ -27,7 +27,7 @@ fun Meta.typeAlias(
  */
 class TypeAlias(
   override val value: KtTypeAlias,
-  val `@annotations`: ScopedList<KtAnnotationEntry> = ScopedList(value.annotationEntries),
+  val `@annotationEntries`: ScopedList<KtAnnotationEntry> = ScopedList(value.annotationEntries),
```
raulraja: I'm not sure what happened here but we want to preserve the name annotations vs annotationEntries.

PR #133, opened 2019-11-13
ObjectDeclaration and refactoring ClassOrObjectScope

ahinchman1:
... so that it can be parent to both class and objectDeclaration.

### Checklist for `KtObjectDeclaration`

### DIFF
--- a/compiler-plugin/src/main/kotlin/arrow/meta/quotes/parentscopes/ClassOrObjectScope.kt
+++ b/compiler-plugin/src/main/kotlin/arrow/meta/quotes/parentscopes/ClassOrObjectScope.kt
@@ -3,7 +3,7 @@ package arrow.meta.quotes.parentscopes
 import arrow.meta.quotes.ClassBodyScope
 import arrow.meta.quotes.Scope
 import arrow.meta.quotes.ScopedList
-import com.intellij.navigation.ItemPresentation
+import org.jetbrains.kotlin.com.intellij.navigation.ItemPresentation
 import org.jetbrains.kotlin.name.Name
 import org.jetbrains.kotlin.psi.KtAnnotationEntry
 import org.jetbrains.kotlin.psi.KtAnonymousInitializer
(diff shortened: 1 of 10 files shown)


compiler-plugin/src/main/kotlin/arrow/meta/quotes/parentscopes/ClassOrObjectScope.kt
```diff
@@ -0,0 +1,51 @@
...
+  val modality: Name? = value.modalityModifierType()?.value?.let(Name::identifier),
+  val superTypeList: KtSuperTypeList? = value.getSuperTypeList(),
+  val superTypeListEntries: ScopedList<KtSuperTypeListEntry> = ScopedList(prefix = "<", value = value.superTypeListEntries, postfix = ">"),
+  val anonymousInitializers: ScopedList<KtAnonymousInitializer> = ScopedList(value = value.getAnonymousInitializers(), postfix = ","),
+  val body: ClassBodyScope = ClassBodyScope(value.body),
+  val declarations: ScopedList<KtDeclaration> = ScopedList(value = value.declarations, postfix = ", "),
+  val presentation: ItemPresentation = value.presentation,
+  val primaryConstructor: KtPrimaryConstructor? = value.primaryConstructor,
+  val primaryConstructorModifierList: KtModifierList? = primaryConstructor?.modifierList,
+  val primaryConstructorParameterList: KtParameterList? = primaryConstructor?.valueParameterList,
```
ahinchman1: @raulraja @bloderxd What should we be doing with KtElementList types?

The parameter list doesn't seem to matter as much, but `KtModifierList` has methods like:
```
    @NotNull
    public List<KtAnnotation> getAnnotations() {
        return getStubOrPsiChildrenAsList(KtStubElementTypes.ANNOTATION);
    }

    @NotNull
    public List<KtAnnotationEntry> getAnnotationEntries() {
        return KtPsiUtilKt.collectAnnotationEntriesFromStubOrPsi(this);
    }
```

ahinchman1: @bloderxd and myself explored all this and we found that much of this is unnecessary boxing. Changing it to all use `@annotation` for the type `KtAnnotationEntry` and for all lists of KtAnnotation, `ScopedList<KtAnnotationEntry>`

PR #139, opened 2019-11-14
When entry quote and scope

bloderxd:
This PR introduces `KtWhenEntry` quote and its scope mapping. This PR is related to https://github.com/arrow-kt/arrow-meta/issues/89

### Checklist for `KtWhenEntry`

### DIFF
--- a/compiler-plugin/src/main/kotlin/arrow/meta/quotes/WhenEntry.kt
+++ b/compiler-plugin/src/main/kotlin/arrow/meta/quotes/WhenEntry.kt
@@ -1,0 +1,50 @@
+package arrow.meta.quotes
+
+import arrow.meta.Meta
+import arrow.meta.phases.ExtensionPhase
+import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtWhenCondition
+import org.jetbrains.kotlin.psi.KtWhenEntry
+
+/**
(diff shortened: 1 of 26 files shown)


compiler-plugin/src/main/kotlin/arrow/meta/quotes/WhenEntry.kt
```diff
@@ -0,0 +1,50 @@
...
+import arrow.meta.Meta
+import arrow.meta.phases.ExtensionPhase
+import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtWhenCondition
+import org.jetbrains.kotlin.psi.KtWhenEntry
+
+/**
+ * A [KtWhenEntry] [Quote] with a custom template destructuring [WhenEntryScope]. See below:
+ *
+ * ```kotlin
```
raulraja:
```suggestion
 * ```kotlin:ank:silent
```

raulraja: This is so it type-checks the snippets, otherwise the code would evolve separately from the docs as nobody goes back to change example unless they break the build

bloderxd: Thank you for the explanation @raulraja , I've fixed that in the other PRs too

PR #143, opened 2019-11-14
cleanUp in FragmentProviderComponent

i-walker:


idea-plugin/src/main/kotlin/arrow/meta/ide/phases/resolve/FragmentProviderComponent.kt
```diff
@@ -32,15 +33,19 @@ class FragmentProviderComponent(val project: Project) : ProjectComponent, AsyncF
...
-      e.isValid && file != null && file.extension == "class" && VfsUtilCore.isUnder(file, buildFolders)
+  /**
+   *  we only care about changes to .class files inside of build folders
+   */
+  override fun prepareChange(events: MutableList<out VFileEvent>): FragmentProviderComponent? =
+    this.takeIf {
+      project.run {
+        events.any { e: VFileEvent ->
+          e.file?.let { file: VirtualFile ->
+            e.isValid && file.extension == "class" && VfsUtilCore.isUnder(file, buildFolders().toHashSet())
```
jansorg: Could buildFolder() be cached? In a worst-case scenario buildFolders() would be called `n` times for `n` events here . `prepareChange()` is called for bulk processing and buildFolders() isn't using cached data itself.

i-walker: makes sense :) 
Let me change that

PR #146, opened 2019-11-15
testing-plugin: remove wrapper and add documentation

rachelcarmena:
* [x] Remove wrapper
* [x] Add documentation

It also changes the entry point for Testing Plugin to `assertThis` instead of `CompilerTest`.


raulraja:
We seem to have a lot of warnings related to compiler-test including older versions of Kotlin transitively.. I think we should exclude those because it can interfere with meta use of the compiler.

PR #147, opened 2019-11-15
missing pice in HKT

i-walker:
We were missing the diagnostic suppressor in hkt plugin


i-walker:
It's already added. It was somewhere else. 
Closing this :)

Issue #161, opened 2019-11-16
Scopes of KtElements are Traversable Functors


raulraja:
Add the Traversable Functors API to scope and it's elements. This will allow us to fold and transform scopes in templates with a known API.

raulraja:
A `Scope<K>` is currently constrained to `K: KtElement`.
The purpose of this ticket is to add a new interface that exposes the combinators syntax over the scope.
It may look something like:

```kotlin
interface ScopeSyntax {
  fun <A: KtElement, B: KtElement> Scope<A>.map(f: (A) -> B): Scope<B> //ressembles arrow.Functor.map
  //... include other functions from Foldable and Traverse that are possible to implement
}

interface ElementScope : ScopeSyntax
```

This will allow us to do things such as :
```kotlin
"""
  ${body.map { it.replaceExpression() }}
"""
```

PR #173, opened 2019-11-18
[WIP] Add Functor API to ElementScope

pt2121:
#### WIP

re #161 

```
                    """println("Hello ΛRROW Meta!")"""
                                        .expression
                                        .map { exp ->
                                            """ fun ${this.name}(): Unit = ${exp.text}""".function.value
                                        }.synthetic
```

```
                    """private""".modifierList
                                        .fold(
                                                """ |fun helloWorld(): Unit =
                                                |  println("Hello ΛRROW Meta!")
                                                |""".function.value
                                        ) { acc, mod ->
                                            """ ${mod.text} fun ${acc.name}(): Unit =
                                                ${acc.bodyExpression?.text.orEmpty()}
                                            """.trimIndent().function.value
                                        }.scope()
                                        .synthetic
```

### DIFF
--- a/compiler-plugin/src/main/kotlin/arrow/meta/phases/analysis/DefaultElementScope.kt
+++ b/compiler-plugin/src/main/kotlin/arrow/meta/phases/analysis/DefaultElementScope.kt
@@ -113,6 +113,21 @@ class DefaultElementScope(project: Project) : ElementScope {
       } else this
     }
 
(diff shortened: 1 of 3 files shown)


raulraja:
This is going in the right direction now we need to explore up the hierarchy to see how many more function in Applicative, Monad , Foldable etc we can build

PR #184, opened 2019-11-20
File quote and template scope

ahinchman1:
### Checklist for `KtFile`


raulraja:
thanks @ahinchman1 !

PR #186, opened 2019-11-20
Modifier list quote and scope

bloderxd:
This PR introduces `KtModifierList` quote and its scope mapping. This PR is related to #89 

### Checklist for `KtModifierList`

### DIFF
--- a/compiler-plugin/src/main/kotlin/arrow/meta/quotes/ModifierList.kt
+++ b/compiler-plugin/src/main/kotlin/arrow/meta/quotes/ModifierList.kt
@@ -31,7 +31,7 @@ import org.jetbrains.kotlin.psi.psiUtil.visibilityModifier
  *  }
  *```
  *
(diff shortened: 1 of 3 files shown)


compiler-plugin/src/main/kotlin/arrow/meta/quotes/ModifierList.kt
```diff
@@ -0,0 +1,50 @@
...
+ *     Transform.replace(
+ *      replacing = l,
+ *      newDeclaration = """ $`@annotations` $modifier value """.`modifierList`
+ *     )
+ *    }
+ *   )
+ *  }
+ *```
+ *
+ * * @param match designed to to feed in any kind of [KtModifierList] predicate returning a [Boolean]
```
ahinchman1: tiny nitpicky thing here - there's an extra * !

bloderxd: Oh thank you to notice that, I'll remove.

Issue #131, opened 2019-11-13
Arrow LineMarker Text


i-walker:
As the Test's for the existing LM is about to be merged in #129, we may discuss what Text or GoTo actions we want to implement for `Meta`, wether Users click on the Icon, which then directs them on their default browser to Doc's and tutorials and the Text, which is portrayed. 

Generally speaking, LineMarkers have in most cases short messages. If we do want to present more such as we do in the PolyLineMarker, we might as well invest in building API's in the editor, which opens up TextDisplays on the right-hand side with Content we want to communicate, similar to the Toolbar on the right-hand side for Gradle. The same API could also be used for the `AnkIdePlugin` to render the Docs.
cc @raulraja @franciscodr @nomisRev

raulraja:
We need a table that looks like this for all plugins and involve several people to fill this in.

| Plugin | Action     | Icon       | Description                | Expanded Description                                    |
|--------|------------|------------|----------------------------|---------------------------------------------------------|
| Purity | Inspection | [pic here] | This function is impure... | This function is impure..., Docs to IO, suspend, Fx etc |
|        |            |            |                            |                                                         |

i-walker:
| Plugin | Feature | Icon | Description | Expanded Description  |
|---|---|---|---|---|
|Comprehensions | LineMarker | [`arrow-meta-bind.svg`](https://github.com/arrow-kt/arrow-meta/blob/master/idea-plugin/src/main/resources/arrow/meta/plugin/idea/icons/arrow-meta-bind.svg)| "Bind" | -
|Nothing | LineMarker | [`arrow-meta-nothing.svg`](https://github.com/arrow-kt/arrow-meta/blob/master/idea-plugin/src/main/resources/arrow/meta/plugin/idea/icons/arrow-meta-nothing.svg)| "Bottom Type" | -
|HigherKinds | LineMarker | [`arrow-meta-higher.svg`](https://github.com/arrow-kt/arrow-meta/blob/master/idea-plugin/src/main/resources/arrow/meta/plugin/idea/icons/arrow-meta-higher.svg) |  ${class.name} is a Higher Kinded Type that may be used in polymorphic functions expressed over [Kind<F, A>] and with the type classes. For more info visit [https://arrow-kt.io/docs/patterns/glossary/#type-constructors](https://arrow-kt.io/docs/patterns/glossary/#type-constructors) | - 
[...]

PR #201, opened 2019-11-20
draft for QuickDocsDsL

i-walker:
Regarding examples for https://github.com/arrow-kt/arrow-meta/issues/131
This may be one example: 
```
addDocumentationProvider(
        quickNavigateInfo = { element: PsiElement, originalElement: PsiElement ->  
          element.safeAs<KtUserType>()?.takeIf { it.referencedName == "Nothing" }?.run { 
            "LoremImpsum Docs"
          }
        },
        generateDoc = { element: PsiElement, originalElement: PsiElement ->
          element.safeAs<KtUserType>()?.takeIf { it.referencedName == "Nothing" }?.run {
            "LoremImpsum Docs Longer Text"
          }
        }
      )
```
The `KotlinQuickDocumentationProvider` already abstracts that over KDocs and the implementation is extensive. Personally, I would look for another option rather than competing with that implementation.


raulraja:
Can we see a visual example of how this uses the KDoc?

i-walker:
visuals are here https://github.com/arrow-kt/arrow-meta/issues/131#issuecomment-556060870

raulraja:
Thanks @i-walker !

Issue #198, opened 2019-11-20
Arrow Meta Quotes Table


ahinchman1:
Create a table of contents for available AST quotes and its use examples.

See:
- https://scalameta.org/docs/trees/quasiquotes.html
- https://github.com/scalameta/scalameta/blob/master/notes/quasiquotes.md

raulraja:
Just discussed with @franciscodr how this can be integrated into the site and what needs to happen on each function so the table renders nicely. Please coordinate with him for some hints to finish the impl. thanks!

ahinchman1:
Thank you! Will do! I was just about to reach out to @franciscodr anyway!

Issue #130, opened 2019-11-13
Add documentation to explain AST elements in their custom scopes


ahinchman1:
We're building out the quote system with all as many AST elements as we can, but we will need help to go back to explain examples of what each AST element is in their custom scopes.

**Potential helpful sources:**

- the corresponding PsiElements Kotlin compiler code - see `com.kotlin.psi` package.
- the PSIViewer IntelliJ plugin can be helpful for discovering PSI elements and their associated class:
<img width="830" alt="Screen Shot 2019-11-13 at 3 46 09 PM" src="https://user-images.githubusercontent.com/6917498/68807094-ce701100-062c-11ea-9a93-0be4beceb6f7.png">

- Kotlin Academy is sometimes helpful, especially in their segment for Kotlin Programmer Dictionary. For example, one blog covers [Function Type vs Function literal vs Lambda expression vs Anonymous function](https://blog.kotlin-academy.com/kotlin-programmer-dictionary-function-type-vs-function-literal-vs-lambda-expression-vs-anonymous-edc97e8873e)

ahinchman1:
You can find which elements might need documentation by filtering PRs for `AST-scope` and does not have the `documentation` label.

![Screen Shot 2019-11-13 at 3 41 39 PM](https://user-images.githubusercontent.com/6917498/68806807-38d48180-062c-11ea-80eb-964174803a8b.png)

raulraja:
Duplicate of #198

ahinchman1:
@raulraja I apologize, the purpose of this ticket was not initially clear - while we are placing declarations for single lines for meta extensions similar to Scala's quasi-quote system, I still would like for the AST element scopes themselves to further explain what they are in Kotlin. The AST system is sometimes confusing, so I was thinking this is a nice-to-have and a great issue for first timers learning about our quote system/AST/the Kotlin compiler.  For example:

```/**
 * A template destructuring [Scope] for a [KtFunctionLiteral]
 *
 *  A function literal is a special notation to simplify how a function is defined.  There are two types of function literals in Kotlin:
 *  #### Lambda expression
 *  A lambda expression is a short way to define a function.  It tends to be more explicit than anonymous functions:
 * ```kotlin:ank:silent
 * val increment: (Int) -> Unit = { x -> x + 1 }
 * ```
 *  #### Anonymous function
 *  An anonynous function is just another way to define a function:
 * ```kotlin:ank:silent
[...]

PR #202, opened 2019-11-20
Scope structure refactor & Arrow Meta Quotes Table for declarations & dokka upgrade

ahinchman1:
#198 

Whew! This is a giant PR! Some notes about this PR: 

- There are only basic examples available for the time being. Because we're missing exposure for logic and PsiElements within PSIElements, we cannot create PsiElements with total freedom. For example, we do not have access to val/var in `Property`. As a result, in future work, we will need to work at exposing some of these special items across the board for `File`, `TypeAlias`, `NamedFunction`, and so on.  A note has been left in these elements for now.

- There is a massive refactor on the structure of the scopes to give a better idea about the relation of PSI hierarchy existing in Kotlin.

- Because dokka autogenerated tables cannot format markdown for code within their indexes, we have resorted to creating a table that allows you to click on the element and drill down to see specific examples. 

- Upgraded dokka from `0.9.18` to `0.10.0`

i.e.

Screen Shot 2019-11-25 at 10 25 30 PM

Screen Shot 2019-11-25 at 10 25 43 PM

## Validating Quote Scope usage

Quotes needs to be useable, so we will need to validate via testing to verify the elements are correctly being evaluated:

Captura de Tela 2019-11-27 às 02 43 32

If the elements are not supported by Kastree, we will need to see if Converter provides a function for those kinds of trees and implement it ourselves like we have done with classes, functions, and expressions.

See failing test: #259
See elements needed to be implemented with Converter: #260

**ClassOrObjects**
- [x] ObjectDeclaration

**Elements**
- [x] DestructuringDeclaration
- [x] WhenCondition
- [x] CatchClause
- [ ] FinallySection // TODO implement convertFinally in Converter to support FINALLY in AST
- [ ] TypeAlias // TODO Ast to Expr Conversion needed
- [x] ImportDirective
- [x] ValueArgument
- [x] WhenEntry
- [x] ImportDirective

**Expression**
- [x] ReturnExpression
- [x] ForExpression
- [x] WhileExpression
- [x] BinaryExpression
- [x] BlockExpression
- [x] LambdaExpression
- [x] ThrowExpression
- [x] TryExpression
- [x] IfExpression
- [x] IsExpression
- [ ] WhenExpression // Expecting a when-condition, expecting an expression, is-condition or in-condition
- [x] WhenEntry
- [x] WhenCondition 
- [x] TypeReference


docs/quotes/README.md
```diff
@@ -0,0 +1,78 @@
...
+       )
+}
+```
+
+Below you can find a comprehensive map between Kotlin's PSI element constructs and various quotes in arrow.meta.quotes. If something's missing here, please let us know [here in our issue tracker](https://github.com/arrow-kt/arrow-meta/issues).
+
+## ClassOrObject (quotes.classorobject)
+|  Quote                                                                                                                        |  Declaration 
+| ------------------------------------------------------------------------------------------------------------------------------| ----------------------------------------------------------------------------------------------------------- |
+| [Class](/docs/apidocs/compiler-plugin/arrow.meta.quotes/classorobject/class.html)                                             | ```"""$`@annotations` $kind $name $`(typeParameters)` $`(params)` : $supertypes"} { $body""".`class` ```    |
```
raulraja: The table looks great! 
We should be using the dokka docs or this is gonna be hard to manually maintain. Have you tried adding this to the kdoc and trying to use the auto-generated table? 

Maintaining this table manually is going to become tedious as the lang changes and some of these which are not type-checked by ank become outdated with those changes. There are many upcoming lang features that may change the scopes and therefore render this table outdated when that happens.

ahinchman1: @raulraja here's my attempt with autogeneration (both options are still available) let me know what you think:

Screen Shot 2019-11-22 at 5 01 34 PM

Not sure if there's a way to prevent signatures in autogeneration but I can look into it

raulraja: Signature are fine IMO. I discussed with @franciscodr today we should just change those comments below the signature to be the type checked ank code snippets. Are those code snippets with ank or just strings?

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
```diff
@@ -0,0 +1,271 @@
...
+import org.jetbrains.kotlin.psi.KtReturnExpression
+import org.jetbrains.kotlin.psi.KtThrowExpression
+import org.jetbrains.kotlin.psi.KtTryExpression
+import org.jetbrains.kotlin.psi.KtTypeAlias
+import org.jetbrains.kotlin.psi.KtTypeReference
+import org.jetbrains.kotlin.psi.KtValueArgument
+import org.jetbrains.kotlin.psi.KtWhenCondition
+
+/**
+ * ```""" $`@annotations` $expression """.annotatedExpression```
```
raulraja: These should be all ank typechecked snippets or they can fall behind the API as it evolves

docs/quotes/README.md
ahinchman1: ah - these have the markdown for code snippets but without ank - let me see if I'm able to use ank with the correct imports.  It is the same case for the README.md for quotes, but they were not formatted in this case.

ahinchman1: @raulraja Is it possible to run typechecking on all this without having to wrap it in the Transform and meta functions?

ahinchman1: For example, without the wrappings, there's no way for the code to know what I'm talking about I'm sure:
Screen Shot 2019-11-25 at 10 48 55 AM

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
ahinchman1: I'm not sure how this can be run with typechecking - without being wrapped as a declaration within the Transform object, I'm afraid the code doesn't know what this is by itself.

I've marked this "blocked" for now as I'm inexperienced with ank, but I will continue to work on adding more examples in the meantime

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
```diff
@@ -199,7 +173,9 @@ fun Meta.importDirective(
   quote(match, map) { ImportDirective(it) }
 
 /**
- * TODO
+ * ```
+ * """ { $`(params)` -> $bodyExpression } """.lambdaExpression
```
ahinchman1: it's possible for LambdaExpressions to also contain function literals. A weird one I think we would benefit from explaining in examples as mentioned in #130

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
```diff
@@ -190,7 +162,9 @@ fun Meta.ifExpression(
   quote(match, map) { IfExpression(it) }
 
 /**
- * TODO
+ * ```
+ * """ import $importPath.$importedName """.importDirective
```
ahinchman1: Testing would benefit the verification of the correctness of this statement.

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
```diff
@@ -279,7 +261,7 @@ fun Meta.throwExpression(
   quote(match, map) { ThrowExpression(it) }
 
 /**
- * TODO
+ * """someFunction(someParam: ${typeElement}.typeReference)"""
```
ahinchman1: Not sure if it makes sense to not include context for examples like this and `valueArgument`

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
```diff
@@ -248,7 +228,9 @@ fun Meta.parameterList(
   quote(match, map) { ParameterList(it) }
 
 /**
- * TODO
+ * ```
+ * """ $modifier $visibility val $name: $typeReference = $getter by $delegate""".property
+ * ```
```
ahinchman1: I'm starting to wonder if it makes sense to exclude all PSIElements from our scopes i.e. we can only get val/var from a property with this method in `KtProperty.java`:
```    @Override
    @NotNull
    public PsiElement getValOrVarKeyword() {
        PsiElement element = findChildByType(VAL_VAR_TOKEN_SET);
        assert element != null : "Val or var should always exist for property" + this.getText();
        return element;
    }
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
ahinchman1: Update: unblocked by @franciscodr - we agreed to not worry about ank for this particular PR, but to merge in all the examples and the file structure as he looks more into formatting/ank stuff perhaps in another PR

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
```diff
@@ -0,0 +1,315 @@
...
+import org.jetbrains.kotlin.psi.KtReturnExpression
+import org.jetbrains.kotlin.psi.KtThrowExpression
+import org.jetbrains.kotlin.psi.KtTryExpression
+import org.jetbrains.kotlin.psi.KtTypeAlias
+import org.jetbrains.kotlin.psi.KtTypeReference
+import org.jetbrains.kotlin.psi.KtValueArgument
+import org.jetbrains.kotlin.psi.KtWhenCondition
+
+/**
+ * ```
```
raulraja: We may be able to do something like:
```
` ` `kotlin:ank:playground
import ..ElementScope
import ..ClassScope
fun ElementScope.wontRunButWillTypeCheckTheTemplate(): ClassScope.() -> Any? = {
//sampleStart
""" $`@annotations` $expression """.annotatedExpression
//sampleEnd
}
` ` `
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
```diff
@@ -0,0 +1,315 @@
...
+ */
+fun Meta.catchClause(
+  match: KtCatchClause.() -> Boolean,
+  map: CatchClause.(KtCatchClause) -> Transform<KtCatchClause>
+): ExtensionPhase =
+  quote(match, map) { CatchClause(it) }
+
+/**
+ * ```
+ *  """|$`@annotations` $kind $name $`(typeParameters)` $`(params)` : $supertypes"} { $body } """.`class`
```
raulraja:
```suggestion
 *  """|$`@annotations` $kind $name $`(typeParameters)` $`(params)` : $supertypes"} { 
       | $body 
       |}""".`class`
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
```diff
@@ -0,0 +1,315 @@
...
+ */
+fun Meta.destructuringDeclaration(
+  match: KtDestructuringDeclaration.() -> Boolean,
+  map: DestructuringDeclaration.(KtDestructuringDeclaration) -> Transform<KtDestructuringDeclaration>
+): ExtensionPhase =
+  quote(match, map) { DestructuringDeclaration(it) }
+
+/**
+ * ```
+ * """ $packageDirective $importList $classes """.file
```
raulraja: what about other declarations that are not classes? Are there scope properties for properties, functions and type aliases?

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
```diff
@@ -0,0 +1,315 @@
...
+ */
+fun Meta.ifExpression(
+  match: KtIfExpression.() -> Boolean,
+  map: IfExpression.(KtIfExpression) -> Transform<KtIfExpression>
+): ExtensionPhase =
+  quote(match, map) { IfExpression(it) }
+
+/**
+ * ```
+ * """ import $importPath.$importedName """.importDirective
```
raulraja: Are there properties for qualified `as` imports in the Scope of `ImportDirective`?

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
```diff
@@ -0,0 +1,315 @@
...
+ */
+fun Meta.lambdaExpression(
+  match: KtLambdaExpression.() -> Boolean,
+  map: LambdaExpression.(KtLambdaExpression) -> Transform<KtLambdaExpression>
+) : ExtensionPhase =
+  quote(match, map) { LambdaExpression(it) }
+
+/**
+ * ```
+ * """ $`@annotations` $modifier value """.`modifierList`
```
raulraja: Is $modifier a property of the scope or can any soft keyword be a modifier?
If this points to a ScopedList then it probably should be renamed to `modifiers`

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
```diff
@@ -0,0 +1,315 @@
...
+ */
+fun Meta.modifierList(
+  match: KtModifierList.() -> Boolean,
+  map: ModifierList.(KtModifierList) -> Transform<KtModifierList>
+): ExtensionPhase =
+  quote(match, map) { ModifierList(it) }
+
+/**
+ * ```
+ * """ $modality $visibility $receiver.$name($`(typeParameters)`): $returnType = { $body } """.namedFunction
```
raulraja:
```suggestion
 * """ $modality $visibility fun $`(typeParameters)` $receiver.$name $`(args)` : $returnType = { $body } """.namedFunction
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
raulraja: We can wrap that method so it's safe or write a safe version of it directly over `KtProperty`

compiler-plugin/src/main/kotlin/arrow/meta/quotes/element/ValueArgument.kt
```diff
@@ -0,0 +1,15 @@
+package arrow.meta.quotes.element
+
+import arrow.meta.quotes.Scope
+import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtValueArgument
+import org.jetbrains.kotlin.psi.KtValueArgumentName
+
+/**
+ * A template destructuring [Scope] for a [ValueArgument]
```
raulraja: All these comments should be replaced with the same examples we are using in the functions

docs/_data/sidebar.yml
```diff
@@ -20,20 +20,23 @@ options:
...
         url: /docs/apidocs/compiler-plugin/arrow.meta.quotes/index.html
 
-      - title: classOrObject
-        url: /docs/apidocs/compiler-plugin/arrow.meta.quotes/class-or-object.html
+      - title: Quotes
+        url: /quotes
 
-      - title: func
-        url: /docs/apidocs/compiler-plugin/arrow.meta.quotes/func.html
+      - title: class
```
raulraja: This menu will be autogenerated based on what @franciscodr and I discussed on Thursday so you don't have to worry about adding each one of the items.

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
raulraja: I think `Parameter` or `ValueArgument` is simpler `param: $typeReference` and then when we show `TypeElement` we need to show how to destructure it's type args and other members. `$name`$(typeArgs)`

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
ahinchman1: Yes, there are - would you want multiple kinds of examples then?

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
ahinchman1: For whatever reason this is a singular KtModifer - perhaps it's for the whole list

compiler-plugin/src/main/kotlin/arrow/meta/phases/analysis/DefaultElementScope.kt
```diff
@@ -334,5 +412,14 @@ class DefaultElementScope(project: Project) : ElementScope {
 
   override val String.annotatedExpression: AnnotatedExpression
     get() = AnnotatedExpression(expression.value as KtAnnotatedExpression)
```
ahinchman1: @bloderxd @raulraja I've only noticed this is how we're doing a number of string -> psi elements for the ones that are not automatically provided by the `PsiFactory` - why not just do it this way for the remainder of items?

docs/quotes/README.md
```diff
@@ -0,0 +1,79 @@
+---
+layout: docs
+title: Quotes
+permalink: /quotes
+---
```
ahinchman1: @franciscodr is it okay to leave this file in here even if I take it out of the sidebar?

docs/_data/sidebar.yml
ahinchman1: @raulraja does it make sense to remove all these items from the side bar then, or to add them? I will adjust accordingly!

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
ahinchman1: This will have to be another element for future work added to it, I think

compiler-plugin/src/main/kotlin/arrow/meta/phases/analysis/DefaultElementScope.kt
raulraja: yes, we need a matching function in the ElementScope for every Scope we have that turns a template into a scope.

compiler-plugin/src/main/kotlin/arrow/meta/quotes/classorobject/ClassDeclaration.kt
```diff
@@ -13,10 +11,16 @@ import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
...
  *
- * An extension function of [Meta] and inheriting from [ExtensionPhase], [classOrObject] was designed to feed in
+ * The [ClassDeclaration] is projected over the template to allow destructuring of the different parts of the
+ * [KtClass]. The scope enables template syntax where the user may add new members or modify the class structure
+ * before it's compiled.
+ *
+ *  * [`class`] is a function that intercepts all [KtClass] elements that [match]
+ * then uses a [Transform] to change the intercepted AST tree before compilation.
+ *
+ * An extension function of [Meta] and inheriting from [ExtensionPhase], [class] was designed to feed in
```
raulraja:
```suggestion
 * An extension function of [Meta] and inheriting from [ExtensionPhase], [classDeclaration] was designed to feed in
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/classorobject/ObjectDeclaration.kt
```diff
@@ -0,0 +1,38 @@
+package arrow.meta.quotes.classorobject
+
+import arrow.meta.quotes.Scope
+import org.jetbrains.kotlin.psi.KtObjectDeclaration
+
+/**
+ * <code>""" object $name { $body }""".objectDeclaration</code>
+ *
+ * TODO add wrappers for certain boolean logic so that there are more options to create KtObjectDeclaration i.e. companion object
```
raulraja:
```suggestion
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/classorobject/ObjectDeclaration.kt
```diff
@@ -0,0 +1,38 @@
...
+ * @see [DefaultElementScope::companionObject]
```
raulraja:
```suggestion
 * @see [ElementScope::companionObject]
```
We should not link to the implementations but to the public APIs instead. The user is never exposed to `DefaultElementScope`.

compiler-plugin/src/main/kotlin/arrow/meta/quotes/classorobject/ObjectDeclaration.kt
```diff
@@ -0,0 +1,38 @@
...
+ * import arrow.meta.quotes.objectDeclaration
+ *
+ * val Meta.reformatObjectDeclaration: Plugin
+ *   get() =
+ *     "ReformatObjectDeclaration" {
+ *       meta(
+ *        objectDeclaration({ true }) { c ->
+ *          Transform.replace(
+ *            replacing = c,
+ *            newDeclaration = """ object $name { $body }""".`object`
```
raulraja: All templates in all examples should expose all available properties of the scope so the users can understand what a full declaration looks like and which parts they can use to compose the structure again. In this case, the examples are removing from the tree annotations, modifiers and supertypes. This may be true for other examples. All quote examples should include all members of the scope or the user may lose information as they copy and paste it from the docs into their projects.

compiler-plugin/src/main/kotlin/arrow/meta/quotes/expression/ThrowExpression.kt
```diff
@@ -0,0 +1,36 @@
+package arrow.meta.quotes.expression
+
+import arrow.meta.quotes.Scope
+import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtThrowExpression
+
+/**
+ * <code>""" $`throw` """.`throw`</code>
+ *
+ * A template destructuring [Scope] for a [KtThrowExpression]. See below:
```
raulraja:
```suggestion
 * A template destructuring [Scope] for a [KtThrowExpression]
```

raulraja: Same for others, IMO it does not need an introduction

docs/quotes/README.md
```diff
@@ -0,0 +1,79 @@
...
+     Transform.replace(
+      replacing = e,
+        newDeclaration = """ if $`(condition)` $then $`else` """.`if`
+       )
+}
+```
+
+Below you can find a comprehensive map between Kotlin's PSI element constructs and various quotes in arrow.meta.quotes. If something's missing here, please let us know [here in our issue tracker](https://github.com/arrow-kt/arrow-meta/issues).
+
+## ClassOrObject (quotes.classorobject)
```
raulraja: This table has to go as it will be automatically generated

docs/quotes/README.md
raulraja: This file is autogenerated from the docs and we should remove it. Any useful here should appear in the package info

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
ahinchman1: Yes, I think in another PR, I want to go back to add more examples and do you suggest to wrap methods into safer versions.

ahinchman1:
Sounds good! Sorry, which files are you referencing? I removed the files in the nested options if that's what you mean -

Screen Shot 2019-11-26 at 1 34 25 PM

compiler-plugin/src/main/kotlin/arrow/meta/phases/analysis/DefaultElementScope.kt
ahinchman1: Right - but some of the PsiFactory does not have methods available for certain PsiElements that can be used with the `delegate` -  I think what @bloderxd did is an interesting idea, and I wondered if this is a pattern to continue for all elements not easily obtainable from `PsiFactory.kt` or if we should think about changing them all to be similar?

compiler-plugin/src/main/kotlin/arrow/meta/quotes/classorobject/ObjectDeclaration.kt
ahinchman1: Yes, it is definitely true for other examples as well. I'm happy to add on to all that, but in order to create more examples, we need more exposed from PsiElements.

For example, we need additional information that we currently don't have in order to create a `companion object`.  This is what I meant by the Property example where we cannot choose to generate a val/var

ahinchman1: I think for the sake of this PR, it is okay to merge in basic examples and to continue working over all these elements by exposing what is necessary in order to create better examples.

docs/quotes/README.md
ahinchman1: No problem, I can remove this file

compiler-plugin/src/main/kotlin/arrow/meta/quotes/classorobject/ObjectDeclaration.kt
ahinchman1: I'm going to push up an example of an element that has different kinds of examples for Parameter shortly so we can prototype future work (done this week but in different PRs) off that!

ahinchman1: Prototype can be observed in Parameter:

Screen Shot 2019-11-26 at 3 50 42 PM

compiler-plugin/src/main/kotlin/arrow/meta/quotes/element/ParameterList.kt
```diff
@@ -0,0 +1,38 @@
...
+ *      replacing = l,
+ *      newDeclaration = """ $`(params)` """.parameterList
+ *     )
+ *    }
+ *   )
+ *  }
+ *```
+ */
+class ParameterList(
+  override val value: KtParameterList?,
```
ahinchman1: Not sure what to do about this one

compiler-plugin/src/main/kotlin/arrow/meta/quotes/element/WhenEntry.kt
```diff
@@ -0,0 +1,44 @@
...
+ *   }
+ * ```
+ */
+
+class WhenEntry(
+  override val value: KtWhenEntry?,
+  val conditions: ScopedList<KtWhenCondition> = ScopedList(
+    separator = " && ",
+    value = value?.conditions?.toList() ?: listOf()
+  ),
```
ahinchman1: For things like this and also the catch clauses list, I'm wondering if we should be putting the logic in the string transformation since it can't exactly be done via scoped list

compiler-plugin/src/main/kotlin/arrow/meta/quotes/expression/TryExpression.kt
```diff
@@ -0,0 +1,44 @@
...
+ *     )
+ *    }
+ * ```
+ */
+class TryExpression(
+  override val value: KtTryExpression?,
+  val tryBlock: Scope<KtBlockExpression> = Scope(value?.tryBlock),
+  val catchClauses: ScopedList<KtCatchClause> = ScopedList(
+    prefix = "catch ",
+    value = value?.catchClauses ?: listOf()
```
ahinchman1: This was the other side of when formatting lists doesn't go quite right

compiler-plugin/src/main/kotlin/arrow/meta/quotes/expression/loopexpression/ForLoopExpression.kt
```diff
@@ -0,0 +1,40 @@
...
+ *        }
+ *       )
+ *     }
+ * ```
+ */
+class ForLoopExpression(
+  override val value: KtForExpression,
+  val `(param)`: Parameter = Parameter(value.loopParameter),
+  val loopRange: Scope<KtExpression> = Scope(value.loopRange), // TODO KtExpression scope
+  val destructuringDeclaration: DestructuringDeclaration = DestructuringDeclaration(value.destructuringDeclaration)
```
ahinchman1: It's possible to also have a for loop that is destructuring - I suppose we could add another example for that:

```
for ((a, b) in collection) { ... }
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/filebase/File.kt
```diff
@@ -44,7 +61,7 @@ class File(
   val virtualFilePath: String = value.virtualFilePath,
   val danglingAnnotations: ScopedList<KtAnnotationEntry> = ScopedList(value = value.danglingAnnotations, postfix = ", "),
   val fileType: FileType = value.fileType,
-  val declaractions: ScopedList<KtDeclaration> = ScopedList(value = value.declarations, postfix = ", "),
+  val declarations: ScopedList<KtDeclaration> = ScopedList(value = value.declarations, postfix = ", "),
   val stub: KotlinFileStub? = value.stub,
```
ahinchman1: This one also requires more work, it's definitely one of the more versatile elements that requires more examples

compiler-plugin/src/main/kotlin/arrow/meta/quotes/modifierlist/ModifierList.kt
```diff
@@ -30,21 +32,10 @@ import org.jetbrains.kotlin.psi.psiUtil.visibilityModifier
...
-): ExtensionPhase =
-  quote(match, map) { ModifierList(it) }
-
-/**
- * A template destructuring [Scope] for a [KtModifierList]
  */
 class ModifierList(
   override val value: KtModifierList?,
-  val `@annotations`: ScopedList<KtAnnotationEntry> = ScopedList(value?.annotationEntries ?: listOf()),
+  val `@annotations`: ScopedList<KtAnnotationEntry> = ScopedList(value?.annotationEntries
```
ahinchman1: Left this one alone

compiler-plugin/src/main/kotlin/arrow/meta/quotes/nameddeclaration/stub/typeparameterlistowner/Property.kt
```diff
@@ -0,0 +1,154 @@
...
+ *      }
+ *     )
+ *    }
+ * ```
+ */
+class Property(
+  override val value: KtProperty?,
+  val modality: Name? = value?.modalityModifierType()?.value?.let(Name::identifier),
+  val visibility: Name? = value?.visibilityModifierType()?.value?.let(Name::identifier),
+  val `(typeParameters)`: ScopedList<KtTypeParameter> = ScopedList(prefix = "<", value = value?.typeParameters
```
ahinchman1: The only one left off is `returnType` and `(typeParameters)` here - I'll be looking for examples!

compiler-plugin/src/main/kotlin/arrow/meta/quotes/element/WhenEntry.kt
raulraja: yes, it can be done like that but the separator can be any other binary expression that should come from the model somwhere

compiler-plugin/src/main/kotlin/arrow/meta/quotes/expression/TryExpression.kt
raulraja: You can also use `orEmpty()` in place of `?: listOf()`

compiler-plugin/src/main/kotlin/arrow/meta/phases/analysis/DefaultElementScope.kt
raulraja: Let's think about it and discuss it after KC if you are ok with it.

raulraja: Which method does it not have? Some methods like `.expression`as in the example above account for a big number of element types which result after parsing the structure and are all subtypes of KtExpression.

ahinchman1: Works for me! Oh, that would take care of many I'm sure. 

I'll have to check for the rest of them, but one example is for `KtCatchClause`

compiler-plugin/src/main/kotlin/arrow/meta/quotes/classorobject/ObjectDeclaration.kt
```diff
@@ -0,0 +1,42 @@
...
+ *             newDeclaration = """
+ *               |$textOffset
+ *               |$`@annotations`
+ *               |$modality object $name $primaryConstructorParameterList$primaryConstructor: $superTypeListEntries {
+ *               |    $anonymousInitializers
+ *               |    $secondaryConstructor
+ *               |    $declarations
+ *               |}""".trimMargin().`object`
+ *             )
+ *           }
```
ahinchman1: @raulraja the only real difference that seemed worth picking out was the `offsetText`, but that was a very early part of the code and I don't know if it serves much purpose anymore.  

I also assume that the items for primary/secondary constructors will likely come back empty for object declarations but included them in there anyway (unless you want them removed?)

Cleaned up as much as I could that could be grabbed with methods that were able to be drilled down further (i.e. `$declarations` in lieu of `$classBody`)

compiler-plugin/src/main/kotlin/arrow/meta/quotes/element/CatchClause.kt
```diff
@@ -0,0 +1,38 @@
...
+ *      )
+ *     }
+ *    )
+ *   }
+ *```
+ */
+class CatchClause(
+  override val value: KtCatchClause?,
+  val `(parameter)`: Parameter = Parameter(value?.catchParameter),
+  val `{ body }`: Scope<KtExpression> = Scope(value?.catchBody)
```
ahinchman1: @bloderxd dumb question but does this actually format in the strings like the above?

compiler-plugin/src/main/kotlin/arrow/meta/quotes/element/ImportDirective.kt
```diff
@@ -0,0 +1,40 @@
...
+ *      }
+ *     )
+ *    }
+ * ```
+ */
+class ImportDirective(
+  override val value: KtImportDirective,
+  val importedReference: Scope<KtExpression> = Scope(value.importedReference), // TODO KtExpression scope and quote template, appears to be for diagnostics
+  val importContent: KtImportInfo.ImportContent? = value.importContent,
+  val importPath: ImportPath? = value.importPath
```
ahinchman1: I couldn't tell the difference between `ImportContent` and `ImportReference` but I think I read somewhere in the code that `ImportReference` could possibly be used in diagnostics

compiler-plugin/src/main/kotlin/arrow/meta/quotes/expression/TryExpression.kt
ahinchman1: Thanks! I'll try that!

compiler-plugin/src/main/kotlin/arrow/meta/quotes/MetaExtensions.kt
ahinchman1: to amend - perhaps that what importContents is for? I still haven't quite figured that one out

compiler-plugin/src/main/kotlin/arrow/meta/quotes/classorobject/ObjectDeclaration.kt
ahinchman1: removed offsetText.

compiler-plugin/src/main/kotlin/arrow/meta/quotes/element/CatchClause.kt
bloderxd: Yeah, all kind of kt elements has its text property and that's what scope does in its `toString` function: invoke that property, I'm just wondering if we should keep calling it as `{ body }`, or change it to `body`

raulraja:
🎉

PR #214, opened 2019-11-22
Dialogs, PopUps and Notifications

i-walker:
Exampletory PR for UI wrappers, related to https://github.com/arrow-kt/arrow-meta/issues/131.
-  [X] *Dialog*
Screenshot 2019-11-22 at 16 49 00
I will fix the issue by adding new internal FileTemplates in another PR. For now, this creates a `Kotlin File` and is extendable to all existing `FileTemplates`.
- [x] *PopUp*
- [x] *Notification*


raulraja:
This will come handy to initialize patterns like ADTs etc in file templates when you create a new model, algebra, interpreters etc as we can include stubs and instructions to get the user going. I'm thinking in `New ADT` style feature which gives you the entire structure and you just type names and tab over arguments. Nothing for now just some thoughts for the future of this api. Kudos

i-walker:
@raulraja this is ready for review. I will add Docs to Notifications once I know more about them.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/IdeSyntax.kt
```diff
@@ -17,8 +17,12 @@ import arrow.meta.ide.dsl.editor.structureView.StructureViewSyntax
...
 import arrow.meta.ide.dsl.extensions.ExtensionProviderSyntax
+import arrow.meta.ide.dsl.ui.dialogs.DialogSyntax
+import arrow.meta.ide.dsl.ui.notification.NotificationSyntax
+import arrow.meta.ide.dsl.ui.popups.PopupSyntax
 
 interface IdeSyntax : IntentionSyntax, IconProviderSyntax,
   SyntaxHighlighterSyntax, InspectionSyntax, AnActionSyntax, ColorSettingsSyntax, HintingSyntax,
   LanguageSyntax, LineMarkerSyntax, LiveTemplateSyntax, NavigationSyntax, SearchSyntax, StructureViewSyntax,
-  UsageSyntax, ExtensionProviderSyntax, DocumentationSyntax, KotlinIndicesSyntax
+  UsageSyntax, ExtensionProviderSyntax, DocumentationSyntax, KotlinIndicesSyntax, DialogSyntax, PopupSyntax,
```
ahinchman1: wowwww! Very nice @i-walker

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/ui/dialogs/DialogSyntax.kt
```diff
@@ -0,0 +1,106 @@
...
+   *     addFileAction("ExampleAction", "New File", "Creates a new File",
+   *       buildDialog = { project, directory ->
+   *         setTitle("CostumeTitle")
+   *         addKind("File", KotlinFileType.INSTANCE.icon, "Kotlin File")
+   *       }
+   *      )
+   *     )
+   *   }
+   * ```
+   * @param actionId has to be unique
```
ahinchman1: I haven't been very good about this part and plan on coming back later on in the quote-and-template system, but should we only describe tricky params or all of them eventually?

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/ui/dialogs/DialogSyntax.kt
```diff
@@ -0,0 +1,106 @@
...
+          if (super.isAvailable(ctx))
+            LangDataKeys.IDE_VIEW.getData(ctx)?.let { ideView: IdeView ->
+              PlatformDataKeys.PROJECT.getData(ctx)?.let { project ->
+                ideView.directories.any {
+                  ProjectRootManager.getInstance(project).fileIndex.isInSourceContent(it.virtualFile)
+                }
+              }
+            }
+          else
+            false
```
ahinchman1: nit: if we have each case on the next line for both `if` and `else` perhaps we can go ahead and wrap them in brackets. It's more common in Kotlin to have the following conventions:

```
if (something) {
   something_happens
} else {
    something_else_happens
}
```

or 

```
if (something) something_happens else something_else_happens

File: idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/ui/dialogs/DialogSyntax.kt [hunk 4]
```diff
@@ -0,0 +1,106 @@
...
+        } ?: false
+
+      override fun postProcess(createdElement: PsiFile?, templateName: String?, customProperties: MutableMap<String, String>?): Unit =
+        createdElement?.let { el -> templateName?.let { name -> customProperties?.let { prop -> postProcess(el, name, prop) } } }
+          ?: Unit
+
+      override fun createFileFromTemplate(name: String?, template: FileTemplate?, dir: PsiDirectory?): PsiFile? =
+        name?.let { n -> template?.let { t -> dir?.let { d -> createFileFromTemplate(n, t, d) } } }
```
ahinchman1: This is interesting, nice hat trick

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/ui/dialogs/DialogSyntax.kt
```diff
@@ -0,0 +1,106 @@
...
+                }
+              }
+            }
+          else
+            false
+        } ?: false
+
+      override fun postProcess(createdElement: PsiFile?, templateName: String?, customProperties: MutableMap<String, String>?): Unit =
+        createdElement?.let { el -> templateName?.let { name -> customProperties?.let { prop -> postProcess(el, name, prop) } } }
+          ?: Unit
```
ahinchman1: Is it necessary to return a `Unit`? I know we like to keep it pure and avoid Unit if possible but I was wondering about the motivation to do so over something this like this since we're just looking to make a call here:

```
override fun postProcess(createdElement: PsiFile?, templateName: String?, customProperties: MutableMap<String, String>?) =
         createdElement?.let { el -> 
                templateName?.let { name -> 
                      customProperties?.let { prop -> 
                            postProcess(el, name, prop) 
                      } 
               } 
         }
```

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/ui/popups/PopupSyntax.kt
```diff
@@ -0,0 +1,150 @@
...
+   *                  )
+   *                )
+   *              ).show(this)
+   *            }
+   *          }
+   *        )
+   *      )
+   *    )
+   *  }
+   * ```
```
ahinchman1: wow dang I can't believe how easy it is to write these! I love that it's the same user experience across the board. Great work Imran!

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/ui/popups/PopupSyntax.kt
```diff
@@ -0,0 +1,150 @@
...
+      popUps.run {
+        createListPopup(object : BaseListPopupStep<A>(title, map { it.element }, sameIcon) {
+          override fun isSelectable(value: A): Boolean =
+            this@run.firstOrNull { it == value }?.isSelectable ?: false
+
+          override fun hasSubstep(selectedValue: A): Boolean =
+            this@run.firstOrNull { it == selectedValue }?.hasSubStep ?: false
+
+          override fun onChosen(selectedValue: A, finalChoice: Boolean): PopupStep<*> =
+            this@run.firstOrNull { it == selectedValue }?.let { a -> a.onChosen(selectedValue, finalChoice) }
```
ahinchman1: Is `a` meant to also be `a: A`?  perhaps this can be made a little clearer, like `selectedValueType`

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/ui/dialogs/DialogSyntax.kt
i-walker: It depends on which of them are important to know. For example with `actionId`. the ide crashes if there are more than 2 Actions with the same Id. Those `actionIds` are used by the`AnActionManager`, which orchestrates all existing Actions and their lifecycle.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/ui/dialogs/DialogSyntax.kt
i-walker: Your function has a different return Type than the one this features needs.

Issue #219, opened 2019-11-23
Transform replace tests


bloderxd:
This issue proposes the creation of tests to Transform replace.

PR #220, opened 2019-11-23
Transform replace tests

bloderxd:
This PR provides tests to Transform replace. This is related to https://github.com/arrow-kt/arrow-meta/issues/219


rachelcarmena:
Absolutely, thanks @bloderxd !!! :clap: I'm going to move them to `compiler-plugin` in order to keep `testing-plugin` just with simple examples about how to write tests. Thanks!!!

Issue #177, opened 2019-11-18
Support additional ops for Quote transformations.


raulraja:
Currently, quote transformations are limited to `Transform.replace` which replaces a returned node by a list of declarations. There is an additional set of operations that we would like transformation to be used on.

We would like to support further transformations:

| Transform | Description                                                                    |
|-----------|--------------------------------------------------------------------------------|
| replace   | Replaces selected node by list of new declarations                             |
| remove    | Removes one or n nodes                                                         |
| relocate  | Places this node elsewhere in the tree                                         |
| newSources   | wraps other transforms and outputs their result in other KtFiles             |
| diagnose  | Returns a list of `Diagnostic` that the compiler should present to the user    |
| many      | wraps a list of transforms and when interpreted executes then in order         |
| +         | many(this, other)                                                              |
| empty     | noop                                                                           |
| debug     | wraps a transform and enables tracing thru its a progression over the internals |

We should do one PR per transform and ensure these transformations are tested with the `quoteOutputMatches`combinators in the testing library. The testing library would need to receive a few additional algebras to verify produced sources in the `newSources` transform

PR #235, opened 2019-11-25
Transform many

bloderxd:
This PR proposes the implementation of a Transform that combine other transformations:

```kotlin
private val Meta.transformManySimpleCase: Plugin
  get() = "Transform Many" {
    meta(
      `class`({ name == "ManySimpleCase" }) { c ->
        changeClassVisibility("ManySimpleCase", c, this) + removeFooPrint(c, this)
      }
    )
  }

private fun CompilerContext.changeClassVisibility(className: String, context: KtClass, declaration: ClassDeclaration): Transform<KtClass> = declaration.run { Transform.replace(
  replacing = context,
  newDeclaration = """
    | private class $className {
    |   $body
    | } """.`class`.synthetic
)}

private fun CompilerContext.removeFooPrint(context: KtClass, declaration: ClassDeclaration): Transform<KtClass> = declaration.run { Transform.remove(
  removeIn = context,
  declaration = """ fun printFirst() = println("Foo") """.expressionIn(context)
)}
```
As showed in this example we can use the operator `+` to combine how many transforms we want. This is related to #177


compiler-plugin/src/main/kotlin/arrow/meta/quotes/Transform.kt
```diff
@@ -106,5 +149,10 @@ sealed class Transform<out K : KtElement> {
...
     val empty: Transform<Nothing> = Empty
   }
-
 }
+
+operator fun <K : KtElement> Transform<K>.plus(transform: Transform<K>): Transform.Many<K> =
+  Transform.Many(arrayListOf(this, transform))
+
+operator fun <K : KtElement> Transform.Many<K>.plus(transform: Transform<K>): Transform.Many<K> =
+  Transform.Many(this.transforms.also { it.add(transform) })
```
i-walker:
```suggestion
  Transform.Many(this.transforms.add(transform))
```
can we reduce it to this?

bloderxd: ArrayList `add` function actually returns a boolean with the result of the addition, that's why I put it on an apply

Issue #96, opened 2019-11-08
Improve IDE (as you type) synth resolution running the Quote system


raulraja:
Synthetic resolution based on computeCache only happens when the files are compiled and in .class form in the local module output.

https://github.com/arrow-kt/arrow-meta/blob/cee5d0eebe12f89fdc414eb7490150f1135df5c2/idea-plugin/src/main/kotlin/arrow/meta/ide/phases/resolve/MetaSyntheticPackageFragmentProvider.kt#L182

This is inconvenient because users get redlines and no auto-completion help until they compile.

We would like to be able to run Meta there automatically as the user is typing on the editor buffer to offer the same SynthDescriptors we are offering now but before it compiles the full project.

Additionally, @jereksel  pointed out an extension in the IDE we are not using and apparently apparently may help us solve some of these  problems. We need to investigate it and see how we can use it. `org.jetbrains.kotlin.idea.core.extension.KotlinIndicesHelperExtension`

I have the feeling that this extension may help us inject automatically functions and declarations which are callable descriptors to contribute to their scope as the user types. Running the quote system on each keystroke may not be expensive if we do something like:

1. Grab the KtFile we are on and feed to a quote
2. The quote then already traverses the PSI tree and matches it
3. The quote internally already generates a new KtFile
4. We use the resolve API to partially type check the new KtFile as it is done now in the `computeCache` function linked above and add it to the computed cache before compiling. This generates the synth descriptors which are fed to the PackageFragmentProvider and SynthResolverExtension automatically since these are invoked as the user types and the user should see the new synth descriptors visible in the editor.

jansorg:
@raulraja, I wasn't sure who's best to answer this.

Here are questions for a few problems I encountered:

1. Starting the quote system on the current file only is not enough. The current buffer may reference Meta-data which is contained in another file or even in another module of the same project. The complete solution is probably to run all `*.kt` files of all modules where Meta is active through the quote system. This is what we had when we used the .class files generated by the compiler.
**Does this sound reasonable?** 
1. We have to be extra careful here to
    - avoid memory leaks. This could happen easily because we're caching PSI elements.
    - be fast. My thinking is that the current buffer should be run through the quote system whenever it changes. All the other `*.kt` files should be only run through the quote system at startup and whenever they change. We could also cache this on disk, if it turns out to be slow.
[...]

PR #252, opened 2019-11-26
synthetic resolution using quote system

jansorg:
Partial implementation of https://github.com/arrow-kt/arrow-meta/issues/96

This PR should be good for the first batch of changes. It's not yet ready for production use, though.
At least this should help to discuss the approach and possible changes to the changes and the integration of the quote system.

Known issues:
- The initial highlighting after startup isn't reliably updated. This seems to be timing issue and the enforced annotator refresh doesn't seem to be applied in this case.

## Changes
- [x] the new tests are passing and the resolving should be okay for smaller projects. I've tested this with the arrow-meta project (140 .kt files) and the editor experience is good (no meta transformations applied afaik).
- [x] Implement incremental updates to the cache transformations, if necessary

## Changes for a follow-up prs:
- [ ] Fix the initial highlighting (see above)
- [ ] More tests, cover all compiler plugins, if possible
- [ ] Cancel ongoing cache refresh when a new refresh is triggered (or: skip new refresh if there's an active refresh in progress). Refresh requests shouldn't be queued endlessly. This will improve the editor experience when the user is slowly typing, for example.
- [ ] Only apply transformations to a .kt file if it's in a module where the arrow gradle plugin is enabled. Otherwise we might generate unwanted elements.
- [ ] Code completions withs arrow's synthetic elements are not yet implemented. Add test and implement.
- [ ] Implement tests without a flush of the transformation worker pool
- [ ] Update tests to test syntax maintained by i-walker?

## Follow up changes to the Quote system?
I'm unclear if this makes sense or is possible to implement.
- It would be great to provide a string instead of a PsiFile for the source of a transformation. At the moment a PsiFile is required and we have to commit the PsiFile in IntelliJ after every change to the editor. This results in degraded performance.
[code where we commit the PsiFile](https://github.com/arrow-kt/arrow-meta/blob/5a3c0e99c0cec6e6a9ee0de434c8fb52922d1157/idea-plugin/src/main/kotlin/arrow/meta/ide/phases/resolve/QuoteSystemCache.kt#L135)
- afaik the transformation of the Quote system creates a new PsiFile for each single transformation. It would be great to improve this.

## Changes

Implements a fragment provider and synthetic resolve extension with descriptors based on the transformations of the Quote system. The quote system is executed on the files of the current project when the editor changed or .kt files on disk changed.

At startup, the full set of kt files is transformed in the background to get an initial set of full transformations.

This commit re-enables the synthetic resolve extension, which was disabled previously. The synthetic extension was extracted from the `MetaSyntheticFragmentProvider`. Previously, the fragment provider was implementing two extension points and therefore would have been instantiated twice. Now fragment provider and synthetic resolve extension are distinct implementations. They both access the cache maintained by `QuoteSystemCache`.

This also adds tests to make sure that resolving works for static content and when a user types into the editor. 
[QuoteSystemCacheTest](https://github.com/arrow-kt/arrow-meta/pull/252/files#diff-f06e16587c73fae825c3fc35940deaee), [QuoteTransformationTest](https://github.com/arrow-kt/arrow-meta/pull/252/files#diff-56bc6d1e95145a07d4dc03da8eca8fd7)

## Performance
These are preliminary results. I've tests this with a small Kotlin project (69 .kt files in total).
Currently, computing the list of transformations is fast (~500ms).

Applying the transformations is slow ([this method is called](https://github.com/arrow-kt/arrow-meta/blob/bc16383741beb183c986d568756383544393c7f2/compiler-plugin/src/main/kotlin/arrow/meta/quotes/Quote.kt#L271)). It took ~5s to apply the transformations to the 69 files. imho, this is too slow and an argument to implement incremental updates instead of applying transformations to all files in the project.


compiler-plugin/src/main/kotlin/arrow/meta/quotes/Quote.kt
```diff
@@ -151,6 +152,17 @@ inline fun <reified K : KtElement, S : Scope<K>> Meta.quote(
 ): ExtensionPhase =
   quote(QuoteFactory(transform), match, map)
 
+// fixme: AnalysisDefinition and analysisIdeExtensions are in place until a better integration with the IDE is found
```
raulraja: Since the IDE does not expose analysis as a phase this can potentially be implemented in the future using the KotlinCacheService and intercepting the creation of the ResolutionFacade which is what the IDE type proofs plugin does currently just to log some info in the sv-test-proofs branch

idea-plugin/src/main/kotlin/arrow/meta/ide/phases/resolve/MetaSyntheticPackageFragmentProvider.kt
```diff
@@ -135,340 +62,39 @@ class MetaSyntheticPackageFragmentProvider(val project: Project) :
...
-        descriptorCache[fqName]?.filterIsInstance<ClassifierDescriptor>()?.map { it.name }?.toSet()
+        cache.resolved(fqName)?.filterIsInstance<ClassifierDescriptor>()?.map { it.name }?.toSet()
 
       override fun getContributedClassifier(name: Name, location: LookupLocation): ClassifierDescriptor? =
-        descriptorCache[fqName]?.filterIsInstance<ClassifierDescriptor>()?.firstOrNull { it.name == name }
+        cache.resolved(fqName)?.filterIsInstance<ClassifierDescriptor>()?.firstOrNull { it.name == name }
 
       override fun getContributedDescriptors(kindFilter: DescriptorKindFilter, nameFilter: (Name) -> Boolean): Collection<DeclarationDescriptor> =
-        descriptorCache[fqName]?.filter { it.name == name } ?: emptyList()
+        cache.resolved(fqName)?.filter { it.name == name } ?: emptyList()
```
raulraja: Not that it really matters but if you wanted to get rid of these Elvis operators you could potentially just chain orEmpty() which should be available in all nullable types collections.

idea-plugin/src/main/kotlin/arrow/meta/ide/phases/resolve/QuoteSystemCache.kt
```diff
@@ -0,0 +1,388 @@
...
+        }
+      }
+    }
+  }
+
+  private val metaCache: MetaTransformationCache = DefaultMetaTransformationCache()
+
+  // pool where the quote system transformations are executed.
+  // this is a single thread pool to avoid concurrent updates to the cache.
+  // we keep a pool per project, so that we're able to shut it down when the project is closed
```
raulraja: Nice!

idea-plugin/src/main/kotlin/arrow/meta/ide/phases/resolve/QuoteSystemCache.kt
```diff
@@ -0,0 +1,388 @@
...
+  private val pool: ExecutorService = Executors.newFixedThreadPool(1) { runnable -> Thread(runnable, "arrow worker, ${project.name}") }
```
raulraja: Do submitted running and on hold tasks get automatically canceled if a user action produces invalidation or requires a fresh computation of the cache ?

idea-plugin/src/main/kotlin/arrow/meta/ide/phases/resolve/QuoteSystemCache.kt
```diff
@@ -0,0 +1,388 @@
...
+
+          // the kotlin facade needs files with source and target elements
+          val kotlinCache = KotlinCacheService.getInstance(project)
+          val facade = kotlinCache.getResolutionFacade(updatedFiles + transformed.map { it.second })
+
+          // fixme: remove descriptors which belong to the newly transformed files
+          transformed.forEach { (sourceFile, transformedFile) ->
+            // fixme this triggers a resolve which already queries our synthetic resolve extensions
+            val synthDescriptors = transformedFile.declarations.mapNotNull {
+              val desc = facade.resolveToDescriptor(it, BodyResolveMode.FULL)
```
raulraja: Is full resolution required?

raulraja:
Excellent progress @jansorg! Left some comments but nothing blocking

compiler-plugin/src/main/kotlin/arrow/meta/quotes/Quote.kt
raulraja: Nevermind, just saw you are doing that already

PR #262, opened 2019-11-28
helloWorld ide Plugin 

i-walker:
Aims to solve #210

PR #264, opened 2019-11-28
Core Ide Docs 

i-walker:
Covers:
- [X] Testing
- [x] ExtensionProvider / -Syntax
- [x] IntentionExtensionProvider / -Syntax
- [x] AnActionExtensionProvider / -Syntax
- [x] SyntaxHighlighterExtensionProvider / SyntaxHighlighterSyntax


idea-plugin/src/main/kotlin/arrow/meta/ide/testing/env/IdeTestInterpreter.kt
```diff
@@ -4,18 +4,84 @@ import arrow.meta.ide.testing.IdeTest
...
+ *       * That means an interface such as [LineMarkerSyntax] from the [IdeSyntax] is tested with [LineMarkerTestSyntax] from the [IdeTestSyntax].
+ *       **/
+ *     },
+ *     result = // here we define what behavior is expected
+ *    ),
+ *    ... // you may add more test suit's for the same feature with different code examples and test's
+ *  )
+ * ```
+ *
+ * Given the example ide plugin from [TODO: Add URL] one concrete example may look like this:
```
i-walker: This depends on #262

idea-plugin/src/main/kotlin/arrow/meta/ide/testing/IdeTest.kt
```diff
@@ -5,6 +5,12 @@ import com.intellij.testFramework.fixtures.CodeInsightTestFixture
 
 typealias Source = String
 
+/**
+ * [IdeTest] is a polymorphic aggregation of one complete test suite.
+ * [myFixture] is a key component of the underlying IntelliJ Testing environment.
+ * [test] defines what exact test is run on the [code]
+ * [result] describes the expected shape of [A] with a costume message.
```
JorgeCastilloPrz:
```suggestion
 * [result] describes the expected shape of [A] with a custom message.
```
"costume" or "custom"?

idea-plugin/src/main/kotlin/arrow/meta/ide/testing/env/IdeTestInterpreter.kt
```diff
@@ -4,18 +4,84 @@ import arrow.meta.ide.testing.IdeTest
 import arrow.meta.ide.testing.IdeTestEnvironment
 import org.junit.Assert
 
-fun <A> IdeTest<A>.runTest(interpreter: (IdeTest<A>) -> Unit = ::interpreter): Unit =
+/**
+ * [runTest] executes the tests with a costume [interpreter], which allows users to run their test in every environment.
```
JorgeCastilloPrz:
```suggestion
 * [runTest] executes the tests with a custom [interpreter], which allows users to run their test in every environment.
```

idea-plugin/src/test/kotlin/arrow/meta/ide/plugins/example/ExampleTest.kt
```diff
@@ -0,0 +1,29 @@
...
+  fun `test if lineMarker is displayed`() =
+    ideTest(
+      IdeTest(
+        myFixture = myFixture,
+        code = """
+        | fun hello(): String =
+        |   "Hello world!"
+        """.trimIndent(),
+        test = { code, myFixture ->
+          collectLM(code, myFixture, ArrowIcons.PURE)
```
i-walker: The appropriate Icon will be added here.

idea-plugin/src/main/kotlin/arrow/meta/ide/testing/IdeTest.kt
i-walker: true haha. Thank you @JorgeCastilloPrz

i-walker:
I need reviews on this PR, because it explains the key API's in `Meta` for the editor.
Be blunt. This needs to be as good as the CLI `Meta` Docs. 
Maybe, better 😉

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/action/AnActionSyntax.kt
```diff
@@ -10,55 +10,113 @@ import com.intellij.openapi.actionSystem.TimerListener
...
+ * When a user triggers [AnAction] the [AnAction.actionPerformed] function is executed.
+ * [AnAction] can form starkly different purposes, from changing CompilerConfigurations in the editor to opening UI element's with media content.
+ * Here are a few Links to the Action System from IntelliJ's DevGuid:
+ * - [Plugin Actions](http://www.jetbrains.org/intellij/sdk/docs/basics/plugin_structure/plugin_actions.html?search=act)
+ * - [Executing and updating actions](http://www.jetbrains.org/intellij/sdk/docs/basics/action_system.html?search=act#executing-and-updating-actions)
+ */
 interface AnActionSyntax : AnActionUtilitySyntax {
+
+  /**
+   * Registers the [action] with [actionId] as it's identifier.
```
jansorg: nit: `it's` -> `its`

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/action/AnActionSyntax.kt
```diff
@@ -10,55 +10,113 @@ import com.intellij.openapi.actionSystem.TimerListener
...
 import javax.swing.Icon
 
-// TODO: Check Default of actionId
+/**
+ * [AnActionExtensionProvider] is in charge of the lifecycle of [AnAction].
+ * [AnAction] is an overly broad concept to access and execute nearly anything in the editor.
+ * Put differently, [AnAction] contains a computation which can be evaluated on-demand by the user.
+ * When a user triggers [AnAction] the [AnAction.actionPerformed] function is executed.
+ * [AnAction] can form starkly different purposes, from changing CompilerConfigurations in the editor to opening UI element's with media content.
+ * Here are a few Links to the Action System from IntelliJ's DevGuid:
```
jansorg: nit: `DevGuid` -> `Development Guide` or `DevGuide`?

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/action/AnActionSyntax.kt
```diff
@@ -10,55 +10,113 @@ import com.intellij.openapi.actionSystem.TimerListener
 import com.intellij.openapi.application.ModalityState
 import javax.swing.Icon
 
-// TODO: Check Default of actionId
+/**
+ * [AnActionExtensionProvider] is in charge of the lifecycle of [AnAction].
+ * [AnAction] is an overly broad concept to access and execute nearly anything in the editor.
```
jansorg: That sounds a bit critical on the concept of `AnAction` ;) I agree that it's a broad concept.

imho, the main point of an action is that it provides the functionality to be shown in menus, toolbars, context menus and more. It's sometimes used to just contain an computation on-demand without the ui part, but it's usually connected with a piece of the ui. An action can be reused - a single instance can be added to several places and it's always updated properly (via method `AnAction.update()` and the presentation attached to the event).
Of course, the implementations of an action need to be able to run arbitrary code.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/action/AnActionSyntax.kt
```diff
@@ -10,55 +10,113 @@ import com.intellij.openapi.actionSystem.TimerListener
...
+   *        }
+   *       )
+   *      )
+   *   //sampleEnd
+   *    )
+   *  }
+   * ```
+   *
+   * This Action `Unique` has an Icon and opens a ToolWindow with a registered Id `MetaToolWindow`, assuming this ToolWindow is registered.
+   * User's are now able to search this Action through his [actionId]
```
jansorg: nit: `User's` -> `Users`?
The action in the sample code doesn't have a title, only an icon. Perhaps provide a value for `title` for completeness and easier searching in the actions, e.g. via `Help->Find Action`?

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/intention/IntentionSyntax.kt
```diff
@@ -14,54 +15,107 @@ import org.jetbrains.kotlin.idea.quickfix.KotlinSingleIntentionActionFactory
...
+  /**
+   * unregisters [intention] from the editor
+   */
   fun IdeMetaPlugin.unregisterIntention(
     intention: IntentionAction
   ): ExtensionPhase =
     IntentionExtensionProvider.UnregisterIntention(intention)
 
+  /**
+   * registers an Intention and allows to manipulate the editor environment or [K]
```
jansorg: What does `[K]` stand for?

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/extensions/ExtensionProviderSyntax.kt
```diff
@@ -13,65 +13,179 @@ import com.intellij.openapi.util.ClassExtension
...
+import arrow.meta.ide.dsl.editor.icon.IconProviderSyntax
+import arrow.meta.dsl.analysis.AnalysisSyntax
+import arrow.meta.ide.dsl.editor.lineMarker.LineMarkerSyntax
+import arrow.meta.ide.dsl.editor.search.SearchSyntax
+import arrow.meta.ide.dsl.editor.inspection.InspectionSyntax
 
-
+/**
+ * The [ExtensionProvider] phase sits at the core of the main API in IntelliJ's Platform for ExtensionPoints.
+ * ExtensionPoints regardless of their nature are means to interact with the editor.
```
jansorg: nit: `editor`->`ide`? In IntelliJ wording `editor` is the text editor component to modify a (text) file.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/extensions/ExtensionProviderSyntax.kt
```diff
@@ -13,65 +13,179 @@ import com.intellij.openapi.util.ClassExtension
...
+   *     object : IconProvider(), DumbAware {
+   *     // [DumbAware] signifies that this implementation is available when the editor performs an index update.
+   *       override fun getIcon(p0: PsiElement, p1: Int): Icon? =
+   *         transform(p0, p1)?.run { icon }
+   *      },
+   *     LoadingOrder.FIRST
+   *   )
+   * //sampleEnd
+   * ```
+   * More importantly, using [extensionProvider] and all it's variations lift's any ide workflow for `Extension's` to `Meta` and is evident
```
jansorg: nit: `it's` -> `its`?

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/extensions/ExtensionProviderSyntax.kt
```diff
@@ -13,65 +13,179 @@ import com.intellij.openapi.util.ClassExtension
...
+   * }
+   * //sampleStart
+   * val IdeMetaPlugin.registeringIdeExtensions: Plugin
+   *   get() = "Register ExtensionPoints" {
+   *     meta(
+   *        registerExtensionPoint(MetaProvider.EP_NAME, MetaProvider::class.java)
+   *     )
+   *   }
+   * //sampleEnd
+   * ```
```
jansorg: great example on custom extension points!

idea-plugin/src/main/kotlin/arrow/meta/ide/testing/env/types/HeavyTestSyntax.kt
```diff
@@ -1,3 +1,6 @@
 package arrow.meta.ide.testing.env.types
 
+/**
+ * [HeavyTestSyntax] utilises common patterns in test environments for fully created ide instances
```
jansorg: quick note: with https://github.com/arrow-kt/arrow-meta/pull/252 it'll be (probably) be possible to drop heavy tests

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/extensions/ExtensionProviderSyntax.kt
i-walker: is there another synonym I can use instead?

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/intention/IntentionSyntax.kt
i-walker: [K] is the TypeParameter for this function, in this case any KtElement.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/extensions/ExtensionProviderSyntax.kt
i-walker: thx

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/extensions/ExtensionProviderSyntax.kt
i-walker: Thanks 😄

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/extensions/ExtensionProviderSyntax.kt
jansorg: Perhaps just `IDE` instead of `editor`? IDE is the whole thing, editor a part of it

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/action/AnActionSyntax.kt
i-walker: thx

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/action/AnActionSyntax.kt
i-walker: What do you suggest we should use instead?
I didn't mean to sound critical, though both of us agree on the essence if I am not mistaken. Does this sound better?
```
[AnAction] can be placed, executed and composed into any workflow, 
throughout the whole application lifecycle, 
whether on-demand or as a standalone workflow. 
They're able to access almost every part of the ide.
```
I'd appreciate @jkmcclellan feedback on this 😄

jkmcclellan: Sounds pretty straightforward to me! Your modification does make it clearer.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/action/AnActionSyntax.kt
i-walker: Good call! Makes more sense. I couldn't find it without a title. I thought users can index Actions with `actionId`. Is that somehow possible @jansorg ?

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/extensions/ExtensionProviderSyntax.kt
i-walker: Thank you, I changed it.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/action/AnActionSyntax.kt
jansorg: It‘s not possible, afaik. The id is used internally and not user visible, to my knowledge.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/action/AnActionSyntax.kt
jansorg: Sounds good to me

idea-plugin/src/main/kotlin/arrow/meta/ide/testing/env/types/HeavyTestSyntax.kt
i-walker: Yes, for sure I'd like to keep it, because #252 won't be merged before KotlinConf.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/action/AnActionSyntax.kt
i-walker: Sounds good. Then I'll swap this out.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/intention/IntentionSyntax.kt
```diff
@@ -0,0 +1,159 @@
...
+import org.jetbrains.kotlin.idea.quickfix.KotlinSingleIntentionActionFactory
+import org.jetbrains.kotlin.idea.quickfix.QuickFixContributor
+import org.jetbrains.kotlin.psi.KtElement
+import com.intellij.openapi.actionSystem.AnAction
+
+/**
+ * The IDE analysis user code and provides [IntentionAction]'s to either signal error's to user's or resolve them if triggered.
+ */
+interface IntentionSyntax : IntentionUtilitySyntax {
+  // TODO: [addIntentionWithMetaData] Bails if there is no html and intentionDescription. Add MetaData in a function and not with through the resource folder.
```
jansorg: Should this be kept?

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/syntaxHighlighter/SyntaxHighlighterSyntax.kt
```diff
@@ -0,0 +1,108 @@
...
+import com.intellij.openapi.fileTypes.SyntaxHighlighterFactory
+import com.intellij.psi.tree.IElementType
+import org.jetbrains.kotlin.idea.highlighter.KotlinHighlighter
+import com.intellij.lang.annotation.Annotator
+import com.intellij.lang.PsiParser
+
+/**
+ * [SyntaxHighlighterExtensionProvider] provides means to register [SyntaxHighlighter]s for Languages.
+ * SyntaxHighlighters solely specify highlighted tokens, which are generated from the Lexer.
+ * [ColorSyntax] allows to highlight additional descriptors from the `Parser` or [Annotator]. You may create a `Parser` with [PsiParser] an example is here [org.jetbrains.kotlin.parsing.KotlinParser].
```
jansorg: I'd remove the note about the parser, as it might be a bit confusing to new users.
Creating a parser isn't helping the user to customize highlighting for an existing file type/language.

The parser is registered when the language is registered (here, by the Kotlin plugin). Even though a parser and annotator use the additional highlighting information, the parser is usually not extendable.

Customization of editor highlighting is usually implemented with an annotator, unless the user is implementing a new parser for a new language.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/extensions/ExtensionProviderSyntax.kt
```diff
@@ -13,65 +13,181 @@ import com.intellij.openapi.util.ClassExtension
...
+   * @see [registerExtensionPoint] for [ExtensionPointName]
+   */
   fun <E> IdeMetaPlugin.registerExtensionPoint(
     EP_NAME: BaseExtensionPointName,
     aClass: Class<E>
   ): ExtensionPhase =
     ExtensionProvider.RegisterBaseExtension(EP_NAME, aClass)
 
+  /**
+   * Interestingly enough, [ExtensionProvider] registers new workflows to the ide.
```
jansorg: A thought on this which came to me when I read this.

Custom extension points can be used in two ways, at least.
1. Add the custom extension point to allow 3rd-party plugins to implement it.
2. Use the custom extension point in the same plugin where it's defined. This is used in many JetBrains plugins.

Usually, custom extension points are defined in the `plugin.xml`. With this data, IntelliJ is able to offer code completion and validation. 
Therefore, adding custom extension points might be one of the few things which should be defined in the plugin.xml for better interoperability, at least if it's added to allow other plugins to use it.

I didn't test, but 3rd-party plugins should be able to use custom extension points via plugin.xml, which were defined this way. They just need to know the name and need to ignore the error markers in the `plugin.xml` file.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/extensions/ExtensionProviderSyntax.kt
```diff
@@ -13,65 +13,181 @@ import com.intellij.openapi.util.ClassExtension
...
+   */
   fun <E> IdeMetaPlugin.registerExtensionPoint(
     EP_NAME: ExtensionPointName<E>,
     aClass: Class<E>
   ): ExtensionPhase =
     ExtensionProvider.RegisterExtension(EP_NAME, aClass)
 
-  fun IdeMetaPlugin.addContainerProvider(f: (PsiElement) -> PsiElement?): ExtensionPhase =
+  /**
+   * Add's a [ContainerProvider].
```
jansorg: nit: Add's -> Adds?

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/intention/IntentionSyntax.kt
i-walker: Good call let's deprecate it until we have a stable version

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/extensions/ExtensionProviderSyntax.kt
i-walker: The goal for `Meta` regarding the Ide is to circumvent all that noise, which happens in the plugin.xml. Big Ide Plugin's like `KotlinPlugin` have not only one `plugin.xml`, but multiple `sub.xml` files, which has different purposes and depend on each other. This is a major bug and unnecessary mental overhead for plugin developers in the IntelliJ-Platform. Let alone, the added complexity, which comes with the cost of parsing and instantiating all `Extension` instances - most of that logic is kept in the `PicoContainer`. 

Observing, the internals class definitions are not important at RunTime .
The IntellIJ-Platform constructs each Plugin as a DataStructure, where the focus is the value of the DT not where it comes from. Every computation we add is kept in that DT as a [ure value, which means the ide traverses and folds this DT, depending on the current lifecycle. 

Obviously, the way `Meta` approaches the `IntelliJ-Platform` is different from what is out there, but it comes with the advantage to be more explicit about the features your building. 
If we're able to flatten this complexity, we can build better APIs in the future, and that is possible with libraries like `Meta`.

i-walker: The last comment is what we observed about the key components and core philosophy in the `IntelliJ-Platform`. With Kotlin, we can exploit those flaws and construct everything as a function or value.

Regarding your comment:
You can still do everything that you mentioned. Only that you don't have to host all your extensions in the `plugin.xml` and you have a lot better visibility to new-comers if all your workflows and extensions are described in code rather than in Xml files scattered around the project.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/syntaxHighlighter/SyntaxHighlighterSyntax.kt
i-walker: Agree, I think I was too excited, once you send me the Docs about `Annotator` 😄

i-walker:
We have issues with `Ank` resolving Descriptors from `IntelliJ` and `Kotlin`
For that we out commented the examples. We will undo that once that bug is fixed

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/extensions/ExtensionProviderSyntax.kt
jansorg: Thanks for the explanations on Meta's core philosophy.
The API is in an early stage and I bet that it'll be easier to use in the future! The loss of IntelliJ's tool support in plugin.xml might be added by the Meta plugin to the plugin API offered by Meta.

@i-walker 
There's one thing which will still need an xml file is optional dependencies. At least I don't see another way to implement this without a sub .xml file. See http://www.jetbrains.org/intellij/sdk/docs/basics/plugin_structure/plugin_dependencies.html?search=dep#optional-plugin-dependencies for details on this.

i-walker: May you reiterate on this, which functionality are you referring to here?:
```
The loss of IntelliJ's tool support in plugin.xml might be added by the Meta plugin to the plugin API offered by Meta.
```
We can investigate `optional dependencies`. Most of the time there is an API in the internals, who handles that part. We're not against the `plugin.xml`, but `Meta` tries to enhance the developer experience and if declaring and maintaining `optional dependencies` in xml files is a pain, we might find a way to resemble it in `Meta`.
btw, if you or anyone reading have requests and doubts on API features what `Meta` might be missing, we'll be glad if you open an Issue for it and add the appropriate label to it.

PR #281, opened 2019-12-02
HintingSyntax Docs

i-walker:


idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/hints/HintingSyntax.kt
```diff
@@ -14,114 +14,374 @@ import com.intellij.lang.parameterInfo.ParameterInfoHandlerWithTabActionSupport
...
+   *  get() = "Hints for KtExpressions" {
+   *   meta(
+   *    addExpressionTypeProviderForKotlin(
+   *     informationHint = { expr ->
+   *      get(BindingContext.DECLARATION_TO_DESCRIPTOR, expr).safeAs<CallableDescriptor>()?.returnType
+   *       ?.let { result: KotlinType ->
+   *         typeRenderer.renderType(result)
+   *        } ?: "Type is unknown"
+   *    },
+   *    expressionAt = {
```
jansorg: I have to admit that I don't understand this example. I've no experience with the extension points of the Kotlin plugin.
The docs of Kotlin's expressionAt is
```
  /**
   * Returns the list of all possible targets at specified position.
   */
  @NotNull
  public abstract List<T> getExpressionsAt(@NotNull PsiElement elementAt);
```

This example isn't using the input element passed to `expressionAt`, but is using the current editor instead. If this is correct, then as a user reading this I'd like to know why the input parameter isn't used.

Is it certain that this function is always called from an editor? Accessing the editor might be a bit expensive, depending on how often this is called (I didn't check).

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/hints/HintingSyntax.kt
```diff
@@ -14,114 +14,374 @@ import com.intellij.lang.parameterInfo.ParameterInfoHandlerWithTabActionSupport
...
+   * //sampleEnd
+   *
+   * private val IdeMetaPlugin.typeRenderer: DescriptorRenderer
+   *  get() = DescriptorRenderer.COMPACT_WITH_SHORT_TYPES.withOptions {
+   *    textFormat = RenderingFormat.HTML
+   *    classifierNamePolicy = classifierNamePolicy()
+   *  }
+   *
+   * /**
+   * * evaluates the SelectionTextRang of [PsiElement]'s`containedFile`,
```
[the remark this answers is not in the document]
i-walker: thx

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/hints/HintingSyntax.kt
i-walker: It is you may have overwritten it but `it` in the first line is `elementAt` in your example. 
in `Meta` most of these functions are flatten for a straight forward implementation.

jansorg: Thanks. 
So this is `KotlinExpressionTypeProvider.getExpressionAt(elementAt: PsiElement)` with a few modifications. `KotlinExpressionTypeProvider`  also uses the passed PsiElement only to get the editor.

i-walker: It doesn't use the editor. it solely passes valid KtExpressions, which are at the beginning of the textRange.
It collects all valid KtFunctions and returns the one which starts at the same `startoffSet`.

jansorg: I meant that it only uses `selectionTextRangeOnTextEditor()` on the input element, which retrieves the selected text range from the editor, where the containing file is edited.

i-walker: exactly :)

PR #282, opened 2019-12-02
removing underdeveloped DSL's

i-walker:
These uncurated DSL's were kept unfinished, due to several reasons during development and require additional time after KotlinConf: 
- QuickFixSyntax 
- Refactoring 
- LanguageInjectionSyntax 
- LanguageSyntax 
- RunLineMarker

We may exclude them temporarily before KotlinConf, so we're not running into issues with bad user experience.
If it so happens, that I have more time before the conference, I'll try to reestablish them into `IdeSyntax`. Otherwise, we'll see them evolve at a later point in time.

### DIFF
--- a/compiler-plugin/src/main/kotlin/arrow/meta/dsl/analysis/AnalysisSyntax.kt
+++ b/compiler-plugin/src/main/kotlin/arrow/meta/dsl/analysis/AnalysisSyntax.kt
@@ -26,15 +26,15 @@ import java.util.ArrayList
 
 /**
  * The Analysis phase determines if the parsed AST type checks and resolves properly.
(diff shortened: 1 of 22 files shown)


i-walker:
I discovered a Bug with RunLineMarker. The registration process was wrong and it won't be ready for KotlinConf. It has to do with an uncovered case in the `IdeInternals`. There are 2 different Registration methods for LanguageExtensions in the ide. The interpreter covers cases like `LineMarkerProvider, which is a LanguageExtension, but fails to register `bean` Extensions. I'll revisit this at a later point. AFAIK, RunLineMarkers are the only Extension in our DSL who have that need that special registration.

PR #298, opened 2019-12-04
aAdded DotQualifiedExpression Scope and template plus testing validation

ahinchman1:


compiler-plugin/src/test/kotlin/arrow/meta/quotes/scope/plugins/DotQualifiedExpression.kt
```diff
@@ -0,0 +1,27 @@
+package arrow.meta.quotes.scope.plugins
+
+import arrow.meta.Meta
+import arrow.meta.Plugin
+import arrow.meta.invoke
+import arrow.meta.phases.CompilerContext
+import arrow.meta.quotes.Transform
+import arrow.meta.quotes.dotQualifiedExpression
+
+open class DotQualifiedExpressionPlugin : Meta {
```
raulraja: Unless you plan to override these ad-hoc testing plugins in other places this can be just final by fault visibility instead of an open class.

PR #306, opened 2019-12-06
Update Ide to latest eap

### LINKED ISSUE #218: Bump Ide to 2019.3
Kotlin released 1.3.60 with 2.1, which we support, but there will be another release soon with 2019.3. We need to update and refactor the existing implementation. 
We may consider applying a similar workflow for Dev or EAP versions in the editor as we have it in the compiler, e.g.: `Github-Bot` creates PR's and Issues for upgrades.

i-walker:
resolves #218 and updates the Ide to the lates SnapShot automatically.
This method implies that at anytime one snapShot may break a PR.

This version has a weird error, it is not possible to register any Extensions wether with Meta or vanilla IntelliJ to the editor.
reproduce the problem: 
-  checkout of this branch
- runIde

the OpticsIdePlugin LineMarker should at least be added, becausse it uses vanilla IntelliJ.

My guess is that the gradle intelliJ plugin is not updated to 2019.3, because it can not be resolved in our plugin.xml dependencies.


idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/optics/OpticsIdePlugin.kt
```diff
@@ -18,3 +23,10 @@ val IdeMetaPlugin.opticsIdePlugin: Plugin
       )
     )
   }
+
+class OpticsIdePlugin : LineMarkerProvider, LineMarkerSyntax {
```
i-walker: Even though we register an Intention with vanilla `intelliJ`, the optics tests still fail. You may test it with an example project and this code:
```
data class TestLenses(val a: String, val b: String)

data class Person(
    val name: String,
    val age: Int,
    val home: Home
)

data class Home(val street: String)
```

i-walker:
The problem were some issues with the unresolved gradle dependency. 
The interpreter for Meta was not changed, but in an upcoming PR, we will open the Scope of Meta to any Language. Currently, Language Extensions and FileTypeExtensions are hardcoded for Kotlin.

rachelcarmena:
Thanks @i-walker :tada:

PR #308, opened 2019-12-07
remove TODO from example

i-walker:
Thank you for pointing it out :)


i-walker:
Closing because #304 resolved it

PR #311, opened 2019-12-09
FunctionLiteral Quote and Template Scope Testing Validation

ahinchman1:
### Checklist for `KtFunctionLiteral`

### DIFF
--- a/compiler-plugin/src/test/kotlin/arrow/meta/quotes/scope/FunctionLiteralTest.kt
+++ b/compiler-plugin/src/test/kotlin/arrow/meta/quotes/scope/FunctionLiteralTest.kt
@@ -19,7 +19,7 @@ class FunctionLiteralTest : AnnotationSpec() {
 
   @Test
   fun `Validate function literal as an anonymous function`() {
-    // TODO
+    validate("""val increment: (Int) -> Unit = fun(x) { x + 1 }""".functionLiteral())
   }
 
   @Test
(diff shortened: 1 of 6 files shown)


compiler-plugin/src/test/kotlin/arrow/meta/quotes/scope/FunctionLiteralTest.kt
```diff
@@ -0,0 +1,55 @@
...
+
+import arrow.meta.plugin.testing.Code
+import arrow.meta.plugin.testing.CompilerTest
+import arrow.meta.plugin.testing.CompilerTest.Companion.source
+import arrow.meta.plugin.testing.assertThis
+import arrow.meta.quotes.scope.plugins.FunctionLiteralPlugin
+import io.kotlintest.specs.AnnotationSpec
+
+/**
+ * Function Literals are a little unclear - see more here: https://kotlinlang.org/docs/reference/lambdas.html
```
raulraja: I think you should just add here what you find confusing. In my case what I find most confusing is the many ways in which you can define function values but I understand there are ergonomic reasons and also it caters to a broader audience allowing more shapes and styles.

ahinchman1: I think I have some examples of clearer ideas I've added in this morning!

PR #314, opened 2019-12-09
version rename

i-walker:
To prevent confusion we should rename the property to what it is.
The actual IntelliJ version is currently set to the latest EAP-SNAPSHOT.


rachelcarmena:
Small and very useful change, thanks @i-walker :tada:

PR #318, opened 2019-12-10
Lift Application Services to Meta

i-walker:
This PR lifts application services to Meta, to In order for `Meta` to not only host ide-plugins but also add additional Services on-demand.

In particular, Services are lightweight Components, which allow users to add ide-plugins without reloading. Resulting, in a smooth developer experience to work with `Meta` and libraries who depend on it. 
More importantly, `Jetbrains` is moving away from the legacy component-based architecture of ide-plugins to promote the experience from above. Interestingly enough, Services are on-demand and are not loaded at application start or when opening a Project - as we see in popular Plugins. Services and their Extensions are an on-demand plugin based on rules we can define.
These changes are already introduced in 2019.3 - what we're running rn, but are planned to be manifested in the ecosystem after 2020.1. 
This change is not surprising, as other vendors like VSC has been introducing this to the market for years.


i-walker:
Closing this temporarily. I'll reopen if this once it is part of the project scope.

PR #324, opened 2019-12-11
ThisExpression quote and template

ahinchman1:
### Checklist for `KtThisExpression`


ahinchman1:
Sorry @raulraja I'm going to open another PR because changes from the master had been merged in accidentally instead of a rebase and it's causing the files to freak out!

PR #326, opened 2019-12-12
Classbody quote and scope

bloderxd:
### Checklist for `KtClassBody`

#89


compiler-plugin/src/main/kotlin/arrow/meta/internal/kastree/ast/Writer.kt
```diff
@@ -80,26 +80,13 @@ open class Writer(
...
-              lineBegin().also { children(enumEntry) }
-              when (index) {
-                members.size - 1 -> lineEnd()
-                enumEntries.size - 1 -> lineEnd(";").line()
-                else -> lineEnd(",")
-              }
-            }
-            // Now the rest of the members
-            childrenLines(members.drop(enumEntries.size), extraMidLines = 1)
-          }.lineBegin("}")
```
ahinchman1: @bloderxd I was wondering if I could get a little more context for removing this part regarding `EnumEntries`?

ahinchman1: ohh - I see it got moved down under ClassBody.  I'd still like to hear more about it!

compiler-plugin/src/test/kotlin/arrow/meta/quotes/scope/ClassBodyTest.kt
```diff
@@ -0,0 +1,130 @@
...
+           |   private val x = "x"
+           |   private val y = "y"
+           |   private val z = "z"
+           |
+           |   fun x() = x
+           |   fun y() = y
+           |   fun z() = z
+           | }
+           """.source
+  }
```
ahinchman1: This looks good! We're going to want to add more testing for increased robustness i.e. moved order of companion objects, missing components of the test, etc.

compiler-plugin/src/main/kotlin/arrow/meta/internal/kastree/ast/Writer.kt
daniel-leme-ifood: As we've created a specific `ClassBody` AST node we start to provide a separated scope to that, then what I did here was just traverse its node children instead of handle it as a `List<Decl>`, because now we have this list in a specific class body node

bloderxd: @ahinchman1 As we've created a specific ClassBody AST node we start to provide a separated scope to that, then what I did here was just traverse its node children instead of handle it as a List<Decl>, because now we have this list in a specific class body node

PR #327, opened 2019-12-12
TypeAlias Quote and Template Testing Validation

ahinchman1:
### Checklist for `KtTypeAlias`

### DIFF
--- a/compiler-plugin/src/main/kotlin/arrow/meta/internal/kastree/ast/psi/Converter.kt
+++ b/compiler-plugin/src/main/kotlin/arrow/meta/internal/kastree/ast/psi/Converter.kt
@@ -931,14 +931,14 @@ open class Converter {
 internal val PsiElement.ast: Node get() = when(this) {
   is KtClassOrObject -> Converter.convertDecl(this)
   is KtNamedFunction -> Converter.convertFunc(this)
-  is KtExpression -> Converter.convertExpr(this)
+  is KtTypeAlias -> Converter.convertTypeAlias(this)
   is KtWhenCondition -> Converter.convertWhenCond(this)
   is KtWhenEntry -> Converter.convertWhenEntry(this)
   is KtCatchClause -> Converter.convertTryCatch(this)
   is KtImportDirective -> Converter.convertImport(this)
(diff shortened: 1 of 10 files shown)


compiler-plugin/src/main/kotlin/arrow/meta/internal/kastree/ast/psi/Converter.kt
```diff
@@ -926,13 +926,13 @@ open class Converter {
...
-  is KtExpression -> Converter.convertExpr(this)
+  is KtTypeAlias -> Converter.convertTypeAlias(this)
   is KtWhenCondition -> Converter.convertWhenCond(this)
   is KtWhenEntry -> Converter.convertWhenEntry(this)
   is KtCatchClause -> Converter.convertTryCatch(this)
   is KtImportDirective -> Converter.convertImport(this)
   is KtValueArgument -> Converter.convertValueArg(this)
-  is KtTypeAlias -> Converter.convertTypeAlias(this)
   is KtTypeReference -> Converter.convertTypeRef(this)
+  is KtExpression -> Converter.convertExpr(this)
```
ahinchman1: TypeAlias can also be considered an Expression, so we're going to want larger fragmented PSI elements to go towards the bottom.

compiler-plugin/src/test/kotlin/arrow/meta/quotes/scope/TypeAliasTest.kt
```diff
@@ -1,27 +1,53 @@
...
+
+  @Ignore // issues with type erasure when AST parsing the compiled code?
+  @Test
+  fun `Validate type alias properties with generics`() {
+    val typeAlias = """
+                         | //metadebug
+                         | 
+                         | typealias Predicate<T> = (T) -> Boolean
+                         | """.source
+
```
ahinchman1: As expected, analysis for compilation erases generics at runtime.  @raulraja do you think there is a way to work around this?

raulraja: What is the difference in the output or exception stack-trace that makes the test fail?

PR #333, opened 2019-12-12
Fix: IfExpression testing and added validation

ahinchman1:
### Checklist for `KtIfExpression`
 - [x] Add documentation for element, including an example using all properties

#89 + #202

### DIFF
--- a/compiler-plugin/src/main/kotlin/arrow/meta/phases/analysis/ElementScope.kt
+++ b/compiler-plugin/src/main/kotlin/arrow/meta/phases/analysis/ElementScope.kt
@@ -271,12 +271,6 @@ interface ElementScope {
   
   fun String.blockCodeFragment(context: PsiElement?): Scope<KtBlockCodeFragment>
   
-  fun `if`(
-    condition: KtExpression,
-    thenExpr: KtExpression,
-    elseExpr: KtExpression? = null
-  ): IfExpression
-  
   fun argument(
     expression: KtExpression?,
     name: Name? = null,
--- a/compiler-plugin/src/main/kotlin/arrow/meta/quotes/expression/IfExpression.kt
+++ b/compiler-plugin/src/main/kotlin/arrow/meta/quotes/expression/IfExpression.kt
@@ -6,7 +6,7 @@ import org.jetbrains.kotlin.psi.KtExpression
 import org.jetbrains.kotlin.psi.KtIfExpression
 
 /**
- * <code>"""if ($condition) $`else`""".`if`</code>
+ * <code>"""if ($condition) $then else $`else`""".`if`</code>
  *
  * A template destructuring [Scope] for a [KtIfExpression].
  *
@@ -18,17 +18,20 @@ import org.jetbrains.kotlin.psi.KtIfExpression
  * import arrow.meta.quotes.ifExpression
  *
  * val Meta.reformatIf: Plugin
- *  get() =
- *  "ReformatIf" {
- *   meta(
- *    ifExpression({ true }) { e ->
- *     Transform.replace(
- *      replacing = e,
- *      newDeclaration = """if ($condition) $`else`""".`if`
- *     )
- *    }
- *   )
- *  }
+ *    get() =
+ *      "Reformat If Expression" {
+ *        meta(
+ *          ifExpression({ true }) { expression ->
+ *            Transform.replace(
+ *              replacing = expression,
+ *              newDeclaration = when {
+ *                `else`.value == null -> """if ($condition) $then""".`if`
+ *                else -> """if ($condition) $then else $`else`""".`if`
+ *              }
+ *            )
+ *          }
+ *        )
+ *      }
  *```
  */
 class IfExpression(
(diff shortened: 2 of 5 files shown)


compiler-plugin/src/main/kotlin/arrow/meta/phases/analysis/ElementScope.kt
```diff
@@ -276,12 +276,6 @@ interface ElementScope {
   
   fun String.blockCodeFragment(context: PsiElement?): Scope<KtBlockCodeFragment>
   
-  fun `if`(
-    condition: KtExpression,
-    thenExpr: KtExpression,
-    elseExpr: KtExpression? = null
-  ): IfExpression
```
ahinchman1: I'm starting to favor using Expression shapes over using some of these methods from `PsiFactory`. For example, there is an expectation that `condition` and `thenExpr` cannot be nullable for the following method:
```
    fun createIf(condition: KtExpression, thenExpr: KtExpression, elseExpr: KtExpression? = null): KtIfExpression {
        return (if (elseExpr != null)
            createExpressionByPattern("if ($0) $1 else $2", condition, thenExpr, elseExpr) as KtIfExpression
        else
            createExpressionByPattern("if ($0) $1", condition, thenExpr)) as KtIfExpression
    }
```
but this does not work well naturally with the PsiElement since the following are marked nullable:
```
public class KtIfExpression extends KtExpressionImpl {
    ...
    @Nullable @IfNotParsed
    public KtExpression getCondition() {
        return findExpressionUnder(KtNodeTypes.CONDITION);
    }
    ...
    @Nullable
    public KtExpression getThen() {
        return findExpressionUnder(KtNodeTypes.THEN);
    }
   ...
   @Nullable
    public KtExpression getElse() {
        return findExpressionUnder(KtNodeTypes.ELSE);
    }
}
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/expression/IfExpression.kt
```diff
@@ -17,22 +17,25 @@ import org.jetbrains.kotlin.psi.KtIfExpression
...
+ *        )
+ *      }
  *```
  */
 class IfExpression(
-  override val value: KtIfExpression?,
-  val condition: Scope<KtExpression> = Scope(value?.condition),
-  val then: Scope<KtExpression> = Scope(value),
-  val `else`: Scope<KtExpression> = Scope(value?.`else`)
+  override val value: KtIfExpression,
```
bloderxd: I don't think we should let scopes value property as non nullables, if we need to provide in somewhere a nullable `KtIfExpression` reference we will need to let the scope itself as nullable like: `Scope<KtIfExpression>?`, this breaks our scope contract and we have nullable values handling in its `toString` function

compiler-plugin/src/test/kotlin/arrow/meta/quotes/scope/plugins/IfExpressionPlugin.kt
```diff
@@ -20,9 +20,11 @@ val Meta.ifExpressionPlugin
         ifExpression({ true }) { expression ->
           Transform.replace(
             replacing = expression,
-            newDeclaration = """if ($condition) $`else`""".`if`  // {"""$then""".`if`} also works
+            newDeclaration = when {
+              `else`.value == null -> """if ($condition) $then""".`if`
+              else -> """if ($condition) $then else $`else`""".`if`
+            }
```
rachelcarmena: This declaration is located in `identity` function for `IfExpression` class.

compiler-plugin/src/test/kotlin/arrow/meta/quotes/scope/IfExpressionTest.kt
```diff
@@ -22,7 +22,11 @@ class IfExpressionTest : AnnotationSpec() {
 
   @Test
   fun `Do not validate if expression without else scope properties`() {
-    // TODO
+    validate(
+      """
+      | if (2 == 3) {
+      |   println("FAKE NEWS")
+      | }""".ifExpression())
```
rachelcarmena: Code examples for `IfExpression` are located in:
```
./compiler-plugin/src/test/kotlin/arrow/meta/quotes/scope/expressions/IfExpression.kt
```

And this comment can be removed with this PR:

```
// TODO: Do not validate if expression without else scope properties
```

compiler-plugin/src/main/kotlin/arrow/meta/quotes/expression/IfExpression.kt
ahinchman1: @bloderxd I think that sounds good - let's definitely visit this because I have some thoughts as well - the only thing holding us back from this choice is depending on the `PsiFactory`, but I think as we move away from that we will have the ability to make better choices like how you suggested!

compiler-plugin/src/main/kotlin/arrow/meta/phases/analysis/ElementScope.kt
```diff
@@ -271,12 +271,6 @@ interface ElementScope {
   
   fun String.blockCodeFragment(context: PsiElement?): Scope<KtBlockCodeFragment>
   
-  fun `if`(
```
raulraja: I think this method makes more sense if instead of taking `KtExpression` values it takes our scope values instead

PR #334, opened 2019-12-12
Quote and template minor cleanups

ahinchman1:
Review current code for elements not covered, update scopes, reformat and clarify documentation. Tidying up loose ends if possible!


compiler-plugin/src/test/kotlin/arrow/meta/quotes/scope/IfExpressionTest.kt
```diff
@@ -50,3 +54,11 @@ class IfExpressionTest : AnnotationSpec() {
       | """.source
   }
 }
+
+class Wrapper {
```
bloderxd: What is this code is used for?

ahinchman1: This should be cleaned up!

PR #337, opened 2019-12-15
Quote testing: re-organization

rachelcarmena:
# Main changes

- Include a generic plugin which replaces all the quote testing plugins: `arrow.meta.quotes.scope.plugins.GenericPlugin`.
- Include a generic test which replaces all the quote tests: `arrow.meta.quotes.scope.QuoteTest`.
- Move new declarations to `identity`.
- Recover the previous `Test` annotation in order to have single test execution.
- Temporary functions that will be removed when finishing all the quotes.

# Screenshots

## From

Screenshot from 2019-12-16 01-34-00

## To

Screenshot from 2019-12-16 01-34-14

# Other changes

- Rename `ImportDirective` because `ImportList` exists.
- Add missing element to `internal val PsiElement.ast: Node get()` though there are other missing elements.


compiler-plugin/src/main/kotlin/arrow/meta/quotes/Scope.kt
```diff
@@ -17,6 +18,8 @@ open class Scope<out K : KtElement>(open val value: K?) {
   companion object{
     fun <A> empty() = Scope(null)
   }// java null snicking in
+
+  open fun ElementScope.identity(): Scope<K> = Scope(value) // TODO:  TODO("not implemented")
```
raulraja:
```suggestion
  open fun ElementScope.identity(): Scope<K> = Scope(value)
```

compiler-plugin/src/test/kotlin/arrow/meta/quotes/scope/plugins/GenericPlugin.kt
```diff
@@ -0,0 +1,108 @@
...
+          },
+          { transform(it) }
+        )
+      )
+    }
+  )
+
+  // TODO: remove at the end
+  private fun implemented(ktElement: KtElement): Boolean = (
+    ktElement is KtBinaryExpression
```
raulraja: Optionally replace with `when (ktElement) { is KtBlockExpression.. }` and it may read better

PR #354, opened 2019-12-17
Sync versions: Kotlin, IDEA plugin and Intellij IDEA

rachelcarmena:
## Changes

- Extract Intellij IDEA version from `idea-plugin/build.gradle` to `gradle.properties`
- Sync versions: 
  - Kotlin
  - Kotlin IDEA plugin
  - Intellij IDEA

## Notes

- @i-walker, now, `LATEST-EAP-SNAPSHOT` = `193.5662.46` build number = `2019.3.1` version number. If we use the version number, it will be easier to get the Kotlin IDEA plugin version. It's important to sync the 3 versions to avoid problems during development.
- **Next step**: update the bot to consider the latest release `1.3.61` and to keep the 3 versions.


gradle.properties
```diff
@@ -4,8 +4,9 @@ VERSION_NAME=1.3.60-SNAPSHOT
 # Dependencies versions
 ARROW_VERSION=0.10.4-SNAPSHOT
 ANK_GRADLE_PLUGIN_VERSION=0.7.1
-KOTLIN_VERSION=1.3.60-eap-143
-KOTLIN_IDEA_VERSION=1.3.61-release-IJ2019.3-1
+KOTLIN_VERSION=1.3.60
+KOTLIN_IDEA_VERSION=1.3.60-release-IJ2019.3-1
+INTELLIJ_IDEA_VERSION=2019.3.1
```
i-walker: The goal is to stay at `LATEST-EAP-SNAPSHOT`, so that the ide can adapt continuously to the IntelliJ Platform. I suggest once we release we can tag the release but not for development. 
Otherwise, there is always a huge effort to update to the latest version - and delays release cycles. 

I am open to Feedback.
What do you think @raulraja ?

rachelcarmena: Of course, it's just a change of name. 

`LATEST-EAP-SNAPSHOT` = `193.5662.46` build number = `2019.3.` version number. 

If we use the version number, it will be easier to get the Kotlin IDEA plugin version. 

It's important to sync the 3 versions to avoid problems during development. 

The bot will update the 3 versions.

i-walker: Nop, we have to stay on LATEST-EAP-SNAPSHOT. The actual version number changes everytime they release and the build.number is only for stable major releases AFAIK. Unless the bot can detect the latest EAP release, and merges it as we do with the Kotlin Compiler. The actual workflow has not been communicated.

raulraja: @i-walker this is not the goal. please read @rachelcarmena comment. Those are the same version and we need to go by version number not release name to support the compatibility matrix the Kotlin plugin currently supports. Raquel is in charge of bringing that to Meta.

rachelcarmena: Thanks for the review, Imran, I understand your concerns, however we cannot use an EAP Intellij IDEA version without the corresponding Kotlin IDEA plugin.

it's necessary to sync the 3 versions: Kotlin, Kotlin IDEA plugin and Intellij IDEA.

This change comes from a problem raised during Type Proofs development.

i-walker: Makes sense.

PR #360, opened 2019-12-18
When expression  quote and template validation

matiaslev:
#348 When expression identity changes and tests


compiler-plugin/src/main/kotlin/arrow/meta/quotes/expression/WhenExpression.kt
```diff
@@ -44,8 +44,7 @@ class WhenExpression(
   val `else`: Scope<KtExpression> = Scope(value?.elseExpression)
 ) : Scope<KtWhenExpression>(value) {
   override fun ElementScope.identity(): Scope<KtWhenExpression> =
-    """
-    | when $`(expression)`{ 
-    |   $entries
-    | }""".`when`
+    """|when ${if (`(expression)`.toString().isNotEmpty()) "($`(expression)`)" else `(expression)`} {
```
raulraja: Would it be better if $expression renders conditionally to `(expression)` or just ``?
that can be controlled by wrapping this expression in a new scope `WhenExpressionTarget` that knows how to render itself in the presence of a nullable expression. Normally whenever a template needs to do some fiddling to render itself beside printing the variables is usually an indication that there is a missing intermediary model scope that Meta can properly type better than PSI.

raulraja: If you think that is worth exploring we can add that in a separate scope. The reference design here is that the template that is defined in `identity` is the same template we would expect the user to use to destructure so we want to try to avoid method calls and things like that to keep it as a clean as possible even if that means more scopes.

matiaslev: thanks for your comments! :)
I'm was wondering how to clean it, I am going to see if I can do it with this idea as a guide.

matiaslev: please let me know if this makes sense, feels better, identity is clean :)
https://github.com/arrow-kt/arrow-meta/pull/363

PR #367, opened 2019-12-19
Sync: test latest Kotlin EAP and DEV versions

rachelcarmena:
- Remove EAP repository from `build.gradle` by default.
- Add DEV or EAP repository according to the test.
- Check `compiler-plugin` with latest Kotlin DEV version and latest Kotlin EAP version. 
- Create an issue in case of failure.


rachelcarmena:
Ready to be merged :tada:

Issue #256, opened 2019-11-27
Auto install Gradle plugin from IDEA plugin


rachelcarmena:
Use case:
- Arrow Meta IDEA plugin is installed and a project doesn't have any reference to Arrow Meta

Solution:
- Show global dialog: "This project doesn't have Arrow Meta included"
- Yes -> Include Gradle plugin

Links:
* https://www.jetbrains.org/intellij/sdk/docs/user_interface_components/notifications.html

rachelcarmena:
When using `GradleProjectImportHandler`:
* It'll just work when loading the Gradle data the first time. Once the project is cached, it won't execute again.
* A way to know if Arrow Meta Gradle Plugin is loaded: reviewing extensions.

rachelcarmena:
I don't find the way to change the `build.gradle` with the plugin. Other things: it can be opened or not, it's possible to have several build.gradle.

As a first solution, I'm going to show a dialog with:
- The name of the project which hasn't included the Gradle Plugin.
- How to include it.

rachelcarmena:
Doing tests with a single project and a multi-module project.

rachelcarmena:
First approach: #379

rachelcarmena:
Discarded in favor of #245

PR #379, opened 2019-12-24
IDEA Plugin: show notification if Gradle Plugin is missing

rachelcarmena:
First approach for #256

## Changes

- If Arrow Meta Gradle Plugin is missing in a Gradle project (or sub-project for multi-project), a notification will be shown in its associated frame (`Notifications.Bus.notify()` is used with Project parameter).
- The notification shows the name of the project where the plugin is missing. If there are more than one project in that situation, a notification for each project is shown.
- The notification includes an action to copy the code to add the plugin.
- The notification includes an action to open the corresponding file where plugin is missing.
- The proposed version for Gradle Plugin will be the same as IDEA Plugin.

### One notification
Screenshot from 2019-12-24 20-28-41

Screenshot from 2019-12-24 20-28-49

### Event Log with one notification

Screenshot from 2019-12-24 20-29-10

### Several notifications

If it's missing in several modules

Screenshot from 2019-12-24 20-25-48

### Event Log with several notifications
Screenshot from 2019-12-24 20-26-13

### New entry in notifications window

Screenshot from 2019-12-24 15-22-46


raulraja:
There are many cases where the user may have the Arrow plugin for IDEA already nstalled but not necessarily wants to turn their project into a project that is managed by arrow-meta.

This is extremely useful if we can detect the code uses some of the arrow meta features but it's not in the Gradle file but otherwise it can't be for all cases always showing this notification since there is many legit cases for not wanting to have meta in a project.

raulraja:
Maybe we should describe all user scenarios regarding installation as a flow and issue to make sure we cover all the use cases of the Gradle plugin and the idea plugin

PR #386, opened 2019-12-31
synthetic renamed to syntheticScope and syntheticElement added for use Property inside LensPlugin

matiaslev:
In order to replace Scope<KTProperty> to Property in LensPlugin I performed these changes


rachelcarmena:
> Thanks @matiaslev! I suppose this will require changes in the hello world plugins that are calling now `.synthetic`. //cc @rachelcarmena

Thanks for the warning, @raulraja !! :+1: 

Thanks @matiaslev!!

PR #387, opened 2020-01-02
Allow multiple Languages in Meta

i-walker:
Allow implementations in multiple Languages for FiletypeExtensions and LanguageExtension.


i-walker:
Thanks @jansorg. We will consider that, if that topic comes into consideration :)

PR #388, opened 2020-01-02
Deprecations and Improvements on Intentions

i-walker:
- The Java API's on registering Intentions with MetaData is deprecated and will be deleted
Furthermore, we can still achieve the same things with Intentions without Metadata, by having `KtPsiFactory` as a receiver in `applyTo`. This allows us to do the same things as we do with `Inspections`

- FileTemplate Syntax should be deleted as it does not add any advantages over the IntelliJ platform. Going through the internals it is easier setting up a FileTemplate and the reference rather than the lifting it to Meta. That is exactly due to how those Templates are registered and referenced in the internals. The actual registration looks a bit like this `FileTemplateManager.getInstance(project).addTemplate(name="MyFileTemplate", extension="kt")`, but the FileTemplates are written in `.tf` files, which allow users to intercept variables of the project and ide to craft the template in the resource folder. Thereby, there is little value rather than using the XML API `<internalFileTemplate name="MyADT"/>` and writting:
```
//resources/fileTemplates/internal/MyADT.kt.tf
#if (${PACKAGE_NAME} && ${PACKAGE_NAME} != "")package ${PACKAGE_NAME}

#end
#parse("File Header.java")
sealed class ${NAME} {
  // add your leaves
}
```
One possible improvement is to hijack the extension, which has all variables that exist in a `.tf` file like `PACKAGE_NAME` or `NAME` for improved ergonomics and autocomplete.


idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/intention/IntentionSyntax.kt
```diff
@@ -142,23 +138,6 @@ interface IntentionSyntax : IntentionUtilitySyntax {
...
-   * Bare with us we will improve upon this issue at a later point.
-   * @param category are used to order Intentions
-   * @see com.intellij.codeInsight.intention.IntentionManager.registerIntentionAndMetaData
-   * @sample org.jetbrains.kotlin.idea.intentions.ValToObjectIntention
-   */
-  fun IdeMetaPlugin.addIntentionWithMetaData(
-    category: String,
-    intention: IntentionAction
-  ): ExtensionPhase =
-    IntentionExtensionProvider.RegisterIntentionWithMetaData(intention, category)*/
```
jansorg: Should class `RegisterIntentionWithMetaData` be kept? It's still used in `IdeInternalRegistry`.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/intention/IntentionSyntax.kt
```diff
@@ -79,7 +75,7 @@ interface IntentionSyntax : IntentionUtilitySyntax {
     text: String = "",
     kClass: Class<K> = KtElement::class.java as Class<K>,
     isApplicableTo: (element: K, caretOffset: Int) -> Boolean = Noop.boolean2False,
-    applyTo: (element: K, editor: Editor?) -> Unit = Noop.effect2,
+    applyTo: KtPsiFactory.(element: K, editor: Editor) -> Unit = Noop.effect3,
```
jansorg: I don't fully understand the need for a `KtPsiFactory` as receiver here. 

This is always called with a new instance (via `it.project?.ktPsiFactory?` below) and it might be unused by the callers of this method `IntentionSyntax.ktIntention`.
Can't callers of `IntentionSyntax.ktIntention` just create their own instance when needed? 

If it's required and necessary to pass as a receiver, then you might want to cache the instance. For example, with a key attached to the project (which is a UserDataHolder). Something like this:
```
val KEY = Key.create<KtPsiFactory>("arrow.ktpsifactory")
KEY.set(project, KtPsiFactory(project))
// later
val ktPsiFactory = KEY.get(project)
```

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/intention/IntentionSyntax.kt
```diff
@@ -109,12 +105,12 @@ interface IntentionSyntax : IntentionUtilitySyntax {
...
     priority: PriorityAction.Priority = PriorityAction.Priority.LOW
   ): SelfTargetingIntention<K> =
     object : SelfTargetingIntention<K>(kClass, text), PriorityAction {
       override fun applyTo(element: K, editor: Editor?): Unit =
-        applyTo(element, editor)
+        editor?.let { it.project?.ktPsiFactory?.let { factory -> applyTo(factory, element, editor) } } ?: Unit
```
jansorg: see above

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/intention/IntentionSyntax.kt
i-walker: Thanks :)

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/intention/IntentionSyntax.kt
i-walker: They can, but `KtPsiFactory` has a lot of utilities users will end up using, when they're manipulating the intercepted KtElement.
IMO there is no need to cache the instance unless we have sever performance drawbacks.
And I could detect them, yet, as we haven't used a lot of Intentions or Inspections in our project.

jansorg: I agree about the caching.
I still don't entirely agree about the `KtPsiFactory`, though. There's a lot of other, potentially useful  classes which may be used by the users of this method. This may be my personal opinion about style of apis. In the end it's your call what's best here :)

i-walker: @jansorg Do you have suggestions which other APIs may be interesting for users? We can add them.
Nothing prevents the user to use other APIs or classes manually, but we can also provide them as top-level constructs/ services in each function so that users have direct access without researching for rather `common` undocumented pit-holes.

jansorg: I wanted to say that assumptions like this might be too smart for a generic api :)
This method assumes that the user will certainly need a `KtPsiFactory`. Some will need it, some won't. Some users will need a `com.intellij.psi.PsiElementFactory` for Java, some will need both factories, other will need something completely different.
(And a reference to `KtPsiFactory` requires a hard dependency on the Kotlin plugin.)

Personally, I'd probably add a note in the method's documentation to educate the user instead of always creating a new instance.

PR #391, opened 2020-01-06
QuoteTest uncommented but ignored, lambdaExpression tests improvement and added to QuoteTest

matiaslev:
#390

### DIFF
--- a/compiler-plugin/src/test/kotlin/arrow/meta/quotes/pbt/QuoteTest.kt
+++ b/compiler-plugin/src/test/kotlin/arrow/meta/quotes/pbt/QuoteTest.kt
@@ -1,49 +1,79 @@
 package arrow.meta.quotes.pbt
 
+import arrow.meta.plugin.testing.Code
+import arrow.meta.plugin.testing.CompilerTest
+import arrow.meta.plugin.testing.assertThis
+import arrow.meta.quotes.scope.templates.BlockExpressionTest.Companion.blockExpression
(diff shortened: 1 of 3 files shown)


compiler-plugin/src/test/kotlin/arrow/meta/quotes/pbt/QuoteTest.kt
```diff
@@ -1,49 +1,79 @@
...
+import arrow.meta.quotes.scope.templates.WhenEntryTest.Companion.whenEntryExpression
+import arrow.meta.quotes.scope.templates.WhenExpressionTest.Companion.whenExpressions
+import arrow.meta.quotes.scope.templates.WhileExpressionTest.Companion.whileExpression
+import org.junit.Ignore
 import org.junit.Test
 
 class QuoteTest {
 
   @Test
+  @Ignore
```
rachelcarmena: Thanks for this new contribution, @matiaslev!! :tada: Please, I think it would be preferable not to change this file until there is a solution for it. Maybe it's better to remove it. Otherwise, it can lead to confusion.  Thanks!

matiaslev: thanks!, I will revert the changes on this file.

rachelcarmena: Thanks to you!! :+1:

PR #405, opened 2020-01-11
Add refinements in prelude

i-walker:
Are these a good addition to the prelude? 
I am completely open to feedback, if there is a better way to define them.

### DIFF
--- a/prelude/src/main/kotlin/arrow/refinement.kt
+++ b/prelude/src/main/kotlin/arrow/refinement.kt
@@ -1,2 +1,50 @@
 package arrow
 
+import arrow.TypeProof.Extension
+import arrow.TypeProof.Refinement
+
+/**
+ * val y: PositiveInt = -3 // not ok
+ * val x: PositiveInt = 1 // ok
+ * val impossible: PositiveInt = PositiveInt(-5) // fails at compile time
+ * val result: PositiveInt = y.value?.let(Int::inc)
+ */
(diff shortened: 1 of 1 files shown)


prelude/src/main/kotlin/arrow/refinement.kt
```diff
@@ -1,2 +1,50 @@
...
+  this?.safeInt() ?: IntString(this)
+
+@Proof(Extension)
+fun IntString.string(): String? =
+  value
+
+/**
+ * ```kotlin
+ * val element: IntString = "44"
+ * val eleven: Int? = element.value?.div(4)
```
i-walker: I reckon this would be using an inductive proof from (String?) -> (Int?), thereby we may access `Int::div`. But I am not sure if that is the right way to define it.

prelude/src/main/kotlin/arrow/refinement.kt
```diff
@@ -1,2 +1,50 @@
 package arrow
 
+import arrow.TypeProof.Extension
+import arrow.TypeProof.Refinement
+
+/**
+ * val y: PositiveInt = -3 // not ok
+ * val x: PositiveInt = 1 // ok
+ * val impossible: PositiveInt = PositiveInt(-5) // fails at compile time
```
i-walker: Does this fail at compile time through static analysis of literals?

i-walker:
Closing this because of the old encoding

PR #440, opened 2020-02-12
we can use named functions as SyntheticElement

matiaslev:
if this makes sense, we should update the hello world example.


raulraja:
@matiaslev, thanks for this. We are going to close this as synthetic elements, in general, are supposed to be handled with the upcoming FIR API, which may supersede how we do quotes.

PR #442, opened 2020-02-12
Revert "Gradle Plugin: add compiler-plugin dependency from the plugin"

rachelcarmena:
Reverts arrow-kt/arrow-meta#422 because it doesn't work as expected.


rachelcarmena:
The reason of the build error: Kotlin team has removed an EAP version from the repository. I'm going to add another commit to update the patch used to check the compatibility with upcoming versions.

rachelcarmena:
Instead of using a specific version of EAP or DEV, the patch will have a generic version. For instance: `1.3.70-eap.diff` instead of `1.3.70-eap-22.diff`

PR #451, opened 2020-02-13
Remove `@sample` from KDoc for Hello World IDE plugin

cb372:
It was causing a [confusing doc page](https://meta.arrow-kt.io/docs/apidocs/idea-plugin/arrow.meta.ide.plugins.helloworld/hello-world.html) to be generated, with all the information shown twice.


cb372:
Build failed with:

```
Could not GET 'https://dl.bintray.com/kotlin/kotlin-dev/org/jetbrains/kotlin/kotlin-gradle-plugin/1.3.70-eap-222-14/kotlin-gradle-plugin-1.3.70-eap-222-14.pom'. Received status code 401 from server: Unauthorized
```

Not sure what that's about. Maybe just needs a retry?

rachelcarmena:
> https://dl.bintray.com/kotlin/kotlin-dev/org/jetbrains/kotlin/kotlin-gradle-plugin/1.3.70-eap-222-14/kotlin-gradle-plugin-1.3.70-eap-222-14.pom

Thanks @cb372 !! Right, you can try a re-run because that file exists. We had a similar error a few days ago because a version was removed. However, in this case, the version exists.

In any case, it has been useful for me because I realized that Kotlin has 2 EAP versions from today, so we'll need to update the checks for upcoming versions. 

Thanks again!!

cb372:
I don't think I have permission to re-run the build

PR #453, opened 2020-02-14
wip: Improve editor experience with synthetic resolution

jansorg:
This is a follow-up to https://github.com/arrow-kt/arrow-meta/pull/252
cc @raulraja 

Changes:
- cancel an ongoing transformation and cache update when the user continues to type in the editor
- don't trigger a transformation immediately after each typed character, do this after a small delay only for the latest change
- use non-blocking read actions instead of blocking read actions. IntelliJ follows a single-writer/many-readers model. PSI access (i.e. our transformation) has to happen in a read action. Typing triggers a write action. Therefore typing is delayed if there's an ongoing, long-running read action. non-blocking read actions are run in the background and cancelled when there's a pending write action (and then repeated).
- use a bulk-aware document listener for documents. A bulk update now only triggers one transformation  instead of many.
- skip a transformations, which were triggered by document updates. Code completion triggers document updates in LightVirtualFiles, i.e. in-memory dummy files just for the collection of completion items. We don't need to handle this. This puts less load on the editor / quote system.
- Use a MergingUpdaterQueue instead of our own pool thread. The queue is also limited to just one task at a time, but uses IntelliJ's application-wide background worker thead pool. 

Open:
This should be much better than before. But the current architecture makes this a bit difficult to handle well. At the moment, the transformation needs read access to the original psi file. Long-running transformations will definitely block the editor. But, transformations are fast at this time. But this may happen in the future or for very large source files.
A better solution would be to take the source text in a read action and then transform outside of a read action without access to the original PSI.
But I still need to validate that this is possible.


idea-plugin/src/main/kotlin/arrow/meta/ide/phases/resolve/QuoteSystemCache.kt
```diff
@@ -176,60 +213,72 @@ class QuoteSystemCache(private val project: Project) : ProjectComponent, Disposa
...
-        val start = System.currentTimeMillis()
-        try {
-          transformFiles(updatedFiles)
-        } catch (e: Exception) {
-          LOG.warn("error transforming files $updatedFiles. Falling back to empty transformation.", e)
-          emptyList()
-        } finally {
-          val duration = System.currentTimeMillis() - start
-          LOG.warn("kt file transformation: %d files, duration %d ms".format(updatedFiles.size, duration))
+    // non–blocking read action mode may execute multiple times until the action finished without being cancelled
```
raulraja: this is very useful info, thanks for all these comments 👏

jansorg:
@i-walker thank you for your review. I'll merge this now. I'll work on improvements to the editor and transformation handling  after your refactorings.

PR #462, opened 2020-02-21
Refactor Quote System 

### LINKED ISSUE #444: Remove project components
The following componts need to be refactored to their application-level or project-level representation.
- [x] arrow.meta.ide.MetaPluginRegistrarComponent
- [ ] arrow.meta.ide.phases.resolve.QuoteSystemCache

i-walker:
resolves #444 
- removes component implementation of the Quote-System
- separates the resolution lifecycle and the transformation mechanism into a project-level service: `QuoteSystemService` and a ProjectLifecycleListener

The advantage is that 3rd Parties can override the`QuoteSystemService`  extension to provide their own service implementation. Though this needs a few more fixes to promote such behavior - this would be handled in another PR.


idea-plugin/src/main/kotlin/arrow/meta/ide/phases/resolve/QuoteSystem.kt
```diff
@@ -0,0 +1,267 @@
...
+            .flatten()
+            .forEach { descriptor ->
+              try {
+                descriptor.safeAs<LazyEntity>()?.forceResolveAllContents()
+              } catch (e: IndexNotReadyException) {
+                LOG.warn("Index wasn't ready to resolve: ${descriptor.name}")
+              }
+            }
+
+          transformed.forEach { // refresh the highlighting of editors of modified files, using the new cache
```
jansorg: A new file which also modifies some of the extracted code is a bit hard to review. 
The old code had `metaCache.updateTransformations(sourceFile, transformedFile, synthDescriptors)` after the calls of `forceResolveAllContents()`. I don't see it anymore or am I missing something?

idea-plugin/src/test/kotlin/arrow/meta/ide/phases/resolve/QuoteSystemCacheTest.kt
```diff
@@ -14,9 +16,7 @@ class QuoteSystemCacheTest : LightPlatformCodeInsightFixture4TestCase() {
   fun testIncrementalCacheUpdate() {
     val file = myFixture.addFileToProject("testArrow/source.kt", "package testArrow")
 
-    // this is the initial rebuild and the initial cache population
-    val cache = QuoteSystemCache.getInstance(project)
-    cache.forceRebuild()
+    project.testQuoteSystem()?.forceRebuild(project)
```
jansorg: could non-null be enforced? There would be something quite wrong if the quote system isn't available in the test.

idea-plugin/src/main/kotlin/arrow/meta/ide/phases/resolve/MetaSyntheticResolveExtension.kt
```diff
@@ -60,7 +60,7 @@ open class MetaSyntheticResolveExtension(project: Project) : SyntheticResolveExt
     if (!thisDescriptor.isMetaSynthetic()) {
       thisDescriptor.findPsi().safeAs<KtClassOrObject>()?.containingKtFile?.packageFqName?.let { packageName ->
         result.replaceWithSynthetics(packageName, declarationProvider)
-        cache.resolved(packageName)?.filterIsInstance<LazyClassDescriptor>()?.find { it.fqNameSafe == thisDescriptor.fqNameSafe }?.let { classDescriptor ->
+        service?.cache?.descriptors(packageName)?.filterIsInstance<LazyClassDescriptor>()?.find { it.fqNameSafe == thisDescriptor.fqNameSafe }?.let { classDescriptor ->
```
jansorg: The internal data structurs of the cache were change. Could you add a note in this PR why this was done? Is this more efficient and/or simpler code?

idea-plugin/src/main/kotlin/arrow/meta/ide/phases/resolve/MetaSyntheticPackageFragmentProvider.kt
```diff
@@ -12,82 +13,69 @@ import org.jetbrains.kotlin.descriptors.PropertyDescriptor
...
+        // fixme always provide a value or only when a cached value exists?
+        return listOf(BuildCachePackageFragmentDescriptor(module, fqName, project))
+      }
 
-    override fun getSubPackagesOf(fqName: FqName, nameFilter: (Name) -> Boolean): Collection<FqName> {
-      //fixme optimize this
-      return cache.packageList().filter { it.parent() == fqName }
+      override fun getSubPackagesOf(fqName: FqName, nameFilter: (Name) -> Boolean): Collection<FqName> {
+        //fixme optimize this
+        return project.quoteCache()?.packages()?.filter { it.parent() == fqName } ?: emptyList()
```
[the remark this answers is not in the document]
i-walker: Yes I added that :)

idea-plugin/src/main/kotlin/arrow/meta/ide/phases/resolve/MetaSyntheticResolveExtension.kt
i-walker: Sure will do :)

idea-plugin/src/test/kotlin/arrow/meta/ide/phases/resolve/QuoteSystemCacheTest.kt
i-walker: If the service is not available the test will automatically fail. But we can throw a refined Error message without a stacktrace in that case.

idea-plugin/src/main/kotlin/arrow/meta/ide/phases/resolve/QuoteSystem.kt
i-walker: Yes the `update` is added in the next commit

idea-plugin/src/main/kotlin/arrow/meta/ide/phases/resolve/QuoteSystem.kt
```diff
@@ -0,0 +1,267 @@
...
+              }
+            }
+          }
+        }*/
+
+        // update the resolved data as soon as index access is possible
+        // fixme protect against multiple transformations at once
+        // fixme this might be delayed, make sure we're handling this correctly with a test
+        DumbService.getInstance(project).runReadActionInSmartMode {
+          if (strategy.resetCache) {
```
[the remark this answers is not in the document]
i-walker: The old code indicates https://github.com/arrow-kt/arrow-meta/blob/f97ba5643745548f6c2d327c3686d366173f6288/idea-plugin/src/main/kotlin/arrow/meta/ide/phases/resolve/QuoteSystemCache.kt#L181 that a valid refresh only happens on nonEmptyLists.

i-walker:
Thanks @jansorg and @raulraja for the review. I kept these changes as a blueprint for a retry branch, as there are more refactorings needed from the ground up.
The next iteration has a similar approach.

PR #464, opened 2020-02-27
Refactor QuoteSystem

### LINKED ISSUE #444: Remove project components

i-walker:
resolves #444 
Still WIP!

Design Goals for this PR:
- [X] explicit dependencies of functions to promote Services.
- [x] the test environment for Quotes is a wrapper over the ide implementation `TestQuoteSystemService`.
- [X] remove false positive in the test environment that allows resetting the Cache, which is not reflecting a "usual" user experience. -> We should create `anAction` for users, though.
- [X] remove QuoteCache corner cases, where removed files were still persisted.
- [x] Break down the QuoteSystem into one Service `QuoteService`, which encompasses all operations without containing any State and can be manipulated by 3rd parties.
- [x] `QuoteCache` persist its underlying data structure (currently a `Map<KtFile, QuoteInfo>`) -- QuoteInfo contains the minimal amount of data after a quote transformation, considering that quotes are able to transform package directories there is an intuitive mapping from `(KtFile) -> Pair<FqName, List<Descriptor>>`. We utilize this initial idea in the type `QuoteInfo` with an implicit KtFile. Furthermore, immutability is favored to utilize the morphisms provided by `QuoteService`. 
- [x] implement a persistent `QuoteCache` that caches instances after ide restart and can be transformed by Services.
- [x] generalize configurations for `refreshCash` to an interface `CacheStrategy` to prevent changing the signature in `refreshCash` for incremental dependency changes.
- [x] operations for the ide lifecycle are moved to a ProjectLifecycleListener - previously initComponent, projectClosed, etc.
- [x] transform changes in #469 to a persistent service, which stores the Atomic Boolean value.
 
With those changes there is no need for a component.


jansorg:
@i-walker I wasn't sure if I understood all of the plan, but just in case. 
If it's planned to persist the state of a QuoteCache on disk and to restore when a project is opened, then it should be done cautiously. Manually persisting state based on PSI and VirtualFile quickly gets out of sync, e.g. by external changes, projects sharing source files, etc. 
The IntelliJ API has a whole bunch of classes to handle these cases, but sadly these are hard to integrate with the quote transformations. IntelliJ has the concept of "Index" to manage transformations based on PSI, for example. (There's also a "CacheManager", but that's used to cache without persisting cached data.)
My thinking is that the open issues of editor and resolving should be solved first (I'm working on this now some more) and when that is all working, then persisting the quote cache can be tackled.

i-walker:
This PR has an architectural goal in mind rather than on changing implementation details and I agree that we should focus on resolution. 

Meta will allow plugin developers to manipulate endpoints over specific features that `Arrow` proposes in the IDE - such as Quotes, Purity, etc. 
That means there is an algebra that swaps the implementation of behavior defined by services, with user-defined primitives. 
And the swapped implementation is used throughout the rest of the ide lifecycle.
There is no need to mock other ide parts of a feature because there is a separation of behavior and data. 
Consistency is given by default, as the IDE ensures at program start that there is only one instance of a service. 
This allows everyone in the community to refine the IDE experience, based on their needs without relying on evasive methods or hacks.

This PR may serve as an example, how that can be implemented.

idea-plugin/src/main/kotlin/arrow/meta/ide/MetaRegistrar.kt
```diff
@@ -34,17 +45,43 @@ private val metaPlugin = IdeMetaPlugin()
...
+    },
+    afterProjectClosed = { project: Project ->
+      project.quoteConfigs()?.let { (quoteSystem, cache) ->
+        try {
+          quoteSystem.context.cacheExec.safeAs<BoundedTaskExecutor>()?.shutdownNow()
+        } catch (e: Exception) {
+          LOG.warn("error shutting down pool", e)
+        } finally {
+          cache.clear()
+          project.getService(QuoteHighlightingCache::class.java)?.map { // resets the highlighter
```
jansorg: Is this needed? The project is closed here

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/application/cache/PersistentService.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import com.intellij.openapi.components.PersistentStateComponent
+
+/**
+ * [PersistentService] caches a [value] and solely allows monomorphic updates.
+ * This Service is one way to replace global variables stored in components to their service representation,
+ * store state in either project-level or application-level, as an effort to migrate to Services.
+ * This interface shouldn't be used to store PsiElements or VirtualFiles.
+ */
+interface PersistentService<A> : PersistentStateComponent<Id<A>> {
```
jansorg: Review notes of https://github.com/arrow-kt/arrow-meta/pull/490#pullrequestreview-381077404 apply here

idea-plugin/src/main/kotlin/arrow/meta/ide/internal/registry/IdeInternalRegistry.kt
```diff
@@ -195,9 +194,9 @@ internal interface IdeInternalRegistry : InternalRegistry {
       }
     }
 
-  fun <E> registerExtensionProvider(phase: ExtensionProvider<E>, ideCtx: IdeContext = IdeContext): Unit =
+  fun <E> registerExtensionProvider(phase: ExtensionProvider<E>): Unit =
     when (phase) {
-      is ExtensionProvider.AddExtension -> phase.run { Extensions.getRootArea().getExtensionPoint(EP_NAME).registerExtension(impl, loadingOrder, ideCtx.dispose) }
+      is ExtensionProvider.AddExtension -> phase.run { Extensions.getRootArea().getExtensionPoint(EP_NAME).registerExtension(impl, loadingOrder, ApplicationManager.getApplication()) }
```
jansorg: Could you add a note why this is always the application? I don't see why at first glance.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/quotes/lifecycle/QuoteLifecycle.kt
```diff
@@ -0,0 +1,185 @@
...
+      LOG.info("Initializing quote system cache...")
+      val files: List<KtFile> = runReadAction {
+        project.quoteRelevantFiles().also {
+          LOG.info("collected ${it.size} quote relevant files for Project:${project.name}")
+        }
+      }
+      quoteSystem.refreshCache(cache, project, files, analysisIdeExtensions, cacheStrategy())
+    } finally {
+      /*try { // why
+      flushData()
```
jansorg: The purpose of this call was to wait until the pending cache updates are all done. Then the latch below is counted down to let the highlighter pass proceed. The highlighter is waiting until the latch is counted down to avoid highlighting with incomplete cache data.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/quotes/lifecycle/QuoteLifecycle.kt
```diff
@@ -0,0 +1,185 @@
...
+    } catch (e: Exception) {
+    }*/
+      project.getService(QuoteHighlightingCache::class.java)?.run {
+        map {
+          it.copy(initialized = AtomicBoolean(true))
```
jansorg: This is possibly a bug. This creates a new `AtomicBoolean` instead of updating the existing value with `true`. That's used to allow thread-safe updates of the value. 
Now it's a reassignment to the `PersistentState.value` variable via `map`. `value` isn't volatile, so this may break thread-safety.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/quotes/lifecycle/QuoteLifecycle.kt
```diff
@@ -0,0 +1,185 @@
...
+        }
+        state.extract().latch.countDown()
```
jansorg: This is a bit confusing, because this is modifying the state, which was copied by `map` above.

idea-plugin/src/main/kotlin/arrow/meta/ide/MetaRegistrar.kt
i-walker: Yes to reset the lifecycle of the Highlighter.

i-walker: I removed it

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/quotes/lifecycle/QuoteLifecycle.kt
i-walker: added it

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/quotes/lifecycle/QuoteLifecycle.kt
```diff
@@ -0,0 +1,189 @@
...
+    getService(QuoteCache::class.java)?.let { cache ->
+      QuoteConfigs(quoteSystem, cache)
+    }
+  } ?: throw UnavailableServices(listOf(QuoteSystemService::class.java, QuoteCache::class.java))
+
+internal fun quoteProjectOpened(project: Project): Unit = // previously in QuoteSystemCache.projectOpend()
+  project.quoteConfigs()?.let { (quoteSystem: QuoteSystemService, cache: QuoteCache) ->
+    // add a startup activity to populate the cache with a transformation of all project files
+    //StartupManager.getInstance(project).runWhenProjectIsInitialized {
+    //runBackgroundableTask("Arrow Meta", project, cancellable = false) {
```
jansorg, line 62: Removing the task here is a change of behaviour. 
`quoteProjectOpened` is executed on the EDT, so any long-running tasks shouldn't be performed in this thread. Otherwise the UI would block. 
It's executed on the EDT, because it's wrapped in a non-dumbaware startup activity, see the [docs of StartupActivity](https://github.com/JetBrains/intellij-community/blob/193/platform/core-api/src/com/intellij/openapi/startup/StartupActivity.java#L11) for the threading guidelines.

imo at least the use of the task should be re-activated here.

i-walker: made the Startup activity dumbAware in the last commit

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/quotes/cache/QuoteCacheService.kt
```diff
@@ -0,0 +1,48 @@
...
+
+import arrow.meta.ide.dsl.application.services.Id
+import arrow.meta.ide.dsl.application.services.IdService
+import org.jetbrains.kotlin.descriptors.DeclarationDescriptor
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.KtFile
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * a QuoteCache implementation as a IdService, which stores Quote transformations on disk.
```
jansorg, line 11: nit: this isn't storing on disk

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/quotes/cache/QuoteCacheService.kt
```diff
@@ -0,0 +1,48 @@
...
+    get() = value.extract().size
+
+  override fun packages(): List<FqName> =
+    value.extract().values.map { it.first }.distinct()
+
+  override fun descriptors(packageFqName: FqName): List<DeclarationDescriptor> =
+    value.extract().values.filter { it.first == packageFqName }.map { it.second }.flatten()
+
+  override fun removeQuotedFile(file: KtFile): QuoteInfo? =
+    value.extract().remove(file).also { _ ->
```
jansorg, line 31: isn't this calling `remove(file)` twice on the same instance of the ConcurrentHashMap? `map` returns a copy of the `Id`, but both reference the same instance of the HashMap.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/utils/IdeUtils.kt
```diff
@@ -101,6 +110,33 @@ fun <F : CallableDescriptor> F.returns(ktFunction: KtNamedFunction, types: Kotli
 inline fun <reified K : PsiElement> K.replaceK(to: K): K? =
   replace(to).safeAs()
 
+// fixme use ViewProvider's files instead?
+@Suppress("UNCHECKED_CAST")
+fun <F : PsiFile> List<VirtualFile>.files(project: Project): List<F> =
```
jansorg, line 115: Isn't this returning all PsiFiles, regardless of the type of `F`? 
`F` isn't reified and the concrete type of `F` is erased at runtime, afaik.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/quotes/resolve/QuoteHighlightingPassFactory.kt
```diff
@@ -0,0 +1,58 @@
...
+   * is finished. This is necessary to implement fully working highlighting of .kt files, which
+   * access data from the Quote transformations during resolving.
+   */
+  fun waitToInitialize(): Unit =
+    value.extract().let { cache ->
+      if (!cache.initialized) { // this is not executed anymore
+        cache.latch.await(5, TimeUnit.SECONDS)
+        println("BOOOOOOMMMMMMM")
+      }
+      println("NOOO BOOOOM")
```
jansorg, line 56: nit: put this into an `else` branch? Currently this is always printed

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/quotes/cache/QuoteCacheService.kt
```diff
@@ -0,0 +1,48 @@
...
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * a QuoteCache implementation as a IdService, which stores Quote transformations on disk.
+ * This is an initial implementation, as it does not populate the index of other connected projects,
+ * and the cache stored on disk get out of sync with external changes to the disk, projects sharing source files or other situations.
+ * fixme cache both per module? modules may define different ktFiles for the same package fqName
+ * TODO: register this via Meta.
+ */
+private object QuoteCacheService : IdService<ConcurrentHashMap<KtFile, QuoteInfo>>, QuoteCache {
```
jansorg, line 17: It's an object, but it's registered as a project service. 
There's one service instance per project, so it should be a `class`, not an `object`.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/quotes/resolve/QuoteSyntheticPackageFragmentProvider.kt
```diff
@@ -19,53 +20,51 @@ import org.jetbrains.kotlin.resolve.scopes.DescriptorKindFilter
...
-val IdeMetaPlugin.metaSyntheticPackageFragmentProvider: ExtensionPhase
+val IdeMetaPlugin.quoteSyntheticPackageFragmentProvider: ExtensionPhase
   get() = packageFragmentProvider { project, module, _, _, _, _ ->
     descriptorCachePackageFragmentProvider(module, project)
   }
 
 private val descriptorCachePackageFragmentProvider: (ModuleDescriptor, Project) -> PackageFragmentProvider
   get() = { module, project ->
-    val cache = QuoteSystemCache.getInstance(project)
+    val quoteCache: QuoteCache? = project.getService(QuoteCache::class.java)
```
jansorg, line 30: I think that accepting an unavailable service here may introduce bugs, which are hard to spot. If `null`, then resolving may silently fail. As far as I can see there should always be a `QuoteCache` in the project.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/quotes/cache/QuoteCacheService.kt
i-walker: Nop, the first remove is a pure computation that won't affect the underlying data structure. the second is impure in the sense that it uses the impure map function from `Id<A>` transforms the data structure here ConcurrentHashMap and `removeQuotedFile` results with the output of the pure computation above. 

It's ugly because morphisms defined used in `Id<A>` need to be pure and `map` returns the new data structure, whereas in this case for  `removeQuotedFile` is interested in the new mapping to other descriptors. 
I am totally open to change the algebra in `QuoteCache`, but we should open a new Issue for that.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/utils/IdeUtils.kt
i-walker: Nop this will only filter files which have the correct Type `F`.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/quotes/resolve/QuoteSyntheticPackageFragmentProvider.kt
i-walker: Yes, before it was ill typed. And that silent fail was implied. This will silently fail, and let other SyntheticPackageFragmentProvider proceed

i-walker:
Thanks for the review. I am ready for merging and addressing upcoming issues in further patches in `type-proofs` and master as we go. 
@jansorg 🎉 my ide hero.

PR #469, opened 2020-03-06
Force flush of Quote transformations before highlighting

jansorg:
@raulraja This is finally some progress with this unnerving issue of broken highlighting.
The reason was, that resolving was triggered very early during highlighting (i.e. by the folding outline builder). At this point the initial transformation wasn't finished yet. Therefore the highlighting working with missing or incomplete Quote transformation datat. This resulted in broken highlighting.

Although we forced a restart of highlighting, this did not seem to be enough. I did not dig into the highlighting engine to find out why.

This PR waits for the initial highlighting in a .kt file until the initial Quote transformations, which are registered when the project is opened, are finished.

Questions:
- This approach is a bit brutal as it delays highlighting. But I don't know of another solution at this time. There might be one, but it'll probably be time-consuming to debug and find. My suggestion is to continue with this, work on other important issues of the editor integration, and then get back to this. Does this sound okay to you? Or do you know an extension point of the Kotlin plugin which might be used here instead? This needs to be used very early in the resolving.
- Highlighting worked even with `MetaSyntheticResolveExtension` disabled. Do you know why this was added? I noticed that it's replacing classes with synthetic classes, but the data of the `PackageFragmentProvider` already seems to be used here.

### DIFF
--- a/idea-plugin/src/main/resources/META-INF/plugin.xml
+++ b/idea-plugin/src/main/resources/META-INF/plugin.xml
@@ -30,7 +30,7 @@
     </extensions>
 
     <extensions defaultExtensionNs="org.jetbrains.kotlin">
-        <!-- syntheticResolveExtension implementation="arrow.meta.ide.phases.resolve.MetaSyntheticResolveExtension" /-->
+        <syntheticResolveExtension implementation="arrow.meta.ide.phases.resolve.MetaSyntheticResolveExtension" />
     </extensions>
 
 </idea-plugin>
--- a/idea-plugin/src/main/resources/META-INF/plugin.xml
+++ b/idea-plugin/src/main/resources/META-INF/plugin.xml
@@ -25,11 +25,12 @@
     </project-components>
 
     <extensions defaultExtensionNs="com.intellij">
+        <highlightingPassFactory implementation="arrow.meta.ide.ArrowHighlightingPassFactory"/>
         <applicationInitializedListener implementation="arrow.meta.ide.MetaRegistrar"/>
     </extensions>
 
     <extensions defaultExtensionNs="org.jetbrains.kotlin">
-        <syntheticResolveExtension implementation="arrow.meta.ide.phases.resolve.MetaSyntheticResolveExtension" />
+        <!-- syntheticResolveExtension implementation="arrow.meta.ide.phases.resolve.MetaSyntheticResolveExtension" /-->
     </extensions>
 
 </idea-plugin>
--- a/idea-plugin/src/main/kotlin/arrow/meta/ide/ArrowHighlightingPassFactory.kt
+++ b/idea-plugin/src/main/kotlin/arrow/meta/ide/ArrowHighlightingPassFactory.kt
@@ -1,0 +1,21 @@
+package arrow.meta.ide
+
+import arrow.meta.ide.phases.resolve.QuoteSystemCache
+import com.intellij.codeHighlighting.Pass
+import com.intellij.codeHighlighting.TextEditorHighlightingPassFactoryRegistrar
+import com.intellij.codeHighlighting.TextEditorHighlightingPassRegistrar
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiFile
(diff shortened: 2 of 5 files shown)


raulraja:
This approach looks good to me. Maybe Imran knows about an earlier phase in which this can be hooked up. Thanks and feel free to merge when ready

i-walker:
May you define what intended behavior your trying to achieve - from a lifecycle perspective?
That why its easier to locate the right extension.

IIRC project lifecycle has these extensions - which are executed before `ProjectComponent.initComponent`:
- ProjectLifecycleListener.beforeProjectOpend: `(Project) -> Unit`
- StartupActivities.runActivity: `(Project) -> Unit` // order: DumbAware, Background and the last is the regular one.
- ProjectManager.TOPIC

There is also the chance that the quote system should target EventListeners instead of the project lifecycle.

idea-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -25,11 +25,12 @@
     </project-components>
 
     <extensions defaultExtensionNs="com.intellij">
+        <highlightingPassFactory implementation="arrow.meta.ide.ArrowHighlightingPassFactory"/>
         <applicationInitializedListener implementation="arrow.meta.ide.MetaRegistrar"/>
     </extensions>
 
     <extensions defaultExtensionNs="org.jetbrains.kotlin">
-        <syntheticResolveExtension implementation="arrow.meta.ide.phases.resolve.MetaSyntheticResolveExtension" />
+        <!-- syntheticResolveExtension implementation="arrow.meta.ide.phases.resolve.MetaSyntheticResolveExtension" /-->
```
i-walker: `SyntheticResolveExtension` and `PackageFragmentProvider` come from the compiler API. We could move cli specific parts of the quote system to the compiler to prevent confusion, but for now, we should keep them registered for consistency reasons. 
```suggestion
        <syntheticResolveExtension implementation="arrow.meta.ide.phases.resolve.MetaSyntheticResolveExtension" />
```

jansorg: Hm, I don't fully understand the note about the compiler API. Isn't this just enabled for the IDE plugin via plugin.xml? 
I couldn't find out what it does for the IDE and the resolving, at least I couldn't find a feature which was broken after I disabled this.
If you want me to, I could reenable this, of course.

jansorg:
> May you define what intended behavior your trying to achieve - from a lifecycle perspective?
> That why its easier to locate the right extension.
> 
> IIRC project lifecycle has these extensions - which are executed before `ProjectComponent.initComponent`:
> 
> * ProjectLifecycleListener.beforeProjectOpend: `(Project) -> Unit`
> * StartupActivities.runActivity: `(Project) -> Unit` // order: DumbAware, Background and the last is the regular one.
> * ProjectManager.TOPIC
> 
> There is also the chance that the quote system should target EventListeners instead of the project lifecycle.

This is about the initial highlighting of open files at startup.
The quote system already uses modification event listeners to update the cache. But there's a full transformation of all .kt files at startup to initialize the quote system.

A background task is started after the project is opened to create the initial cache data. This happens after project open because the task is accessing indexes.

The highlighting of files at startup seems to start before the background tasks etc are finished. 

This means that highlighting and initial cache update are overlapping. Highlighting is broken due to this incomplete cache update.
This PR now waits that the background task finishes before highlighting of .kt files starts.

I meant if there's an extension of the Kotlin plugin, which is called at the very beginning of the .kt resolving. This might be an alternative to hooking into the highlighting engine to force a flush.

idea-plugin/src/main/resources/META-INF/plugin.xml
i-walker: Yes, please remove your change here. 
We should investigate this behavior in another dedicated PR.

jansorg: Okay. Fixed.

jansorg:
@i-walker any objections that I merge this into your branch is-qs-new  ?

i-walker:
No lets merge it :)

PR #474, opened 2020-03-12
Resolves ide in type-proofs

i-walker:


rachelcarmena:
Hi @i-walker ! please :pray: where is the fix in this PR? There are several commits from master branch and I don't find the fix. Thanks in advance!

i-walker:
The fix was merging master into the type-proofs branch, resolving members, references and ide related changes that we updated in master.

rachelcarmena:
> The fix was merging master into the type-proofs branch, resolving members, references and ide related changes that we updated in master.

Thanks @i-walker !! Please, it would be useful if we don't use PR to merge branches. We'll have to repeat the same again. Thanks!

Issue #465, opened 2020-03-01
Add ResolveScopeEnlarger extension


i-walker:
- [ ] add a directory in `arrow.meta.ide.dsl` called `resolve`
- [ ] create an interface `ResolveScopeSyntax`, which creates a meta dsl for `com.intellij.psi.ResolveScopeEnlarger`
- [ ] create an minimal example and Docs on what an `ResolveScopeEnlarger` does. Please add those in a Kdoc.

Note:
This [link]( https://meta.arrow-kt.io/docs/apidocs/idea-plugin/arrow.meta.ide.dsl.extensions/-extension-provider-syntax/extension-provider.html) is helpful to understand how the meta dsl in the ide is constructed.
Check out these instances `org.jetbrains.android.AndroidResolveScopeEnlarger`, `org.jetbrains.kotlin.idea.CommonModuleResolveScopeEnlarger` in the [kotlin compiler](https://github.com/JetBrains/kotlin/tree/1.3.70)

matiaslev:
please let me look at it

i-walker:
Resolved with #479 :) Thanks @matiaslev

PR #479, opened 2020-03-16
WIP: ResolveScopeEnlarger ide dsl

matiaslev:
#465 
please let me know if these make sense and if something more needs to be added in the scope of this ticket.


idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/resolve/ResolveScopeSyntax.kt
```diff
@@ -0,0 +1,23 @@
...
+import arrow.meta.phases.ExtensionPhase
+import com.intellij.openapi.extensions.LoadingOrder
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.psi.ResolveScopeEnlarger
+import com.intellij.psi.search.SearchScope
+
+interface ResolveScopeSyntax {
+    fun IdeMetaPlugin.resolveScope(
+        getAdditionalResolveScope: (file: VirtualFile, project: Project) -> SearchScope
```
i-walker:
```suggestion
        additionalResolveScope: (file: VirtualFile, project: Project?) -> SearchScope?
```
Nit's in the type signature

i-walker:
Yes, that's a good angle 👍 
Now the algebra needs extension functions like `ResolveScopeSyntax.searchScope(...): SearchScope` and other smaller functions, which are then composed together in one function like `IdeMetaPlugin.addResolveScopeEnlarger(...): ExtensionPhase`.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/resolve/ResolveScopeSyntax.kt
matiaslev: thanks!

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/resolve/ResolveScopeSyntax.kt
```diff
@@ -0,0 +1,23 @@
...
+    ): ExtensionPhase =
+      extensionProvider(
+        ResolveScopeEnlarger.EP_NAME,
+        object : ResolveScopeEnlarger() {
+            override fun getAdditionalResolveScope(file: VirtualFile, project: Project?): SearchScope? =
+                getAdditionalResolveScope(file, project)
+        },
+        LoadingOrder.FIRST
```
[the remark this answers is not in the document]
matiaslev: thanks! I don't understand the implications of the different LoadingOrder so far haha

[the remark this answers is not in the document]
i-walker: No worries, for the standard behavior of this extension it should not matter.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/resolve/ResolveScopeSyntax.kt
i-walker: I updated the suggestion from above. But that means to change the signature in the overridden function, too.

idea-plugin/src/main/kotlin/arrow/meta/ide/phases/resolve/ResolveScopeProvider.kt
```diff
@@ -0,0 +1,10 @@
+package arrow.meta.ide.phases.resolve
+
+import arrow.meta.phases.ExtensionPhase
+import com.intellij.psi.search.SearchScope
+
+sealed class ResolveScopeProvider : ExtensionPhase {
```
i-walker, line 6: That file is not needed

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/resolve/ResolveScopeSyntax.kt
```diff
@@ -0,0 +1,40 @@
...
+    ): ExtensionPhase =
+      extensionProvider(
+        ResolveScopeEnlarger.EP_NAME,
+        object : ResolveScopeEnlarger() {
+            override fun getAdditionalResolveScope(file: VirtualFile, project: Project): SearchScope =
+              additionalResolveScope(file, project)
+        }
+      )
+
+    fun IdeMetaPlugin.addResolveScopeEnlarger(
```
i-walker, line 23: This function is not needed

i-walker:
Looking good :)

In retrospect, there is no need for a function `addResolveScopeEnlarger`, that is built through composition. 
You may address the comments and rename `resolveScope` to `addResolveScopeEnlarger`.

i-walker:
🎉 Lets merge this :) Thanks @matiaslev

PR #484, opened 2020-03-18
Downgrade kotlin version to 1.3.61

danimontoya:
We need to temporarily revert all to 1.3.61
The extension point is broken in 1.3.7 because it calls the plugins before resolving the symbols. We have to go back to 1.3.61 until fixed so we can continue working


danimontoya:
Closing this in favor of https://github.com/arrow-kt/arrow-meta/pull/485

PR #487, opened 2020-03-18
Fix INTELLIJ_IDEA_VERSION according to the table on the site

danimontoya:
Fix INTELLIJ_IDEA_VERSION according to the table on the site


gradle.properties
```diff
@@ -6,7 +6,7 @@ ARROW_VERSION=0.10.4-SNAPSHOT
 ANK_GRADLE_PLUGIN_VERSION=0.7.1
 KOTLIN_VERSION=1.3.61
 KOTLIN_IDEA_VERSION=1.3.61-release-IJ2019.3-1
-INTELLIJ_IDEA_VERSION=2019.3.4
+INTELLIJ_IDEA_VERSION=2019.3
```
i-walker: AFAIK the IntelliJ version did not affect any changes in Meta - regarding the IR issues. 
And If a revert is needed the actual version is the following:
```suggestion
INTELLIJ_IDEA_VERSION=2019.3.3
```

danimontoya: Indeed, this change isn't for fixing any IR related issues. But one of the steps in master is failing and we thought this could fix it. We can try `2019.3.3`

PR #490, opened 2020-03-19
Simple Caching

i-walker:
This PR adds a Service, which stores data on disk and allows monomorphic updates. 
One example may look like this:
```kotlin
/** 
 * The following annotation implies that the storage location 
 * for the cache will be either outside of a project directory 
 * or outside of the application configs.
**/
@State(name = "Mylist", storages = [Storage(StoragePathMacros.CACHE_FILE)]) 
class MyList : Persistent<List<Int>> { // stores a List<Int> on disk
    @Volatile
    override var value: Id<List<Int>> = Id.just(emptyList())
}
```
in `plugin.xml`
```xml
 <extensions defaultExtensionNs="com.intellij">
   <projectService serviceImplementation="MyList"/>
</extensions>
```
Because it is registered as a project-level service, we may access it like this:
```kotlin
{ project: Project ->
  val service: MyList? = project.getService(MyList::class.java)
  service?.state // []
  service?.map { listOf(4, 3) } // [4, 3]
 //** some more code**//
  service?.map { it.map { i -> i * i } } // [16, 9]
}// invoking this function again starts with the stored state // [16, 9]
```

I'll be adding Docs, once the algebra for this is enabled in Meta and after #464 is merged.
Thanks, @jansorg for helping me with this!


i-walker:
Yes I have it already working with the quote cache https://github.com/arrow-kt/arrow-meta/pull/464/commits/14445d2e75c13452eb0a373fd1961ae09fa4e5f4.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/application/cache/PersistentService.kt
```diff
@@ -0,0 +1,43 @@
...
+   */
+  fun map(f: (A) -> A): Id<A> =
+    value.map(f)
+      .also { loadState(it) }
+}
+
+/**
+ * minimal port from arrow-core
+ */
+data class Id<A> internal constructor(private val value: A) {
```
jansorg: I took a closer look at the serialization.
IntelliJ serializes the state of a `PersistentStateComponent` into XML. When the state is restored (e.g. at startup), then the xml is converted into the state and applied via `loadState()`.

I don't think that this approach is compatible yet with IntelliJ's requirements.

IntelliJ is able to serialize and deserialize a value of type `Id`. The problems I'm seeing are:
- `Id` has an immutable property `value`. Deserialization (xml -> state) needs to apply a new value, so it needs a `var` or a pair of `get`, `set` methods
- By default, the attribute is serializes as `<Object/>`, this needs a `@get:Attribute("id_value")` in the Id class.
- The attribute annotations restricts the value of `Id` to primitive types, e.g. `Int`, `String`, etc. This is because IntelliJ serializes collections in tags and this isn't compatible with an attribute value.
- This is how JetBrains is using state in a Kotlin class: [source](com.intellij.diff.tools.external.ExternalDiffSettings)

Here's a simple test, which uses the same method `serializeStateInto` which is used by IntelliJ to serialize a persistent component into XML:
```kotlin
package arrow.meta.ide.dsl.application.cache

import com.intellij.configurationStore.serializeStateInto
import com.intellij.testFramework.fixtures.LightPlatformCodeInsightFixture4TestCase
import org.jdom.Element
import org.jdom.output.Format
import org.jdom.output.XMLOutputter
import org.junit.Test

class TestPersistentService : PersistentService<String> {
  override var value: Id<String> = Id.just("foo")
}

class TestMapPersistentService : PersistentService<List<String>> {
  override var value: Id<List<String>> = Id.just(listOf("foo", "bar"))
}

class PersistenceServiceTest : LightPlatformCodeInsightFixture4TestCase() {
  @Test
  fun testSerialization() {
    val settings = TestPersistentService()

    val serializedSettings = Element("settings")
    serializeStateInto(settings, serializedSettings)
    val xmlString = XMLOutputter(Format.getCompactFormat()).outputString(serializedSettings)

    assertEquals("<settings id_value=\"foo\" />", xmlString)
  }
}
```

This definition of `Id` is serializable for simple properties (`@Attribute` annotation and a `var`):
```kotlin
data class Id<A> internal constructor(@get:Attribute("id_value") var value: A) {
  fun <R> map(f: (A) -> R): Id<R> =
    Id(f(value))

  fun extract(): A = value

  companion object {
    fun <A> just(value: A): Id<A> =
      Id(value)
  }
}
```

### Alternative
I don't think that an immutable state is possible due to IntelliJ requirements.
IntelliJ has a preferrence for `BaseState`, which also adds helpers to define the properties. Returning an `Id` from state adds an additional xml tag to the output and complicates the use of non-primitive properties in the state.

Using something like this is better suitable for IntelliJ, but might not be what you'd like to use:
```kotlin
interface PersistentService<A : BaseState> : PersistentStateComponent<A> {
  var value: Id<A>

  override fun getState(): A =
    value.extract()

  override fun loadState(state: A) {
    value = Id.just(state)
  }

  // removed map for shortness
}

```

i-walker:
There is no need to store data on disk for our use-cases. 
Thanks, @jansorg for explaining the whole thing.

PR #491, opened 2020-03-19
Conf: recover branch number and Intellij IDEA version

rachelcarmena:
## Recover branch number

It's necessary to set the compatibility Intellij IDEA versions.

It has been added into build check in order to detect the problems previously.

## Recover Intellij IDEA version

It's not the reason of the last issue when publishing artifacts.


i-walker:
Thanks, @rachelcarmena !! 🎉

Issue #499, opened 2020-03-23
Stabilize IDE in `sv-test-proofs` branch


raulraja:
Currently running the IDE exhibits several exceptions and we also have issues with dependency management. (@rachelcarmena )

Once that is stable this is a list of visual features we want to add to the type proofs plugin in the IDE (@danimontoya, @aballano, @i-walker)

# Type Proofs
- A Visual chart that displays proofs connection between proofed types using nomnoml or diagram plugin 
- Dialog over proof icons that display the effect the proof has with examples
- Include implicit call sites in the usages of proofs when using Find Usages
## Type Classes
- Show an extension is connecting the behavior with the datatype with a type class icon.
- A visual clue for a call site where a type class gets injected
- An intention to make the implicit injection explicit
- An intention to make the explicit injection implicit
## Coercion functions
- A visual representation when a type coercion occurs by proof
- An intention to make the implicit coercion explicit
- An intention to make the explicit coercion implicit
## Union Types
- Simplify display of `Union2<A, B>` to `A | B` when displayed in the editor and back to edit mode when focused
## Tuple Types
- Simplify display of `Tuple2<A, B>` to `(A, B)` when displayed in the editor and back to edit mode when focused
## Refined Types
- A visual representation for refined classes + dialog.
- A visual representation for safe/unsafe usage of call sites that abide to the refinement.
- A visual representation for each one of the refined predicates that showcases they are compile time verified.

rachelcarmena:
> Currently running the IDE exhibits several exceptions and we also have issues with dependency management. (@rachelcarmena )

I checked to add just `compiler-plugin` dependency with different configurations to get transitive dependencies automatically and to avoid the hard-coded dependencies list in Gradle Plugin, however they aren't added. 

I'll get the transitive dependencies list programmatically. Take into account : `compileOnly` dependencies are not transitive dependencies just `compile` or `api` so `org.jetbrains.kotlin:kotlin-compiler-embeddable:$KOTLIN_VERSION"` (`compileOnly`) is not included in the transitive dependencies list.

danimontoya:
Shall we close this issue? `sv-test-proofs` got merged and we've created #543

rachelcarmena:
> I checked to add just `compiler-plugin` dependency with different configurations to get transitive dependencies automatically and to avoid the hard-coded dependencies list in Gradle Plugin, however they aren't added.
> 
[...]

PR #501, opened 2020-03-26
Draft to check for Coercion functions

danimontoya:
Draft to address part of #499 .


idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/ProofsIdePlugin.kt
```diff
@@ -51,11 +52,54 @@ val IdeMetaPlugin.typeProofsIde: Plugin
...
+                  })
+                  compilerContext.module = module
+                  // val proofs = compilerContext.coerceProofs(ktCall.returnType!!, ktCall.calleeExpression?.getType(ctx)!!)
+                  val proofs = compilerContext.coerceProofs(resolvedCall.getReturnType(), resolvedCall.candidateDescriptor.returnType!!)
+                  isSubtypeOf && proofs.isNotEmpty()
+                }
+              }
+          }
+        }),
+      addIntention(
```
i-walker: There is no need to use `addIntention`, this is just duplicating boilerplate.
- Create an Annotation within the `Annotator` for your desired Element
```
holder.createInfoAnnotation(call, "Type proof coerced call-site")?.let { info ->
    info.registerUniversalFix(<MyInspection/ or MyIntention>)
```
- define what the Intention should do, here you can use functions from `InspectionSyntax` or `IntentionSyntax`

danimontoya: Mmm for this specific use-case.. we have decided there is no need to highlight anything.. until you are with the cursor over it and only then provide the option to do the _fix_.. meaning make it implicit / explicit.. thats why we wanna use the `addIntention`
Still work in progress tho

i-walker: That is what an InfoAnnotation does

i-walker: But whatever works better for this use-case :)

compiler-plugin/src/main/kotlin/arrow/meta/plugins/proofs/phases/resolve/ProofResolution.kt
```diff
@@ -148,7 +148,9 @@ fun FunctionDescriptor.asProof(): Proof? =
   extensionReceiverParameter?.type?.let { from ->
     returnType?.let { to ->
       val annotationArgs = annotations.first().allValueArguments
-      val value: ConstantValue<*>? = annotationArgs[Name.identifier("of")]
+      val value: ConstantValue<*> = annotationArgs.getValue(Name.identifier("of"))
```
JorgeCastilloPrz: I'm totally agnostic on this, so please let me know if this doesn't make sense, but. Are we sure this value argument for the annotation is always gonna be there? we can call this function for any function descriptor, not just for type proofs ones.

compiler-plugin/src/main/kotlin/arrow/meta/plugins/proofs/phases/resolve/ProofResolution.kt
```diff
@@ -148,7 +148,9 @@ fun FunctionDescriptor.asProof(): Proof? =
...
+      val coerce: ConstantValue<Boolean> = annotationArgs.getValue(Name.identifier("coerce")) as ConstantValue<Boolean>
```
JorgeCastilloPrz: Same here, we can't know coerce is gonna be there, can we?

compiler-plugin/src/main/kotlin/arrow/meta/plugins/proofs/phases/resolve/ProofTypeChecker.kt
```diff
@@ -41,11 +41,11 @@ private const val logTypeSize: Int = 50
 
 class ProofTypeChecker(private val compilerContext: CompilerContext) : NewKotlinTypeChecker {
 
-  override fun isSubtypeOf(p0: KotlinType, p1: KotlinType): Boolean {
-    val result = baseLineTypeChecker.isSubtypeOf(p0, p1)
-    return if (!result && !p0.isError && !p1.isError) {
-      Log.Verbose({ "ProofTypeChecker.isSubtypeOf: ${p0.toString().take(logTypeSize)} : ${p1.toString().take(logTypeSize)} -> $this" }) {
-        compilerContext.extensionProof(p0, p1) != null
+  override fun isSubtypeOf(subtype: KotlinType, supertype: KotlinType): Boolean {
```
JorgeCastilloPrz: 👍 this rename helped me to understand what's happening here. 

Maybe out of scope, but: Isn't this more like a global function we should have available in the library for all `KotlinType`s maybe as an extension? or the implementation of "subtype of" is different now depending on the type checker?

gradle.properties
```diff
@@ -19,7 +19,7 @@ BINTRAY_VERSION=1.8.4
 GRADLE_VERSION=5.6.4
 OPENAPI_VERSION=7.0.3
 GRADLE_PLUGIN_PUBLISH_VERSION=0.10.1
-INTELLIJ_GRADLE_PLUGIN_VERSION=0.4.16
+INTELLIJ_GRADLE_PLUGIN_VERSION=0.4.17
```
JorgeCastilloPrz: How are we working towards this in meta? do we always target the latest stable intellij gradle plugin?

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/coercion.kt
```diff
@@ -0,0 +1,80 @@
...
+    kClass = KtElement::class.java,
+    isApplicableTo = { ktCall: KtElement, caretOffset: Int ->
+      when (ktCall) {
+        is KtDotQualifiedExpression -> ktCall.isDotQualifiedExpressionCoerced()
+        is KtProperty -> ktCall.isCoerced()
+        else -> false
+      }
+    },
+    applyTo = { ktCall, editor ->
+      // replace with proof previously found
```
JorgeCastilloPrz: This is where we'd provide the quickfix?

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/coercion.kt
```diff
@@ -0,0 +1,80 @@
...
+private fun KtDotQualifiedExpression.isDotQualifiedExpressionCoerced(): Boolean {
+  val bindingContext: BindingContext = analyze(bodyResolveMode = BodyResolveMode.FULL)
+  val calls = bindingContext.getSliceContents(BindingContext.RESOLVED_CALL)
+  return calls.filter { (call, resolvedCall) -> resolvedCall != null }
+    .any { (call, resolvedCall) ->
+      selectorExpression?.let { selectorExpression ->
+        val (type, superType) = receiverExpression.resolveType() to selectorExpression.resolveType()
+        val isSubtypeOf = baseLineTypeChecker.isSubtypeOf(type, superType)
+        val compilerContext = CompilerContext(project = call.callElement.project, eval = {
+          KotlinJsr223StandardScriptEngineFactory4Idea().scriptEngine.eval(it)
```
JorgeCastilloPrz: How does this work? So you create some sort of default script engine for Idea that you can use to evaluate the expression to get what exactly?

compiler-plugin/src/main/kotlin/arrow/meta/plugins/proofs/phases/resolve/ProofResolution.kt
aballano: You're actually right 👍 Dani just changed that now :)

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/coercion.kt
aballano: Exactly, although the name of the function is not straightforward to grasp IMO

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/coercion.kt
aballano: I'll redirect that question to the masters @raulraja @i-walker  :)

compiler-plugin/src/main/kotlin/arrow/meta/plugins/proofs/phases/resolve/ProofResolution.kt
danimontoya: Yeah, actually I was seeing errors in the logs because of that..

compiler-plugin/src/main/kotlin/arrow/meta/plugins/proofs/phases/resolve/ProofResolution.kt
danimontoya: Yeah, actually I was seeing errors in the logs because of that..

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/coercion.kt
danimontoya: Yeah! in `isApplicableTo` we look for the KtElements that we wanna offer the options to change and in `applyTo` the fixes that can be applied 👍

compiler-plugin/src/main/kotlin/arrow/meta/plugins/proofs/phases/resolve/ProofTypeChecker.kt
raulraja: The typechecker is instantiated by the compiler multiple times so we piggy back in every instance they create to replace it for this typechecker that gives precedence to the baseline type checker which is the original one the compiler provides in different parts.

gradle.properties
raulraja: Raquel can answer that better but the CI bot has a compatibility matrix

compiler-plugin/src/main/kotlin/arrow/meta/plugins/proofs/phases/resolve/ProofTypeChecker.kt
JorgeCastilloPrz: But the subtype relationship varies? or is this code that could be delegated to a shared ext fun?

gradle.properties
rachelcarmena: Hi! :raised_hands: 

CI bot considers:
* Intellij IDEA version
* Kotlin IDEA Plugin version
* Kotlin Compiler version

However. Intellij IDEA **Gradle Plugin** version consideration is missing. 

I added a task to include it after reviewing this PR: https://github.com/arrow-kt/arrow-meta/pull/498

compiler-plugin/src/main/kotlin/arrow/meta/plugins/proofs/phases/resolve/ProofTypeChecker.kt
raulraja: yes, you can delegate the body of the type checker to an ext fun but it's not necessary because that is already happening when you call `subtype.isSubtypeOf(superType)` that call that is already part of the compiler uses our injected KotlinTypeChecker so it's unnecessary wrapping IMO

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/coercion.kt
raulraja: This is the IDE impl of the refine types predicate. 

The Kotlin plugin provides a script engine factory that connects to the Kotlin compiler daemon in the user IDE and uses it to evaluate the predicate against values such as `TwitterHandle("@admin")` which in this case would fail to show at compile time in the editor that that is not OK. 

Depending on whether we are running the CLI compiler or the IDEA embedded compiler we need to use different implementations.

Then we will register an intention to ask the user to use `TwitterHandle.from(value)` which is a safe API for a runtime value or `TwitterHandle("@whatever")` which is a valid identifier.

raulraja:
@aballano @danimontoya can we get this merged in the branch? I want to use it to see what we want to show in the line markers.

danimontoya:
Its not finished yet, but feel free to merge it!! 👍

aballano:
Merging as WIP to unblock other developments.

PR #502, opened 2020-03-31
Ide coercion intentions

aballano:
Finished first iteration of coercion intentions part of #499 .

Second iteration belongs to coercion epic (#567 )

#### When explicit: 
* [x] Use the inspection text to show a clarification on why you’d wanna use the proof. Show types `From ~> To` and link to the proof declaration. 

#### When implicit: 
* [x] Use the inspection text to show a clarification on what’s happening. Show types `From ~> To` and link to the proof declaration.
* [x] Line marker that explains the coercion being used (text shown and styles TBD.. this will be done in another PR)

#### Test for line markers:
* [x] Test for implicit and explicit coercions (should only be displayed for the first one)

#### Test for applicable inspections:
* [x] Implicit coercion to make it explicit (highlight and applicable fix)
* [x] Explicit coercion to make it implicit (highlight and applicable fix)


idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/explicitcoercion.kt
```diff
@@ -1,51 +1,109 @@
...
+              replaceWithProof(initializer, compilerContext, pairType)
+            }
+          }
+        }
+
+      }
     }
   )
 
+private fun KtPsiFactory.replaceWithProof(element: KtExpression, compilerContext: CompilerContext, pairType: PairTypes) {
```
raulraja: We have the Element Scope for this, we never use the ktPsiFactory directly. See comprehension plugin for examples

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/explicitcoercion.kt
```diff
@@ -1,51 +1,109 @@
...
+                targetType to arg
+              }
+            }
+            // replace them with the explicit version with the proof
+            ?.forEach { (targetType, arg) ->
+              replaceWithProof(arg, compilerContext, targetType)
+            }
+        }
+
+        is KtProperty -> {
```
raulraja: you can move this out to two functions if the KtClass argument on top constrains the type of node it intercepts

raulraja:
Merge at will once comments are addressed, can't wait to try it out :)

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/implicitcoercion.kt
```diff
@@ -27,11 +27,12 @@ fun IdeMetaPlugin.makeImplicitCoercionIntention(compilerContext: CompilerContext
     }
   )
 
-fun KtElement.implicitParticipatingTypes(): Pair<KotlinType, KotlinType>? =
+fun KtElement.implicitParticipatingTypes(): List<PairTypes> =
   when (this) {
 
     is KtDotQualifiedExpression ->
```
i-walker, line 33: this is `safeAs` within a listOfNotNull

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/explicitcoercion.kt
```diff
@@ -1,51 +1,109 @@
...
-fun KtElement.explicitParticipatingTypes(): Pair<KotlinType, KotlinType>? =
+data class PairTypes(val subType: KotlinType, val superType: KotlinType) {
+  companion object {
+    infix fun KotlinType?.pairOrNull(b: KotlinType?): PairTypes? =
+      if (this != null && b != null) PairTypes(this, b)
+      else null
+  }
+}
+
+fun KtElement.explicitParticipatingTypes(): List<PairTypes> =
```
i-walker, line 83: there is a function `traverseFilter` with PsiElement as its receiver, which simplifies this block.

aballano: you mean instead of using this, to call traverseFilter and do the when inside?

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/implicitcoercion.kt
aballano: Done, this simplifies the code a bit, thanks!

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/explicitcoercion.kt
i-walker: Nop, traverseFilter works somewhat `similar` to what we have in Arrow. The difference is that it starts at a Node and traverses its whole tree hierarchy matching on nodes you care about, which it will then transform with your desired function `f` resulting with a List<B>. 
This removes most of the blocks collecting and recollecting Lists. Check out the examples in the ide and their call-sites.

i-walker: Let me check, maybe that signature has not been ported to type-proofs. I push something into the type-proofs branch with what I mentioned above.

i-walker: I admit the example in the ide is not really expressing to what terms you can go but as a demo, you may  consider something like:
```
fun KtElement.participatingTypes(): List<KtValueArgument> =
  traverseFilter(KtCallElement::class.java) { call: KtCallElement ->
    call.valueArgumentList?.arguments?.mapNotNull { valueArgs: KtValueArgument ->
      valueArgs
      /*Do something with valueArgs*/
    }
  }.flatten()
```
That algebra is available in the last [commit](https://github.com/arrow-kt/arrow-meta/commit/ae335d763a138075fc5440a6f90fe5bf79211f27)

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/implicitcoercion.kt
```diff
@@ -1,37 +1,43 @@
...
 import org.jetbrains.kotlin.idea.debugger.sequence.psi.resolveType
 import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
 import org.jetbrains.kotlin.psi.KtElement
-import org.jetbrains.kotlin.types.KotlinType
 
 fun IdeMetaPlugin.makeImplicitCoercionIntention(compilerContext: CompilerContext): ExtensionPhase =
-  addIntention(
-    text = "Make coercion implicit",
+  addApplicableInspection(
+    defaultFixText = "Make_coercion_implicit",
```
i-walker, line 15: this instance has to enabled by Default

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/explicitcoercion.kt
aballano: @i-walker The only problem i see on this specific case is that then I'll need to do a traverseFilter per element type we need to support, no? So in this case it'll be

traverseFilter(KtCallElement::class.java) { ... } 
+
traverseFilter(KtProperty::class.java) { ... }

or something like that, no? But in this case I guess we're gonna iterate over all elements and then filter, so I don't see the benefit vs a normal when but I'm sure i'm missing something 🤔

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/implicitcoercion.kt
aballano: it is already because the parameter is true by default, but I'll add the explicit argument anyway 👍

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/explicitcoercion.kt
i-walker: let's go for the traverseFilter approach and see what we can abstract over most likely you'll abstract over `f` or choose a type higher in the type hierarchy to simplify the work

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/implicitcoercion.kt
i-walker: It didn't show up in my ide. But we already came across that issue. I'll deprecate the default behaivor in `enableByDefault` for that matter

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/explicitcoercion.kt
aballano: > you mean instead of using this, to call traverseFilter and do the when inside?

Then either I'm missing something or I come back to this :D 


```kotlin
fun KtElement.explicitParticipatingTypes(): List<PairTypes> =
  traverseFilter(KtElement::class.java) {
    when (it) {
      is KtCallElement -> {
        // Obtain the argument types from the current call
        val subTypes = it.valueArgumentList?.arguments.orEmpty().mapNotNull { it.getArgumentExpression()?.resolveType() }

        val superTypes = analyze(bodyResolveMode = BodyResolveMode.FULL)
          .getSliceContents(BindingContext.RESOLVED_CALL)
          // get the calls for the current element
          .filter { (call, _) -> call.callElement == this }
          // then the argument types
          .entries.first().value.valueArguments
          .map { it.key.type }

        //TODO check for named and switched arguments!

        subTypes.zip(superTypes, ::PairTypes)
      }

      is KtProperty -> {
        listOfNotNull(it.initializer?.resolveType() pairOrNull it.type())
      }

      else -> emptyList()
    }
  }.flatten()
```

i-walker: these are two different functions. Please dont use `when`, when using traverse. That doesn't make sense

raulraja:
The current sv-test-proofs branch shows these errors:
```
java.lang.IllegalStateException: [2 intention duplicates found for Make coercion explicit (class arrow.meta.ide.dsl.editor.intention.IntentionSyntax$ktIntention$1; plugin io.arrow-kt.arrow)]
	at com.intellij.codeInsight.intention.impl.config.IntentionManagerImpl.checkForDuplicates(IntentionManagerImpl.java:229)
	at com.intellij.codeInsight.intention.impl.config.IntentionManagerImpl.getAvailableIntentionActions(IntentionManagerImpl.java:201)
	at com.intellij.codeInsight.daemon.impl.ShowIntentionsPass.getActionsToShow(ShowIntentionsPass.java:311)
	at com.intellij.codeInsight.daemon.impl.ShowIntentionsPass.doCollectInformation(ShowIntentionsPass.java:226)
	at com.intellij.codeHighlighting.TextEditorHighlightingPass.collectInformation(TextEditorHighlightingPass.java:55)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.lambda$null$1(PassExecutorService.java:429)
	at com.intellij.openapi.application.impl.ApplicationImpl.tryRunReadAction(ApplicationImpl.java:997)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.lambda$doRun$2(PassExecutorService.java:422)
	at com.intellij.openapi.progress.impl.CoreProgressManager.registerIndicatorAndRun(CoreProgressManager.java:585)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeProcessUnderProgress(CoreProgressManager.java:531)
	at com.intellij.openapi.progress.impl.ProgressManagerImpl.executeProcessUnderProgress(ProgressManagerImpl.java:59)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.doRun(PassExecutorService.java:421)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.lambda$run$0(PassExecutorService.java:397)
	at com.intellij.openapi.application.impl.ReadMostlyRWLock.executeByImpatientReader(ReadMostlyRWLock.java:167)
	at com.intellij.openapi.application.impl.ApplicationImpl.executeByImpatientReader(ApplicationImpl.java:157)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.run(PassExecutorService.java:395)
	at com.intellij.concurrency.JobLauncherImpl$VoidForkJoinTask$1.exec(JobLauncherImpl.java:161)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:177)
```

```
During querying provider arrow.meta.ide.dsl.editor.lineMarker.LineMarkerSyntax$addLineMarkerProvider$1@2b40d814 (class arrow.meta.ide.dsl.editor.lineMarker.LineMarkerSyntax$addLineMarkerProvider$1)

kotlin.KotlinNullPointerException
	at org.jetbrains.kotlin.idea.debugger.sequence.psi.KotlinPsiUtilKt.resolveType(KotlinPsiUtil.kt:37)
	at arrow.meta.ide.plugins.proofs.intentions.ExplicitcoercionKt.explicitParticipatingTypes(explicitcoercion.kt:41)
	at arrow.meta.ide.plugins.proofs.markers.CoercionKt.isCoerced(coercion.kt:34)
	at arrow.meta.ide.plugins.proofs.markers.CoercionKt.access$isCoerced(coercion.kt:1)
	at arrow.meta.ide.plugins.proofs.markers.CoercionKt$coerceProofLineMarker$1.invoke(coercion.kt:21)
	at arrow.meta.ide.plugins.proofs.markers.CoercionKt$coerceProofLineMarker$1.invoke(coercion.kt)
	at arrow.meta.ide.dsl.editor.lineMarker.LineMarkerSyntax$addLineMarkerProvider$3.invoke(LineMarkerSyntax.kt:172)
	at arrow.meta.ide.dsl.editor.lineMarker.LineMarkerSyntax$addLineMarkerProvider$3.invoke(LineMarkerSyntax.kt:33)
	at arrow.meta.ide.dsl.editor.lineMarker.LineMarkerSyntax$addLineMarkerProvider$1.getLineMarkerInfo(LineMarkerSyntax.kt:47)
	at com.intellij.codeInsight.daemon.impl.LineMarkersPass.queryProviders(LineMarkersPass.java:170)
	at com.intellij.codeInsight.daemon.impl.LineMarkersPass.lambda$doCollectInformation$3(LineMarkersPass.java:95)
	at com.intellij.codeInsight.daemon.impl.Divider.divideInsideAndOutsideInOneRoot(Divider.java:79)
	at com.intellij.codeInsight.daemon.impl.LineMarkersPass.doCollectInformation(LineMarkersPass.java:90)
	at com.intellij.codeHighlighting.TextEditorHighlightingPass.collectInformation(TextEditorHighlightingPass.java:55)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.lambda$null$1(PassExecutorService.java:429)
	at com.intellij.openapi.application.impl.ApplicationImpl.tryRunReadAction(ApplicationImpl.java:997)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.lambda$doRun$2(PassExecutorService.java:422)
	at com.intellij.openapi.progress.impl.CoreProgressManager.registerIndicatorAndRun(CoreProgressManager.java:585)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeProcessUnderProgress(CoreProgressManager.java:531)
	at com.intellij.openapi.progress.impl.ProgressManagerImpl.executeProcessUnderProgress(ProgressManagerImpl.java:59)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.doRun(PassExecutorService.java:421)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.lambda$run$0(PassExecutorService.java:397)
	at com.intellij.openapi.application.impl.ReadMostlyRWLock.executeByImpatientReader(ReadMostlyRWLock.java:167)
	at com.intellij.openapi.application.impl.ApplicationImpl.executeByImpatientReader(ApplicationImpl.java:157)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.run(PassExecutorService.java:395)
	at com.intellij.concurrency.JobLauncherImpl$VoidForkJoinTask$1.exec(JobLauncherImpl.java:161)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:177)
```

i-walker:
java.lang.IllegalStateException: [2 intention duplicates found for Make coercion explicit (class arrow.meta.ide.dsl.editor.intention.IntentionSyntax$ktIntention$1; plugin io.arrow-kt.arrow)]
Has been issued in #509 

I can't comment on the second error, probably an ill typed variable.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/explicitcoercion.kt
i-walker: Thanks @aballano for the meet. I misunderstood the intention of the code :) 
Traverse is not needed

aballano: Thank you for the time and the explanations! ❤️

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/ProofsIdePlugin.kt
```diff
@@ -23,8 +25,8 @@ val IdeMetaPlugin.typeProofsIde: Plugin
       refinementLineMarkers(),
       addDiagnosticSuppressor { suppressProvenTypeMismatch(it) },
       refinementAnnotator(),
-      coerceProofLineMarker(ArrowIcons.ICON4, ctx)
-      //makeExplicitCoercionIntention(ctx),
-      //makeImplicitCoercionIntention(ctx)
+      coerceProofLineMarker(ArrowIcons.ICON4, ctx),
+      makeExplicitCoercionIntention(ctx),
```
i-walker, line 29:
```suggestion
      explicitCoercionIntention(ctx),
```
nit

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/ProofsIdePlugin.kt
```diff
@@ -23,8 +25,8 @@ val IdeMetaPlugin.typeProofsIde: Plugin
...
+      makeImplicitCoercionIntention(ctx)
```
i-walker, line 30:
```suggestion
      implicitCoercionIntention(ctx)
```
nit

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/intentions/implicitcoercion.kt
```diff
@@ -1,34 +1,31 @@
...
+    applyTo = { ktCall: KtElement, _, _ ->
+      when (ktCall) {
+        is KtDotQualifiedExpression -> {
+          ktCall.replace(ktCall.receiverExpression)
+        }
+      }
+    },
+    inspectionText = { "TODO impl" },
+    inspectionHighlightType = { ProblemHighlightType.INFORMATION },
+    groupPath = ArrowPath + arrayOf("Coercion")
```
i-walker, line 30: here I suggest creating a variable like `ProofPath = ArrowPath + "Type Proofs"`
```suggestion
    groupPath = ProofPath + arrayOf("Coercion")
```

rachelcarmena:
Please, this PR should change the target branch :raised_hands:

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/lineMarker/LineMarkerSyntax.kt
```diff
@@ -244,6 +246,25 @@ interface LineMarkerSyntax {
       }
     )
 
+  fun IdeMetaPlugin.addLineMarkerProviderM(
+    icon: Icon,
+    transform: (PsiElement) -> KtValueArgument?,
```
i-walker, line 251: There is a way to abstract this to any `A : PsiElement` and add a another type variable `B` for the specified child. 
`e.g.: child : Class<B>` instead of `LeafPsiElement::class.java`.
Then we can reuse this for other use-cases.

danimontoya: I just noticed I can remove it completely.. and re-use the one above by tweaking it a little

i-walker: That's even better :)

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/explicitcoercion.kt
```diff
@@ -0,0 +1,102 @@
...
+
+const val EXPLICIT_COERCION_INSPECTION_ID = "Make_coercion_explicit"
+
+val IdeSyntax.explicitCoercionInspectionSyntax: AbstractApplicabilityBasedInspection<KtElement>
+  get() = applicableInspection(
+    defaultFixText = EXPLICIT_COERCION_INSPECTION_ID,
+    inspectionHighlightType = { ProblemHighlightType.INFORMATION },
+    kClass = KtElement::class.java,
+    inspectionText = { "Not used at the moment because the highlight type used is ProblemHighlightType.INFORMATION" },
+    applyTo = { ktCall: KtElement, _, _ ->
```
JorgeCastilloPrz, line 43: @i-walker is there a reason for the current param ordering in this method? could we reorder to have `isApplicable` before `applyTo` on the DSL declaration?

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/explicitcoercion.kt
```diff
@@ -0,0 +1,102 @@
...
+      } ?: false
+    }
+  )
+
+private fun KtElement.makeExplicit(compilerContext: CompilerContext) {
+  when (this) {
+    is KtValueArgument -> {
+      // Get the coerced types (parameter type and actual definition type)
+      participatingTypes().firstOrNull { (subtype: KotlinType, supertype: KotlinType) ->
+        compilerContext.areTypesCoerced(subtype, supertype)
```
JorgeCastilloPrz, line 62: Aren't we checking this already on `isApplicable`?

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/explicitcoercion.kt
```diff
@@ -0,0 +1,102 @@
...
+
+  val notImported = !importList.imports.any { it.importedFqName == importableFqName }
+  val differentPackage = containingKtFile.packageFqName != throughPackage
+
+  if (notImported && differentPackage) {
+    val proofImport = importableFqName?.let {
+      importDirective(ImportPath(it, false)).value
+    }
+    proofImport?.let { importDirective: KtImportDirective ->
+      importList.add(importDirective as PsiElement)// TODO sort?
```
JorgeCastilloPrz, line 93: Import sorting depends on the user code style settings. I say we add it to the end of the imports list and forget (as it is now), at least for the initial release, then we can gather user feedback. They'll get those reorganized based on their code style settings as soon as they hit the required hotkeys.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/propertyCoercion.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.utils.addToStdlib.safeAs
+
+val IdeMetaPlugin.implicitCoercionPropertyLineMarker: ExtensionPhase
+  get() = addLineMarkerProvider(
+    icon = ArrowIcons.ICON4,
+    composite = KtProperty::class.java,
+    transform = { psiElement ->
+      psiElement.ctx()?.let { ctx ->
+        psiElement.safeAs<KtProperty>()?.takeIf {
```
JorgeCastilloPrz, line 22: Aren't we sure it's a `KtProperty` already given the `composite` param?

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/valueArgumentCoercion.kt
```diff
@@ -0,0 +1,54 @@
...
+import org.jetbrains.kotlin.resolve.calls.callUtil.getResolvedCall
+import org.jetbrains.kotlin.types.KotlinType
+import org.jetbrains.kotlin.utils.addToStdlib.safeAs
+
+val IdeMetaPlugin.implicitCoercionValueArgumentLineMarker: ExtensionPhase
+  get() = addLineMarkerProviderM(
+    icon = ArrowIcons.ICON4,
+    composite = KtValueArgument::class.java,
+    transform = { psiElement: PsiElement ->
+      psiElement.ctx()?.let { ctx ->
```
JorgeCastilloPrz, line 26: This structure to check for coercion seems to be replicated in different line markers in this PR, maybe we should provide some sugar for it and reuse to avoid duplication?

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/valueArgumentCoercion.kt
```diff
@@ -0,0 +1,54 @@
...
+        psiElement.safeAs<KtValueArgument>()?.takeIf {
+          it.isCoerced(ctx)
+        }
+      }
+    },
+    message = { ktElement: KtValueArgument ->
+      ktElement.participatingTypes().mapNotNull { (subtype, supertype) ->
```
JorgeCastilloPrz, line 33: Same for this block, even if we end up printing a different message.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/valueArgumentCoercion.kt
```diff
@@ -0,0 +1,54 @@
...
+        ktElement.ctx()?.coerceProof(subtype, supertype)?.coercionMessage()
+      }.firstOrNull() ?: "Proof not found"
+    }
+  )
+
+fun KtValueArgument.participatingTypes(): List<PairTypes> {
```
JorgeCastilloPrz, line 39: Are these `participatingTypes` only found for coercions or they will potentially be needed for all type / subtype relationships in meta for other proofs / features? Mostly thinking on reusability between features.

compiler-plugin/src/test/kotlin/arrow/meta/plugins/refinement/RefinementTests.kt
```diff
@@ -15,18 +15,20 @@ class RefinementTests {
...
-        override val validate: String.() -> Map<String, Boolean> = {
-          mapOf(
-            "Should start with '@'" to startsWith("@"),
-            "Should have length <= 16" to (length <= 16),
-            "Should not contain the word 'twitter'" to !contains("twitter"),
-            "Should not contain the word 'admin'" to !contains("admin")
-          )
+    @Refinement
+    inline class TwitterHandle(val handle: String) {
+        companion object : Refined<String, TwitterHandle> {
```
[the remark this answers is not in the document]
danimontoya: yeah, I copied it from the generated IDE which I guess has different format for indentation

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/lineMarker/LineMarkerSyntax.kt
```diff
@@ -236,7 +238,12 @@ interface LineMarkerSyntax {
...
-      { transform(it)?.identifyingElement },
+      {
+        transform(it)?.let { psi ->
+          (psi as? PsiNameIdentifierOwner)?.identifyingElement
+            ?: PsiTreeUtil.findChildOfType(psi, LeafPsiElement::class.java)
+        }
+      },
       {
         it.onComposite(composite) { psi: A ->
           mergeableLineMarkerInfo(icon, it, { message(DescriptorRenderer.Companion, psi) }, commonIcon, mergeWith, placed, navigate, clickAction)
```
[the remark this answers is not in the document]
danimontoya: Text or styles might be needed to be added within the DSL too?

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/ProofsIdePlugin.kt
```diff
@@ -20,24 +22,31 @@ import org.jetbrains.kotlin.psi.KtClassOrObject
...
       proofLineMarkers(ArrowIcons.ICON1, KtProperty::isGivenProof),
       proofLineMarkers(ArrowIcons.ICON1, KtFunction::isGivenProof),
       refinementLineMarkers(),
-//      refinementAnnotator(),
       proofsKotlinCache,
-      coerceProofLineMarker(ArrowIcons.ICON4),
       addDiagnosticSuppressorWithCtx { suppressProvenTypeMismatch(it) },
-      //makeExplicitCoercionIntention(ctx),
-      //makeImplicitCoercionIntention(ctx),
+      coercionLineMarker,
```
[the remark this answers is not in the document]
danimontoya: `coercionCallSiteLineMarker` sounds good to me 👍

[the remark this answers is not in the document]
danimontoya: I will use `Composite` to group both inspections as well

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/explicitcoercion.kt
danimontoya: these are just named parameters (so we can sort them as we want).. although having certain order can help to reason about it 👍

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/explicitcoercion.kt
danimontoya: true! maybe we don't need to double check it.. going to try it out 👍

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/ProofsIdePlugin.kt
```diff
@@ -20,24 +22,31 @@ import org.jetbrains.kotlin.psi.KtClassOrObject
...
+      explicitCoercionInspection,
+      implicitCoercionInspection,
       codeFolding
     )
   }
 
+val IdeMetaPlugin.typeProofsCli: CliPlugin
+  get() = "Type Proofs Cli Integration".cli {
+    meta(
+      suppressDiagnostic { suppressProvenTypeMismatch(it) }
```
i-walker, line 50: Not sure why that is added here?

compiler-plugin/src/test/kotlin/arrow/meta/plugins/refinement/RefinementTests.kt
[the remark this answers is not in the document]
i-walker: There is no need for the text here. 
Clone the type-proofs demo repo into the ide test env.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/explicitcoercion.kt
danimontoya: yeah I agree.. any other comments on this?

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/lineMarker/LineMarkerSyntax.kt
```diff
@@ -236,7 +238,12 @@ interface LineMarkerSyntax {
     clickAction: AnAction? = null
   ): ExtensionPhase =
     addLineMarkerProvider(
-      { transform(it)?.identifyingElement },
+      {
+        transform(it)?.let { psi ->
```
i-walker, line 242: the registered element here needs to be set by the user. 
That is a wrong assumption that it is either this or the other way with LeafPsiElement

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/lineMarker/LineMarkerSyntax.kt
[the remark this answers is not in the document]
i-walker: That is enriching `mergeableLineMarkerInfo` with more options

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/explicitcoercion.kt
```diff
@@ -0,0 +1,102 @@
...
+import org.jetbrains.kotlin.resolve.ImportPath
+import org.jetbrains.kotlin.types.KotlinType
+import org.jetbrains.kotlin.utils.addToStdlib.safeAs
+
+/**
+ * [explicitCoercionInspection]: for implicit coercion to make it explicit
+ */
+val IdeMetaPlugin.explicitCoercionInspection: ExtensionPhase
+  get() = addLocalInspection(
+    inspection = explicitCoercionInspectionSyntax,
```
i-walker, line 30: weird name IMO

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/propertyCoercion.kt
danimontoya: yes, but its preferable to use `safeAs` than casting it like.. psi as KtProperty

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/explicitcoercion.kt
i-walker, line 30:
```suggestion
    inspection = explicitCoercion,
```

compiler-plugin/src/test/kotlin/arrow/meta/plugins/refinement/RefinementTests.kt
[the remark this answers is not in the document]
rachelcarmena: @i-walker ,this is a Compiler Plugin test. I think it's OK here :+1:

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/valueArgumentCoercion.kt
danimontoya: hmmm not sure about that.. hints are welcome! 😄

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/explicitcoercion.kt
i-walker: the order doesn't matter from a lib perspective

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/explicitcoercion.kt
```diff
@@ -0,0 +1,102 @@
...
+    groupPath = ProofPath + arrayOf("Coercion")
+  )
+
+const val EXPLICIT_COERCION_INSPECTION_ID = "Make_coercion_explicit"
+
+val IdeSyntax.explicitCoercionInspectionSyntax: AbstractApplicabilityBasedInspection<KtElement>
+  get() = applicableInspection(
+    defaultFixText = EXPLICIT_COERCION_INSPECTION_ID,
+    inspectionHighlightType = { ProblemHighlightType.INFORMATION },
+    kClass = KtElement::class.java,
```
i-walker, line 41: Ideally, this is split up in 2 different inspections. We would split this feature based on the ast node it matches on

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/implicitcoercion.kt
```diff
@@ -0,0 +1,58 @@
...
+import com.intellij.codeInspection.ProblemHighlightType
+import org.jetbrains.kotlin.idea.inspections.AbstractApplicabilityBasedInspection
+import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
+
+/**
+ * [implicitCoercionInspection]: for explicit coercion to make it implicit
+ */
+val IdeMetaPlugin.implicitCoercionInspection: ExtensionPhase
+  get() = addLocalInspection(
+    inspection = implicitCoercionInspectionSyntax,
```
i-walker, line 22: same here

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/implicitcoercion.kt
```diff
@@ -0,0 +1,58 @@
...
+    level = HighlightDisplayLevel.WARNING,
```
i-walker, line 23: Is there a reason, why this is a warning and the explicit is an info?

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/valueArgumentCoercion.kt
danimontoya: this I did not understand.. wanna avoid duplication calling participatingTypes ?

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/util.kt
```diff
@@ -0,0 +1,17 @@
+package arrow.meta.ide.plugins.proofs.inspections
+
+import org.jetbrains.kotlin.idea.caches.resolve.analyze
+import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
+import org.jetbrains.kotlin.types.KotlinType
+
+data class PairTypes(val subType: KotlinType, val superType: KotlinType) {
```
i-walker, line 8:
```suggestion
internal data class PairTypes(val subType: KotlinType, val superType: KotlinType) {
```

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/valueArgumentCoercion.kt
danimontoya: mmmm maybe some can be reusable..

idea-plugin/src/main/kotlin/arrow/meta/ide/testing/dsl/lineMarker/LineMarkerTestSyntax.kt
```diff
@@ -37,7 +37,11 @@ interface LineMarkerTestSyntax {
    */
   fun List<PsiElement>.collectLM(icon: Icon): List<LineMarkerInfo<PsiElement>> =
     LineMarkerProviders.getInstance().allForLanguage(KotlinLanguage.INSTANCE)
-      .mapNotNull { mapNotNull { p: PsiElement -> it.getLineMarkerInfo(p) } }.flatten()
+      .mapNotNull {
+        mapNotNull { p: PsiElement ->
+          it.getLineMarkerInfo(p)
+        }
+      }.flatten()
```
i-walker, line 44:
```suggestion
    .mapNotNull { mapNotNull { p: PsiElement -> it.getLineMarkerInfo(p) } }.flatten()
```

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/ProofsIdePlugin.kt
danimontoya: shouldn't be? can I remove it?

idea-plugin/src/test/kotlin/arrow/meta/ide/plugins/proofs/inspections/CoercionInspectionTest.kt
```diff
@@ -0,0 +1,122 @@
...
+  fun `coercion fix inspection test`() =
+    ideTest(
+      myFixture = myFixture,
+      ctx = IdeMetaPlugin()
+    ) {
+      listOf<IdeTest<IdeMetaPlugin, String>>(
+        IdeTest(
+          code = CoercionInspectionTestCode.code1,
+          test = { code: Source, myFixture: CodeInsightTestFixture, _: IdeMetaPlugin ->
+            val file = myFixture.configureByText(KotlinFileType.INSTANCE, code)
```
i-walker, line 70: this can be replaced with code.ktfile()

idea-plugin/src/test/kotlin/arrow/meta/ide/plugins/proofs/inspections/CoercionInspectionTest.kt
```diff
@@ -0,0 +1,122 @@
...
+            val localFixAction = collectInspections(code, myFixture, listOf(explicitCoercionInspectionSyntax))
+              .filter { it.inspectionToolId == EXPLICIT_COERCION_INSPECTION_ID }
+              .flatMap { it.quickFixActionMarkers ?: emptyList() }
+              .map { it.first.action }
```
i-walker, line 74: this is one map or flatMap

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/lineMarker/LineMarkerSyntax.kt
danimontoya: Line markers should target leaf psi elements.. any help to do it differently is welcome!

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/explicitcoercion.kt
danimontoya: removed the check 👍

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/implicitcoercion.kt
danimontoya: Information doesn't highlight (as we want it):
- Enforces {@link com.intellij.codeHighlighting.HighlightDisplayLevel#DO_NOT_SHOW} severity level. Please ensure that if used from inspection explicitly, corresponding problem is added in {@code onTheFly} mode only.
Warning does highlight:
- We want to make the user to go for less verbosity

compiler-plugin/src/test/kotlin/arrow/meta/plugins/refinement/RefinementTests.kt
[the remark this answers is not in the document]
i-walker: code style is the same as in Arrow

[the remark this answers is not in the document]
i-walker: Ups I didn't see that thanks @rachelcarmena :)

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/implicitcoercion.kt
i-walker: Perfect. Thanks for clarifying

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/lineMarker/LineMarkerSyntax.kt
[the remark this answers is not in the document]
JorgeCastilloPrz: @danimontoya yes, our DSL needs to provide support for all tweakeable content we want to support. This is something I'll work on starting today.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/explicitcoercion.kt
JorgeCastilloPrz: Yes Dani! I was taking into account that named params can be ordered at will on call sites, the question is mostly so we could have them "naturally" ordered for us on the declaration too. If you autocomplete they are always inferred / proposed in the order they are declared. Imho that's part of the library public api design and would help both to library contributors and us.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/explicitcoercion.kt
JorgeCastilloPrz: Nope just thinking out loud so we're on the same page :)

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/propertyCoercion.kt
JorgeCastilloPrz: I see. I would expect the DSL to provide you the already casted `KtProperty` in the `transform` block since you already know that is what it is, but not sure if there's a reason to not do that. @i-walker ?

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/valueArgumentCoercion.kt
JorgeCastilloPrz: Both `implicitCoercionValueArgumentLineMarker` and `implicitCoercionPropertyLineMarker` files seem to be really similar. Can't we:
* Move both to use `addLineMarkerProviderM` (might need to update the method to support all scenarios but still we are gonna need a common linemarker provider for all our proofs as discussed, so it takes care of merging or not depending on the case, and we can also append clickable options to it etc)
* move all the `.participatingTypes()` extension functions to a common extensions file, `ParticipatingTypesSyntax` or similar (so they can be reused from both files and from everywhere)
* Pull out `isCoerced` to somewhere else and reuse since it is also replicated. Overall we need to find a way to synthesize our apis and leverage reusability.
* Reencode both markers to reuse mostly everything.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/valueArgumentCoercion.kt
JorgeCastilloPrz: Explained above in a previous comment, let me know if you wanna have a call 🙏

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/valueArgumentCoercion.kt
JorgeCastilloPrz: Explained above

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/ProofsIdePlugin.kt
JorgeCastilloPrz: Isn't it needed to ensure the IDE does not flag coercion call sites as errors because types do not match?

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/explicitcoercion.kt
JorgeCastilloPrz: Isn't that what's happening already for `explicitcoercion.kt` vs `implicitcoercion.kt`?

danimontoya: noo, @i-walker means.. instead of targeting the KtElement and then work with both KtValueArgument and KtProperty split this.. and target independently these two.. right?

idea-plugin/src/main/kotlin/arrow/meta/ide/testing/dsl/inspection/InspectionTestSyntax.kt
```diff
@@ -24,6 +26,17 @@ interface InspectionTestSyntax {
       code.toKtFile(myFixture)?.run { highlighting(myFixture, toIgnore, changes) }
     }.orEmpty()
 
+  fun HighlightInfo.fixFirstInspection(myFixture: CodeInsightTestFixture, file: KtFile?, project: Project): String {
+    quickFixActionMarkers
+      .map { it.first.action }
+      .firstOrNull()?.let { localFixAction ->
+        project.executeWriteCommand(localFixAction.text, null) {
```
i-walker, line 33: myFixture contains the project

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/explicitcoercion.kt
i-walker: those are autocompleted automatically. But when we started this DSL I left those parameters named for a better learning curve for everyone learning and building ide-plugins. We can remove those in the implementation parts of Meta

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/propertyCoercion.kt
i-walker: this part of the DSL has other intentions. 
transform is always the entry point for targeting the elements the user care about. The composite is only there so one can use it for something like `message` or other complementary parameters. Otherwise, the user has to know how to do that.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/ProofsIdePlugin.kt
i-walker: nop

idea-plugin/src/main/kotlin/arrow/meta/ide/testing/dsl/inspection/InspectionTestSyntax.kt
danimontoya: nice, thanks!! 😄

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/lineMarker/LineMarkerSyntax.kt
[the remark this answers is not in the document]
danimontoya: perfect!! 👌

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/explicitcoercion.kt
danimontoya: nooo, lets keep the names.. it helps!

i-walker: It makes reviewing a lot simpler. Without remembering what parameter is in which position

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/valueArgumentCoercion.kt
danimontoya: this has changed today after the refactor.. lets discuss it tomorrow 👍

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/lineMarker/LineMarkerSyntax.kt
```diff
@@ -244,6 +246,30 @@ interface LineMarkerSyntax {
       }
     )
 
+  /**
+   * Similar to [addLineMarkerProviderM], is an extension for PsiElements that are not leafs
```
JorgeCastilloPrz, line 250: Maybe we can add here some explanation about how it takes the non leaf node and tries to find a leaf node from it.

idea-plugin/src/test/kotlin/arrow/meta/ide/plugins/proofs/inspections/CoercionInspectionTest.kt
```diff
@@ -0,0 +1,70 @@
...
+          test = { code: Source, myFixture: CodeInsightTestFixture, _: IdeMetaPlugin ->
+            val file = code.toKtFile(myFixture)
+            val highlightInfos = collectInspections(code, myFixture, listOf(explicitCoercionKtProperty))
+              .filter { it.inspectionToolId == COERCION_EXPLICIT_PROP }
+            val codeFixed = highlightInfos[0].fixFirstInspection(myFixture, file)
+            Pair(highlightInfos, codeFixed)
+          },
+          result = resolvesWhen("CoercionInspectionTest1 for 1 implicit coercion") { pairResult ->
+            pairResult.first.size == 1
+              && pairResult.second == CoercionInspectionTestCode.code1_after_fix
```
i-walker, line 39: this can be simplified and reused as a function. Same below

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/lineMarker/LineMarkerSyntax.kt
i-walker: I'd prefer to move this part to the proofs implementation until there is a general abstraction and not in the algebra.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/coercionInspections.kt
```diff
@@ -0,0 +1,91 @@
...
+
+private fun KtProperty.makeExplicit(compilerContext: CompilerContext) {
+  participatingTypes()?.let { pairType ->
+    initializer?.replaceWithProof(compilerContext, pairType)
+  }
+}
+
+private fun KtExpression.replaceWithProof(compilerContext: CompilerContext, pairType: Pair<KotlinType, KotlinType>) = with(compilerContext) {
+  val through = coerceProof(pairType.first, pairType.second)!!.through
+  val importList = containingKtFile.importList!!
```
i-walker, line 74: please remove the `!!` and add the types to the variables

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/implicitcoercion.kt
```diff
@@ -0,0 +1,47 @@
...
+ * [implicitCoercionInspection]: for explicit coercion to make it implicit
+ */
+val IdeMetaPlugin.implicitCoercionInspection: ExtensionPhase
+  get() = addLocalInspection(
+    inspection = implicitCoercion,
+    level = HighlightDisplayLevel.WARNING,
+    groupPath = ProofPath + arrayOf("Coercion")
+  )
+
+const val IMPLICIT_COERCION_INSPECTION_ID = "Make_coercion_implicit"
```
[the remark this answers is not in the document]
i-walker: same for `COERCION_EXPLICIT_PROP`

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/coercionInspections.kt
```diff
@@ -0,0 +1,91 @@
...
+  } ?: "Proof not found"
+
+internal fun KotlinType?.pairOrNull(b: KotlinType?): Pair<KotlinType, KotlinType>? =
+  if (this != null && b != null) Pair(this, b)
+  else null
+
+internal fun KtExpression.resolveKotlinType(): KotlinType? =
+  analyze(BodyResolveMode.PARTIAL).getType(this)
+
+internal fun KtElement.makeExplicit(compilerContext: CompilerContext) =
```
i-walker, line 46: nit: remove make in the name

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/coercionInspections.kt
```diff
@@ -0,0 +1,91 @@
...
+  when (this) {
+    is KtProperty -> makeExplicit(compilerContext)
+    is KtValueArgument -> makeExplicit(compilerContext)
+    else -> {
+    }
+  }
+
+private fun KtValueArgument.makeExplicit(compilerContext: CompilerContext) {
```
i-walker, line 54: nit: remove `make` in the name

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/coercionInspections.kt
```diff
@@ -0,0 +1,91 @@
...
+    getArgumentExpression()?.let { ktExpression ->
+      val type = ktExpression.resolveKotlinType()
+      if (pairType.first == type) {
+        ktExpression.replaceWithProof(compilerContext, pairType)
+      }
+    }
+  }
+}
+
+private fun KtProperty.makeExplicit(compilerContext: CompilerContext) {
```
i-walker, line 66: nit: remove `make` in the name

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/explicitCoercionArgumentInspection.kt
```diff
@@ -0,0 +1,43 @@
...
+import arrow.meta.phases.ExtensionPhase
+import arrow.meta.plugins.proofs.phases.areTypesCoerced
+import com.intellij.codeHighlighting.HighlightDisplayLevel
+import com.intellij.codeInspection.ProblemHighlightType
+import org.jetbrains.kotlin.idea.inspections.AbstractApplicabilityBasedInspection
+import org.jetbrains.kotlin.psi.KtValueArgument
+
+/**
+ * [explicitCoercionArgumentInspection]: for implicit coercion on arguments to make them explicit
+ */
```
i-walker, line 15:
```suggestion
 * [explicitCoercionArgumentInspection]: adds an explict call for implicit coercions on arguments 
 */
```

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/coercionInspections.kt
```diff
@@ -0,0 +1,94 @@
...
+  analyze(BodyResolveMode.PARTIAL).getType(this)
+
+internal fun KtElement.explicit(compilerContext: CompilerContext) =
+  when (this) {
+    is KtProperty -> explicit(compilerContext)
+    is KtValueArgument -> explicit(compilerContext)
+    else -> Unit
+  }
+
+private fun KtValueArgument.explicit(compilerContext: CompilerContext) {
```
i-walker, line 52: define this as an expression. same down below in the other 2 functions

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/coercionInspections.kt
```diff
@@ -0,0 +1,94 @@
...
+private fun KtProperty.explicit(compilerContext: CompilerContext) {
+  participatingTypes()?.let { pairType ->
+    initializer?.let { compilerContext.replaceWithProof(it, pairType) }
+  }
+}
+
+private fun CompilerContext.replaceWithProof(ktExpression: KtExpression, pairType: Pair<KotlinType, KotlinType>) {
+  coerceProof(pairType.first, pairType.second)?.let { proof ->
+    // Add import (if needed)
+    val importList = ktExpression.containingKtFile.importList
```
i-walker, line 73: missing types

idea-plugin/src/test/kotlin/arrow/meta/ide/plugins/proofs/inspections/CoercionInspectionTest.kt
danimontoya: how

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/implicitcoercion.kt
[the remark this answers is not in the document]
danimontoya: this will be done in another PR 👍

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/inspections/coercionInspections.kt
danimontoya: can you rephrase this?

idea-plugin/src/test/kotlin/arrow/meta/ide/plugins/proofs/inspections/CoercionInspectionTest.kt
i-walker: something like: `fun Source.collectAndApplyInspecction( CodeInsightTestFixture, InspectionToo): Pair<List<HighlightInfo>, Source >`, where the returned Source is the code after the Inspection is applied

danimontoya:
> 👍 Looks good

Thanks for the efforts to improve it and make it better 🙌

PR #507, opened 2020-04-01
Line marker coercion tests

danimontoya:
Test for coercion line markers


idea-plugin/build.gradle
```diff
@@ -24,6 +24,7 @@ dependencies {
 //  compile "org.jetbrains.kotlin:kotlin-scripting-compiler-impl-embeddable:$KOTLIN_VERSION"
 //  compile "org.jetbrains.kotlin:kotlin-script-runtime:$KOTLIN_VERSION"
   compile project(path: ':compiler-plugin', configuration: 'shadow')
+  compile project(":prelude")
```
raulraja, line 27: this should not be here but added in the test fixtures we don't want the IDEA plugin to have the prelude as a binary

PR #517, opened 2020-04-04
sync with proofs branch

i-walker:


i-walker:
Those are the changes in sv-type-proofs branch that also apply in master. For a smoother transition, when the branch is ready.

rachelcarmena:
> Those are the changes in sv-type-proofs branch that also apply in master. For a smoother transition, when the branch is ready.

Great, good idea!! :raised_hands:

i-walker:
Thanks @rachelcarmena 🎉

PR #522, opened 2020-04-06
Patch for ide-refactor in type proofs

i-walker:
Offers the same goodies as in #519 and stabilizes most API's. 

The Ide runs now fine and one can open multiple projects.


raulraja:
Approving assuming this is what was going to master as a patch and you need to merge master in here.

i-walker:
Already done, prior to the approval :)

PR #529, opened 2020-04-09
Revamp Ide Testing DSL

i-walker:
Based on some feedback in testing. There is a new testing dsl in the Ide, 
which supports features described in the PluginScope `e.g.: IdeMetaPlugin`, which can then be used to describe tests. 
The data type IdeTest has now 3 fields the `code`, `test` to define a test and the expected result. 
The remaining dependencies to unfold that data structure into a test are now subject to functions and are not implicitly added. 
So we get the best out of both worlds, ascribing different interpreters for tests and a neat DSL without duplicating code.

### DIFF
--- a/idea-plugin/src/main/kotlin/arrow/meta/ide/testing/env/IdeTestInterpreter.kt
+++ b/idea-plugin/src/main/kotlin/arrow/meta/ide/testing/env/IdeTestInterpreter.kt
@@ -1,90 +1,131 @@
 package arrow.meta.ide.testing.env
 
+import arrow.meta.ide.dsl.IdeSyntax
 import arrow.meta.ide.plugins.helloworld.helloWorld
+import arrow.meta.ide.testing.IdeEnvironment
 import arrow.meta.ide.testing.IdeTest
-import arrow.meta.ide.testing.IdeTestEnvironment
+import arrow.meta.ide.testing.dsl.IdeTestSyntax
 import arrow.meta.ide.testing.dsl.lineMarker.LineMarkerTestSyntax
+import com.intellij.testFramework.fixtures.CodeInsightTestFixture
 import org.junit.Assert
 
 /**
- * [runTest] executes the test with a custom [interpreter], which facilitates to run a test in any environment.
+ * [runTest] executes the test with a custom [interpreter], which facilitates to run a test in any environment, based on a plugin context [F].
  * Hence, the Testing DSL for IntelliJ supports pure or impure Test environments and allows to compose them in a variety of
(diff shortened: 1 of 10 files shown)


idea-plugin/src/main/kotlin/arrow/meta/ide/testing/env/IdeTestInterpreter.kt
```diff
@@ -1,90 +1,131 @@
...
+ * import arrow.meta.ide.testing.env.ideTest
+ * import com.intellij.testFramework.fixtures.CodeInsightTestFixture
+ * import org.junit.Test
+ * //sampleStart
+ * class ExampleTest : IdeTestSetUp() {
+ *   @Test
+ *   fun `general test schema`(): Unit =
+ *     ideTest(
+ *       myFixture = myFixture, // the IntelliJ test environment spins up [myFixture] automatically at runtime.
+ *       ctx = MyIdePlugin() // add here your ide plugin, to get dependencies and created features in the scope of [test] and [result].
```
danimontoya, line 74: nice!

idea-plugin/src/test/kotlin/arrow/meta/ide/plugins/comprehensions/ComprehensionsTest.kt
```diff
@@ -1,75 +1,75 @@
...
+            it.takeIf { descriptor -> descriptor.lineMarker.size == 2 && descriptor.slowLM.isEmpty() }
+          }
+        ),
+        IdeTest(
+          code = ComprehensionsTestCode.code2,
+          test = { code, myFixture, _ ->
+            collectLM(code, myFixture, ArrowIcons.BIND)
+          },
+          result = resolvesWith("LineMarkerTest for two LM on untyped variables") {
+            it.takeIf { descriptor -> descriptor.lineMarker.size == 2 && descriptor.slowLM.isEmpty() }
```
raulraja, line 34: can these repeated expressions be refactored to a function?, they all seem to do the same except change the source code.

i-walker: Yeah for sure, I'll add that to the new PR

i-walker:
Yes, @danimontoya I agree. I'll open a new PR with the changed position.

danimontoya:
> Yes, @danimontoya I agree. I'll open a new PR with the changed position.

Yeah, no problem.. whenever u want!! It is explained anyways 👍

PR #533, opened 2020-04-09
clean up on testing dsl Ide

i-walker:
adds boolean logic to tests algebra
turns around `<A, F>` to `<F, A>` signatures


idea-plugin/src/main/kotlin/arrow/meta/ide/testing/env/types/LightTestSyntax.kt
```diff
@@ -60,4 +60,7 @@ object LightTestSyntax {
    */
   fun KtFile.highlighting(myFixture: CodeInsightTestFixture, toIgnore: List<Int>, changes: (KtFile) -> Boolean = { it.isScript() }): List<HighlightInfo> =
     CodeInsightTestFixtureImpl.instantiateAndRun(this, myFixture.editor, toIgnore.toIntArray(), changes(this)).filterNotNull()
+
+
+
```
i-walker, line 65:
```suggestion
```

PR #534, opened 2020-04-09
reencode Services

i-walker:
adds a better encoding for project and application services.
```kotlin
addProjectService(
  service = MyService::class.java,
  instance = { project: Project, hijackedService: MyService? ->
    object : MyService {
      override fun printLn(f: KtNamedFunction): Unit =
        println("Function: ${f.name} returns ${f.resolveToDescriptorIfAny()?.returnType ?: "ERROR"}")
    }
  }
)
```
This algebra also provides the hijacked services from 3rd parties, given that the services has been instatiated by a plugn one depends.
- [X] add ModuleListener
- [x] update Docs

PR #538, opened 2020-04-13
Doc: update commands and more details about the bot

rachelcarmena:
- Include steps to fix #534 branch
- `-Dkotlin.compiler.execution.strategy="in-process"` is included for running tests by default: #475


.github/workflows/sandbox/README.md
```diff
@@ -20,14 +20,36 @@ $> git diff > <latest-version>.diff
 ```
 2. Add that patch file into `.github/workflows/sandbox/` directory. For instance:
 ```
-.github/workflows/sandbox/1.4.0-dev-180.diff
+.github/workflows/sandbox/1.4.0-dev.diff
 ```
 
 ### What if a patch file for a related version already exists?
 
 For instance, there are additional changes for `1.4.0-dev-325` and `1.4.0-dev-180.diff` already exists (both of them are related to `1.4.0` version).
```
JorgeCastilloPrz, line 28:
```suggestion
For instance, there are additional changes for `1.4.0-dev-325` and `1.4.0-dev.diff` already exists (both of them are related to `1.4.0` version).
```
Should we do this also to match the example with the previous and following steps?

.github/workflows/sandbox/README.md
```diff
@@ -20,14 +20,36 @@ $> git diff > <latest-version>.diff
...
-Create a new `.github/workflows/sandbox/1.4.0-dev-325.diff` with the new required changes.
+Update `.github/workflows/sandbox/1.4.0-dev.diff` with the new required changes.
+
+### What if there are failures when applying a patch on a development branch?
+
+Follow these steps:
+
+1. Create a stash entry for that patch:
+```
+git checkout master
```
JorgeCastilloPrz, line 38: Is this needed? Can't we just rebase `master` on `<dev-branch>` to ensure it's up to date, then apply the patch and solve conflicts?

rachelcarmena: The patch always works on `master` branch. However, it cannot be applied on another branch. That's the reason why starting on `master`. It's the way to get conflicts and not just a `patch cannot be applied` or something like that.

rachelcarmena: The patch always works on `master` branch. However, it couldn't be applied on another branch. That's the reason why starting on `master`. It's the way to get conflicts and not just a `patch cannot be applied` or something like that.

rachelcarmena:
Thanks for the fix and the review, @JorgeCastilloPrz :raised_hands:

PR #414, opened 2020-01-15
Type Proofs. 

raulraja:
WIP, Do not merge.

PR #545, opened 2020-04-16
CompilerContext in Cli and Ide

i-walker:
Now that there is a somewhat artificial trust relationship for the compiler context. We don’t have to carry the burden of applying changes twice in the compiler and Ide, as there is a morphism from `(Project) -> CompilerContext` or `(PsiElement) -> CompilerContext`.

Changes from both sides are shared downstream in the sense, that Cli changes propagate to the Ide and manipulations of the Compiler Context in the Ide are consistent and scoped solely for the Ide. 
This persists the relationship of both endpoints as the Compiler Context goes through multiple phases with different lifecycles.


JorgeCastilloPrz:
Could you elaborate description a bit more @i-walker? I find it a bit cryptic. Could you give an example of a piece of logic where we were applying the same changes "twice in the compiler and Ide"? 🙏 thanks

rachelcarmena:
@i-walker , I think this PR is related to the last commit on `sv-test-proofs`: https://github.com/arrow-kt/arrow-meta/pull/414/commits/2aa5482626b26fe2f57d9b8ff863e23f52da8f5c

However, it doesn't have the same changes. Could it be reverted on `sv-test-proofs`, merge `sv-test-proofs` on `master` branch and then review this PR?

i-walker:
master in the ide takes precedence when it comes to the DSL and Docs.

i-walker:
I've also updated the patch fr the Ide

i-walker:
The changes are conceptually the same, only that the implementation parts of the type-proofs changes due to the proposal.

i-walker:
Sorry @JorgeCastilloPrz for the late answer. 
The underlying idea can be compared to establishing a symbolic messaging contract for the Cli and the Ide.
If one things of the Ide as a superset of the Cli and the information only flows in one direction, (in this case the to the Ide). We can supply a reference to the Ide for whatever dependency is needed in the Ide. 
This technique can be reproduced for any dependency as long as the reference doesn't imply asynchronously computations. The compiler Context is within a synchronous computation, thereby a simple value reference that is passed into the compilation process is sufficient.  
The Ide takes care of distributing and maintaining the lifecycle of a given reference per project.

We don't have tests to validate all aforementioned claims, but this is a simplified method from what other plugins are doing and this comment is solely to explain the intuition behind that. And this somewhat opinioned method may change as we're discovering other options.

PR #562, opened 2020-04-20
Fix typeclass syntax navigation

JorgeCastilloPrz:
We want to be able to click on a type class syntax method call to go to its declaration. E.g:

### Solution

When we generate the synthetic scope for function members we are losing information about the `source` element on the descriptor. We need to explicitly call to `setPreserveSourceElement()` each time we create a custom copy for a function description using the builder, otherwise it'll likely drop it. Source is needed for the IDE to track the declaration down and enable navigation.


rachelcarmena:
Congrats mate :tada: 

Build check fails because of a flaky test:

```
arrow.meta.plugins.refinement.RefinementTests > Runtime validation for nullable types accepts if valid FAILED
 java.lang.ExceptionInInitializerError at RefinementTests.kt:157
 Caused by: java.lang.ClassCastException at RefinementTests.kt:157
```

Re-running again!

JorgeCastilloPrz:
Yep must be flaky, that one seems to be green on my side right now.

Screenshot 2020-04-20 at 12 53 13

PR #563, opened 2020-04-20
Fix dependencies for Java SMT and Z3

rachelcarmena:
@i-walker , mate, please, try this configuration :raised_hands:


rachelcarmena:
Checks are green now after some fixes :wink:

rachelcarmena:
@i-walker , is Z3 solver necessary? Other solvers like `PRINCESS` works. For instance:

```
  println(ctx(SolverContextFactory.Solvers.PRINCESS).version)
```

prints:

```
Princess (2019-11-20 (assertionless))
```

However, solvers like `Z3` or `MathSAT` (native solvers) require native libraries so it's necessary to install them in the OS.

i-walker:
Thanks @rachelcarmena ! 😄 My cache was in a weird state.

rachelcarmena:
Great, happy for it :raised_hands:

Issue #520, opened 2020-04-06
Ide Test Epic


i-walker:
Currently, there is not a lot of tests in the Ide. The goal is to create a testing environment that resembles a typical user experience. 

Feel free to create sub-issues or action items for those features.
The following titles are paired with a file path prefixed with `arrow.meta.ide`, where those features are expected to be implemented.

### Environment `testing.env`

- [ ] A set-up where the underlying Gradle tasks can be performed.
- [ ] A set-up where the results of Gradle tasks, such as `test`, `clean` and `build` result in "meaningful" Messages.
- [ ] A set-up where an ide instance is based on a GitHub project, that is cloned into the instance and opened.
- [ ] A set-up where multiple projects are opened and run in parallel or sequence #510
- [ ] A set-up where a "local" Gradle project is opened to perform Unit-Tests. 

#### Modes `testing.env.modes`

- [ ] One mode, where settings: build, run and executing tests are set to Gradle.
- [ ] One mode, where settings: build, run and executing tests are set to the IDE.
- Low Priority: One mode, where the ide terminal runs build and test.

### Dsl `testing.dsl`

- [ ] `ProjectSyntax` - for multiple, local or github projects.
- [ ] `GradleSyntax` - everything gradle related.
- [ ] `TestModeSyntax` ideally the algebra models something like `(Mode) -> (Env) -> (Env)`

Feel free to refine any part of this issue.

i-walker:
#556 misses the option to open it in a proper manner. 
Executing the following snippet fails with an Access Error.
```
runWriteAction {
       openGradleProject(file, null, true)
}
```
The slack thread [here](https://jetbrains-platform.slack.com/archives/C5U8BM1MK/p1587198354472200) contains more information to the error.

PR #574, opened 2020-04-22
[wip] Test case setup to compile a Gradle project

jansorg:
**still wip**
Relates to https://github.com/arrow-kt/arrow-meta/issues/520
@i-walker 

I decided to re-use JetBrains's Gradle test setup. This is what JetBrains came up with when they developed their integration, so it's mature and useful. If we wrote our own integration, we'd have to write this from scratch and maintain it for the major versions of the IDE. The test framework and the Gradle integration is complex and would need quite some time to integrate properly with a custom solution.

## Possible simplifications
- It's definitely possible to remove a bunch of the JetBrains code. But I'd still keep the basic setup, because this saves a lot of time and is already doing the setup with heavy tests. Because the Gradle projects are on disk we need heavy tests. Our current test setup is using light tests only, if I'm not mistaken.

## Changes:
- a small change to the git cmd integration which allows to pass a custom root path. `project.rootDir` is null with the heavy test setup the Gradle integration is using
- Parts of JetBrains's test framework for [Gradle integration](https://github.com/JetBrains/intellij-community/tree/master/plugins/gradle/testSources/org/jetbrains/plugins/gradle). I've copied this into a separate source directory to allow easier maintenance of this. Because these are Java files this lives under `idea-plugin/src/test/java/...`
- A test, which uses the test IDE's delegated Gradle build to import and compile arrow-kt/arrow-typeproofs: https://github.com/arrow-kt/arrow-meta/pull/574/files#diff-958cded1c7cff78af140a2b6d7ce0c70

## Not yet working:
- Gradle build using the IDE's internal builder (aka JPS). This fails with a compilation error, which is hard to debug. JPS forks a build server in a seperate process, which then calls javac etc. The test is working in the intellij-community sources, but not in our copy (yet). This could be fixed in a seperate PR, of course.


idea-plugin/src/test/kotlin/arrow/meta/ide/plugins/proofs/repo/GradleTypeproofsTest.kt
```diff
@@ -0,0 +1,36 @@
...
+      assertNotNull(ModuleManager.getInstance(myProject).findModuleByName("arrow-typeproofs.main"))
+      assertNotNull(ModuleManager.getInstance(myProject).findModuleByName("arrow-typeproofs.test"))
+
+      compileModules("arrow-typeproofs")
+      assertThatCode {
+        gradle(
+          myProject,
+          projectPath,
+          tasks = listOf(
+            "clean", "build", "test"
```
rachelcarmena, line 30:
```suggestion
            "clean", "build"
```
because `build` task assembles and tests

rachelcarmena:
with #590 :wink:

i-walker:
Thanks to both @jansorg and @rachelcarmena making this happen 🎉

PR #585, opened 2020-04-27
Fix #499: remove hard-coded dependencies for Gradle Plugin

### LINKED ISSUE #499: Stabilize IDE in `sv-test-proofs` branch

rachelcarmena:
## Changes
- Compiler Plugin define all the transitive dependencies
- Gradle Plugin just needs the Compiler Plugin dependency

## Additional checks
- Checked with [the playground for typeproofs](https://github.com/arrow-kt/arrow-typeproofs)
- Shadowed JAR content doesn't change with this improvement.

### DIFF
--- a/compiler-plugin/build.gradle
+++ b/compiler-plugin/build.gradle
@@ -5,7 +5,6 @@ apply plugin: 'com.github.johnrengelman.shadow'
 dependencies {
     // WARNING: don't change the configuration name because it's used by shadowJar
     compileOnly "org.jetbrains.kotlin:kotlin-stdlib:$KOTLIN_VERSION"
-    compileOnly "org.jetbrains.kotlin:kotlin-compiler-embeddable:$KOTLIN_VERSION"
     compileOnly "com.intellij:openapi:$OPENAPI_VERSION"
     compile "org.jetbrains.kotlin:kotlin-scripting-compiler-embeddable:$KOTLIN_VERSION"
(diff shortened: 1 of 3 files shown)


rachelcarmena:
Thanks for the review @i-walker :raised_hands:

PR #586, opened 2020-04-27
Fix doc generation

rachelcarmena:
When documenting doc generation, I realized that Ank build output is not being used. This PR fixes it.


rachelcarmena:
Thanks for the review @AntonioMateoGomez :raised_hands:

rachelcarmena:
It has been deployed and everything is working :+1:

PR #587, opened 2020-04-28
Rename artifact ID for prelude module

rachelcarmena:
The artifact is being published in local repository with the module name (`prelude`) and in OSS repository with the artifact ID (`arrow-meta-prelude`). 

This PR fixes the name for `arrow-meta-prelude` for local repository because that library will be part of Arrow.

### DIFF
--- a/prelude/build.gradle
+++ b/prelude/build.gradle
@@ -25,4 +25,13 @@ tasks.withType(AbstractArchiveTask) {
     reproducibleFileOrder = true
 }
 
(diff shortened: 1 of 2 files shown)


rachelcarmena:
> +1 Makes sense. Thanks @rachelcarmena !

I changed it after @raulraja 's comment :wink:

rachelcarmena:
Thanks for the review :raised_hands:

Issue #567, opened 2020-04-21
[Feature Epic] Coercion


JorgeCastilloPrz:
This covers the complete Coercion feature.

#### When explicit: 

* [x] Applicable Inspection to highlight the expression (as warning) and show a quick fix to convert to implicit 
* [ ] Use the inspection text to show a clarification on why you’d wanna use the proof. Show types From ~> To and link to the proof declaration. 

#### When implicit: 

* [x] Applicable Inspection to highlight the expression (as information) and show a quick fix to convert to explicit.
* [ ] Use the inspection text to show a clarification on what’s happening. Show types From ~> To and link to the proof declaration.
* [ ] Line marker that explains the coercion being used.

PR #599, opened 2020-05-04
Meta Localization for Coercions features

danimontoya:
Setup for localization.
Addresses part of #567

### DIFF
--- a/idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/inspection/InspectionUtilitySyntax.kt
+++ b/idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/inspection/InspectionUtilitySyntax.kt
@@ -1,5 +1,6 @@
 package arrow.meta.ide.dsl.editor.inspection
 
+import arrow.meta.ide.resources.MetaIdeBundle
 import org.jetbrains.kotlin.idea.util.actualsForExpected
 import org.jetbrains.kotlin.idea.util.liftToExpected
 import org.jetbrains.kotlin.psi.KtDeclaration
@@ -12,8 +13,8 @@ interface InspectionUtilitySyntax {
   }
 
   val ArrowPath: Array<String>
-    get() = arrayOf("Kotlin", "Λrrow")
+    get() = arrayOf(MetaIdeBundle.message("proofs.kotlin"), MetaIdeBundle.message("proofs.arrow"))
 
   val ProofPath: Array<String>
-    get() = ArrowPath + arrayOf("Type Proofs")
+    get() = ArrowPath + arrayOf(MetaIdeBundle.message("proofs.type.proofs"))
 }
--- a/idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/coercions/Utils.kt
+++ b/idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/coercions/Utils.kt
@@ -3,6 +3,7 @@ package arrow.meta.ide.plugins.proofs.coercions
 import arrow.meta.ide.dsl.utils.replaceK
 import arrow.meta.ide.plugins.proofs.markers.coercionMessage
 import arrow.meta.ide.plugins.proofs.markers.participatingTypes
+import arrow.meta.ide.resources.MetaIdeBundle
 import arrow.meta.phases.CompilerContext
 import arrow.meta.plugins.proofs.phases.CoercionProof
 import arrow.meta.plugins.proofs.phases.coerceProof
@@ -27,7 +28,7 @@ internal fun KtDotQualifiedExpression.implicitParticipatingTypes(): Pair<KotlinT
 internal fun CompilerContext.coercionProofMessage(ktDotQualifiedExpression: KtDotQualifiedExpression): String =
   ktDotQualifiedExpression.implicitParticipatingTypes()?.let { (subtype, supertype) ->
     coerceProof(subtype, supertype)?.coercionMessage()
-  } ?: "Proof not found"
+  } ?: MetaIdeBundle.message("proofs.markers.proof.not.found")
 
 internal fun KotlinType?.pairOrNull(b: KotlinType?): Pair<KotlinType, KotlinType>? =
   if (this != null && b != null) Pair(this, b)
--- a/idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/coercions/explicit/locaExplicitCoercionOnKtProperty.kt
+++ b/idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/coercions/explicit/locaExplicitCoercionOnKtProperty.kt
@@ -3,6 +3,7 @@ package arrow.meta.ide.plugins.proofs.coercions.explicit
 import arrow.meta.ide.IdeMetaPlugin
 import arrow.meta.ide.plugins.proofs.coercions.explicit
 import arrow.meta.ide.plugins.proofs.markers.participatingTypes
+import arrow.meta.ide.resources.MetaIdeBundle
 import arrow.meta.phases.ExtensionPhase
 import arrow.meta.plugins.proofs.phases.areTypesCoerced
 import com.intellij.codeHighlighting.HighlightDisplayLevel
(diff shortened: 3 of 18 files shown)


idea-plugin/src/main/resources/localization/MetaIdeBundle_en.properties
```diff
@@ -0,0 +1,3 @@
+example.text=Test example
```
i-walker, line 1: Is there a need to have both ? MetaIdeBundle_en and the other if both are in english

danimontoya: No, there is no need of both! still trying things out.. but my guess is that _en wont be needed and default file will be the one without any suffix

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/editor/inspection/InspectionUtilitySyntax.kt
```diff
@@ -12,8 +13,8 @@ interface InspectionUtilitySyntax {
   }
 
   val ArrowPath: Array<String>
-    get() = arrayOf("Kotlin", "Λrrow")
+    get() = arrayOf(MetaIdeBundle.message("proofs.kotlin"), MetaIdeBundle.message("proofs.arrow"))
```
JorgeCastilloPrz, line 16: I'd say we don't want to localize feature names since people always refers to them by their English name, regardless of the language. You'd never say "Pruebas de tipos". Still we'd likely want to have all strings under the same strings file, so I think using the localization system here is preferred 👍 , but I'd suggest removing the Spanish translations for paths (in case the system is able to fallback to the english one, like Android would do for example). These paths just lead to the library features so they'll likely be always English.

idea-plugin/src/main/resources/localization/MetaIdeBundle_es.properties
```diff
@@ -0,0 +1,13 @@
+proofs.kotlin=Kotlin
+proofs.arrow=Λrrow
+proofs.coercion=Coercion
+proofs.type.proofs=Type Proofs
```
JorgeCastilloPrz, line 4: Have you tried not declaring them for all languages? Feature names don't need to be localized, actually I'd say we don't want to, as explained above.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/coercions/Utils.kt
```diff
@@ -27,7 +28,7 @@ internal fun KtDotQualifiedExpression.implicitParticipatingTypes(): Pair<KotlinT
 internal fun CompilerContext.coercionProofMessage(ktDotQualifiedExpression: KtDotQualifiedExpression): String =
   ktDotQualifiedExpression.implicitParticipatingTypes()?.let { (subtype, supertype) ->
     coerceProof(subtype, supertype)?.coercionMessage()
-  } ?: "Proof not found"
+  } ?: MetaIdeBundle.message("proofs.markers.proof.not.found")
```
JorgeCastilloPrz, line 31: Out of scope of this PR, but did you discover why sometimes you're working with the Proofs and then it randomly starts saying "Proof not found" at some point? Looks like if our algorithm to read the proofs or the mechanism we use to store them would be mutated over time or something.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/coercions/explicit/locaExplicitCoercionOnKtProperty.kt
```diff
@@ -17,18 +18,20 @@ import org.jetbrains.kotlin.types.KotlinType
...
+const val COERCION_EXPLICIT_PROP = "CoercionExplicitProp"
 
 val IdeMetaPlugin.explicitCoercionKtProperty: AbstractApplicabilityBasedInspection<KtProperty>
   get() = applicableInspection(
     defaultFixText = COERCION_EXPLICIT_PROP,
+    staticDescription = MetaIdeBundle.message("proofs.coercions.inspection.explicit.property.static.description"),
     inspectionHighlightType = { ProblemHighlightType.INFORMATION },
     kClass = KtProperty::class.java,
-    inspectionText = { "Not used at the moment because the highlight type used is ProblemHighlightType.INFORMATION" },
+    inspectionText = { MetaIdeBundle.message("proofs.coercions.inspection.explicit.property.inspection.text") },
```
JorgeCastilloPrz, line 34: I'd personally revert these, since this one is a disclaimer for the developer given the text will not show up at runtime because `ProblemHighlightType.INFORMATION`, not a string meant for end users to see. Developers likely know English pretty broadly so I'd just keep these like they were, so we keep the disclaimer as explicit as possible and as close as possible to where you're writing the code so you don't miss it when working over these feature.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/coercions/explicit/localExplicitCoercionOnKtValArg.kt
```diff
@@ -17,16 +18,20 @@ import org.jetbrains.kotlin.types.KotlinType
...
+const val COERCION_EXPLICIT_ARGS = "CoercionExplicitArgs"
+
 val IdeMetaPlugin.explicitCoercionKtValArg: AbstractApplicabilityBasedInspection<KtValueArgument>
   get() = applicableInspection(
     defaultFixText = COERCION_EXPLICIT_ARGS,
+    staticDescription = MetaIdeBundle.message("proofs.coercions.inspection.explicit.args.static.description"),
     inspectionHighlightType = { ProblemHighlightType.INFORMATION },
     kClass = KtValueArgument::class.java,
-    inspectionText = { "Not used at the moment because the highlight type used is ProblemHighlightType.INFORMATION" },
+    inspectionText = { MetaIdeBundle.message("proofs.coercions.inspection.explicit.property.inspection.text") },
```
JorgeCastilloPrz, line 34: Same here

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/coercions/implicit/implicitcoercion.kt
```diff
@@ -17,21 +18,23 @@ import org.jetbrains.kotlin.psi.KtSafeQualifiedExpression
...
   get() = applicableInspection(
     defaultFixText = IMPLICIT_COERCION_INSPECTION_ID,
+    staticDescription = MetaIdeBundle.message("proofs.coercions.inspection.implicit.static.description"),
     inspectionHighlightType = { ProblemHighlightType.WARNING },
     kClass = KtDotQualifiedExpression::class.java,
     inspectionText = { ktCall: KtDotQualifiedExpression ->
       // TODO: research ways to display this nicely and align it with [arrow.meta.ide.plugins.proofs.markers.CoercionKt.coerceProofLineMarker]
-      val coercionMessage = ktCall.ctx()?.coercionProofMessage(ktCall)
-      "Expression: ${ktCall.text} can be replaced for only its receiver because there is a $coercionMessage"
+      val coercionMessage = ktCall.ctx()?.coercionProofMessage(ktCall).orEmpty()
```
JorgeCastilloPrz, line 36: When can context be null on this scenario? Looks like the empty state should never happen since it'd be a UX error.

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/ProofMarkers.kt
```diff
@@ -105,9 +106,4 @@ inline fun <reified A : KtDeclaration> IdeMetaPlugin.proofLineMarkers(icon: Icon
   )
 
 fun CoercionProof.coercionMessage(): String =
```
JorgeCastilloPrz, line 108: Maybe we can rename this to make it a bit more clear it's a shared one to be embed into different specific copies.
```suggestion
fun CoercionProof.coercionDefaultMessage(): String =
```

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/purity/PurityPlugin.kt
```diff
@@ -24,14 +24,16 @@ val IdeMetaPlugin.purity: IdePlugin
       addLocalInspection(
         inspection = purityInspection,
         level = HighlightDisplayLevel.ERROR,
-        groupPath = ArrowPath + "PurityPlugin"
+        groupPath = ArrowPath + "PurityPlugin",
```
JorgeCastilloPrz, line 27: Shouldn't we add these also as string resources? The same way you did for the rest of the paths, and just add the English one if that makes sense (since we don't want to translate them).

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/purity/PurityPlugin.kt
```diff
@@ -24,14 +24,16 @@ val IdeMetaPlugin.purity: IdePlugin
...
+        groupDisplayName = "PurityPlugin"
       )
     )
   }
 
 val IdeSyntax.purityInspection: AbstractApplicabilityBasedInspection<KtNamedFunction>
   get() = applicableInspection(
     defaultFixText = "Suspend",
+    staticDescription = "Purity Inspection",
```
JorgeCastilloPrz, line 36: Same here, this should be a string resource.

idea-plugin/src/main/resources/inspectionDescriptions/CoercionExplicitArgs.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+This inspection help us making the coercions within the arguments of an expression to be explicit
```
JorgeCastilloPrz, line 3:
```suggestion
This inspection help us making the coercions within the arguments of an expression to be explicit.
```
I know we're not into final texts yet but let's try to keep syntax correct just in case some texts make it to the final version 👍 

Also: Isn't it possible to get these localized? It'd be weird to have English here and then Spanish in the rest of the UI.

idea-plugin/src/main/resources/inspectionDescriptions/CoercionExplicitProp.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+This inspection help us making the property coercions to be explicit
```
JorgeCastilloPrz, line 3:
```suggestion
This inspection help us making the property coercions to be explicit.
```

idea-plugin/src/main/resources/localization/MetaIdeBundle_en.properties
JorgeCastilloPrz: Let's confirm this and use it that way since otherwise we'll end up with duplicated files for no reason. We also want to be sure so we can even remove alternative translations for stuff we don't want to translate like feature paths for example, that people always refers to by their English names: "Type proofs" vs "Pruebas de tipos".

danimontoya:
Sorry for all the noise.. but couldn't find the way to change the language for the IDE so might not be worth to make efforts here anymore. So we could do:

1. Close the PR
2. Gather all texts into a single file independently of the resources 

cc @raulraja , @i-walker 

Captura de pantalla 2020-05-12 a las 9 16 48

i-walker:
Thanks @danimontoya 💚 
Our efforts in the IDE are discontinued.

PR #600, opened 2020-05-04
Adds remote CSS support for LineMarkers

JorgeCastilloPrz:
WIP, do not review.


i-walker:
May you place the "boilerplate" part of the PR into a directory `arrow.meta.ide.plugins.external` and for this PR the subdirectory would be `ui.linemarker`. 
The `external` directory will contain all changes that need to be internal and won't be exposed to the library. Also once contributions or extensions become available in newer versions we can easily remove them.
We have the same workflow in testing for the ide.

JorgeCastilloPrz:
Why "external" to refer to internal @i-walker ?

i-walker:
Thanks @JorgeCastilloPrz 💚 
Our efforts in the IDE are currently discontinued

Issue #591, opened 2020-04-28
[BUG] Wrong path for generated files


jknazko:
**Describe the bug**
Transform.newSources generates files in strange global path independent of project location.
I still can reference generated code and project (sometimes) builds fine but code usages in Intellij lights red because generated files are not in source set.

On mac is everything generated into: 
**~/Library/Application Support/kotlin/daemon**/build/generated/source/kap/main

I tried to change environment variable _"arrow.meta.generated.source.output"_ but it only changes relative path within that .../deamon folder

**To Reproduce**
I did just a small change in helloWorld example
https://github.com/jknazko/arrow-meta-examples/commit/b4ecb91fb0d074a7f09cce0f7ef394367176f2f0

Doesn't matter if I build the example from IDE or run gradlew from command line result is that helloWorld_Generated.kt and helloWorld_Generated.kt.meta are generated  in ~/Library/Application Support/kotlin/daemon

I see class file in IDE but no source
![Screenshot 2020-04-28 at 13 55 11](https://user-images.githubusercontent.com/47326661/80489670-76ba2d80-8960-11ea-817a-24554865eda1.png)
 
Not sure if this is related or completely different bug but when I make any change to the code after first successful build then it fails with:
![Screenshot 2020-04-28 at 11 37 48](https://user-images.githubusercontent.com/47326661/80490430-9c940200-8961-11ea-9afc-10c275994a9d.png)

I encountered multiple other errors but don't know how to reproduce, whole experience was very random, broken and usually "fixed" with a lot of intellij restarts + clear caches. But Arrow Meta has great potential.

**Expected behavior**
I would expect all generated files to be created in build folder within the project itself.

**Environment**
I always get exactly same behaviour, tried multiple gradle versions with:
Intellij IDEA CE 2019.3.3 - kotlin plugin 1.3.61
Intellij IDEA CE 2020.1 - kotlin plugin 1.3.72

I also replicated it on both MacOS and Linux

russhwolf:
I also saw this behavior recently when messing with the hello world and using `Transform.newSources()`. Wasn't sure if it was intentional or not. That said, the incorrectly-located source does still get linked correctly and I was able to compile code that referenced it. It didn't resolve in the IDE, but I'm not sure that was expected.

rachelcarmena:
Thanks @jknazko. I think there is a misunderstanding with the environment variable `arrow.meta.generated.source.output` because I realized that it's being used with 2 different purposes:

[...]

PR #608, opened 2020-05-10
Dynamic transform new source path

bloderxd:
**This is related to https://github.com/arrow-kt/arrow-meta/issues/591** 

This PR brings a new feature to `Transform.newSources` api that allows we set custom source pathes in its generation:

```kotlin
Transform.newSources(
  """
    package arrow  
    class ${name}_Generated
  """.file("${name}_Generated.kt", "build/generated/source/kapt/test/files")
)
```

an other `file` function with a second parameter allows us to specify its path, also it could be used with multi source generation:

```kotlin
Transform.newSources(
  """
    package arrow       
    class ${name}_Generated 
  """.file("${name}_Generated", "build/generated/source/kapt/test/files"),
  """
    package arrow      
    class ${name}_Generated_2
  """.file("${name}_Generated_2", "build/generated/source/kapt/test/files/source")
)
```

With that it can be used to create structured and architectured generated sources. Also there's a new api to test custom path generated sources:

```kotlin
quoteFileMatches("NewMultipleSourceWithCustomPath_Generated_2.kt",
  """
    package arrow
    class NewMultipleSourceWithCustomPath_Generated_2
  """.source,
  "build/generated/source/kapt/test/files/source"
)
```

With this new api we can specify which source we want to test.

### DIFF
--- a/compiler-plugin/src/main/kotlin/arrow/meta/internal/kastree/ast/Node.kt
+++ b/compiler-plugin/src/main/kotlin/arrow/meta/internal/kastree/ast/Node.kt
@@ -28,7 +28,8 @@ sealed class Node {
     override val pkg: Package?,
     override val imports: List<Import>,
     val commands: List<Command>,
-    val decls: List<Decl>
+    val decls: List<Decl>,
+    val path: String?
   ) : Node(), Entry
 
   data class Script(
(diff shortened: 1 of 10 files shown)


bloderxd:
> Thanks @bloderxd 🎉 I only miss removing:
> 
> ```
> System.setProperty("arrow.meta.generated.source.output", "build/generated/source/kapt/test")
> ```
> 
> in `meta-test/src/main/kotlin/arrow/meta/plugin/testing/CompilerTestDSL.kt`

Then should we remove all dependencies of this property and replace by a mocked directory, since there're its dependency in many part of the code yet

rachelcarmena:
> > Thanks @bloderxd tada I only miss removing:
> > ```
> > System.setProperty("arrow.meta.generated.source.output", "build/generated/source/kapt/test")
> > ```
> > 
> > 
> > in `meta-test/src/main/kotlin/arrow/meta/plugin/testing/CompilerTestDSL.kt`
> 
> @rachelcarmena Then should we remove all dependencies of this property and replace by a mocked directory, since there're its dependency in many part of the code yet?

I mean about removing that change of the system property that can be set by the user. 

I expected: if system property exists then use its value else default value. The same you made here:  https://github.com/arrow-kt/arrow-meta/blob/65ec37c417151b500424c7a3270930dcc52dc5fd/compiler-plugin/src/main/kotlin/arrow/meta/quotes/Quote.kt#L413

PR #609, opened 2020-05-11
Fix transitive dependencies for Arrow Meta IDEA Plugin

rachelcarmena:
`org.celtric.kotlin:kotlin-html` includes a `meta` function that causes conflicts: https://github.com/arrow-kt/arrow-meta-examples/pull/24#issuecomment-626601226


rachelcarmena:
Thanks for the review @aballano :raised_hands:

PR #621, opened 2020-05-14
Configuration: unify IDE Plugin name and add Gradle tasks metadata

rachelcarmena:
* **Unify IDE Plugin name**: It shouldn't end with `Plugin` word.
* **Add Gradle tasks metadata**: It's useful for finding the task to run it. Related documentation will be simpler.


rachelcarmena:
Thanks for the review mate :raised_hands:

PR #631, opened 2020-05-22
Insert coercion proof for value arguments

danimontoya:
We were not testing this scenario where coercion proof needed to be inserted within value arguments.

Scenario:
```
data class UserName(val name: String)
data class Password(val hash: String)

fun help(id: Union2<UserName, Password>): String? {
	val userName: UserName? = id
	val password: Password? = id
	return userName?.name ?: password?.hash
}

val userName = UserName("userName")
val result = help(userName)
```

Ir before inserting the proof:
```
PROPERTY name:result visibility:public modality:FINAL [val]
    FIELD PROPERTY_BACKING_FIELD name:result type:kotlin.String? visibility:private [final,static]
      EXPRESSION_BODY
        CALL 'public final fun help (id: arrow.Union22<<root>.UserName, <root>.Password, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }, arrow.Impossible{ arrow.ImpossibleKt.🚫 }>{ arrow.UnionsKt.Union2<<root>.UserName, <root>.Password> }): kotlin.String? declared in <root>' type=kotlin.String? origin=null
[...]

PR #637, opened 2020-05-26
Fix patch for #631

rachelcarmena:
@danimontoya `expression.descriptor` changes to  `expression.symbol.descriptor` for 1.3.7x.


danimontoya:
It failed because:
```
> Could not resolve org.jetbrains.kotlin:kotlin-stdlib:1.4-M2-eap-83.
```
So I think we can merge it 👍

PR #641, opened 2020-05-27
Temp fix IDE issues

nomisRev:
This PR temporary disables `Log.Verbose`, IDE Quote support and it re-enables the IDE support for Refinement.

- Left `Log.Verbose` disabled, we probably want to re-enable this later?
- Left `"org.celtric.kotlin:kotlin-html:$KOTLIN_HTML_VERSION"` as `implementation`. We need to double check wheter that change can stay.

cc\\ @raulraja

### DIFF
--- a/compiler-plugin/src/main/kotlin/arrow/meta/MetaPlugin.kt
+++ b/compiler-plugin/src/main/kotlin/arrow/meta/MetaPlugin.kt
@@ -18,7 +18,7 @@ open class MetaPlugin : Meta {
     listOf(
       //higherKindedTypes2,
       //typeClasses,
-      comprehensions,
+      // comprehensions,
       //lenses,
       typeProofs
     )
--- a/compiler-plugin/src/main/kotlin/arrow/meta/log/Log.kt
+++ b/compiler-plugin/src/main/kotlin/arrow/meta/log/Log.kt
@@ -10,9 +10,9 @@ sealed class Log {
 operator fun <A> Log.invoke(
   tag: A.() -> String,
   f: () -> A
-): A =
-  if (this is Log.Verbose) {
-    val (time, result) = measureTimeMillisWithResult(f)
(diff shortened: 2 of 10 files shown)


compiler-plugin/src/main/kotlin/arrow/meta/MetaPlugin.kt
```diff
@@ -18,7 +18,7 @@ open class MetaPlugin : Meta {
     listOf(
       //higherKindedTypes2,
       //typeClasses,
-      comprehensions,
+      // comprehensions,
```
rachelcarmena, line 21: @nomisRev , build check is failing because comprehensions test should be disabled because of this change.

rachelcarmena:
And @danimontoya also commented `quotesCli` in IdeMetaPlugin.kt: https://github.com/arrow-kt/arrow-meta/commit/ec7812a4855c7b0c1894a3f833ce5e173ff6f04c

compiler-plugin/src/main/kotlin/arrow/meta/MetaPlugin.kt
rachelcarmena: Thanks @i-walker :raised_hands:

rachelcarmena:
>     * Left `"org.celtric.kotlin:kotlin-html:$KOTLIN_HTML_VERSION"` as `implementation`. We need to double check wheter that change can stay.

It was a mistake from my side because I thought it was a conflict with a `meta` function that is provided by that dependency. The configuration is `implementation` again in the branch that I'm preparing to fix the compatibility issues. So that's right now. Thanks!

i-walker:
I added also Doc's so we won't run into the same rabbit hole again, as we did previously.

i-walker:
Good luck @raulraja  🙌 We're there for you ❤️

raulraja:
Please merge when ready

PR #643, opened 2020-05-27
Performance warning: LineMarker is supposed to be registered for leafs only

danimontoya:
Same fix that we did for coercion within value arguments with the mergeableLineMarkers.
Performance warning: LineMarker is supposed to be registered for leaf elements only.
`KtDeclaration` isn't a leaf.
This fixes performance line markers for: Refinement, Coercion, Extension, Given

### DIFF
--- a/idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/predef.kt
+++ b/idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/predef.kt
@@ -1,0 +1,61 @@
+package arrow.meta.ide.plugins.proofs.markers
+
+import arrow.meta.ide.IdeMetaPlugin
+import arrow.meta.internal.Noop
+import arrow.meta.phases.ExtensionPhase
+import com.intellij.codeInsight.daemon.MergeableLineMarkerInfo
+import com.intellij.openapi.actionSystem.AnAction
+import com.intellij.openapi.editor.markup.GutterIconRenderer
+import com.intellij.psi.PsiElement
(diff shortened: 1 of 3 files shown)


danimontoya:
for value arguments we made them mergeable.. this one is different, right?

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/markers/predef.kt
```diff
@@ -0,0 +1,61 @@
...
+import org.jetbrains.kotlin.renderer.DescriptorRenderer
+import java.awt.event.MouseEvent
+import javax.swing.Icon
+
+/**
+ * Similar to [arrow.meta.ide.dsl.editor.lineMarker.LineMarkerSyntax.addLineMarkerProvider], is an extension for
+ * PsiElements that are not leafs so it will look for the first Leaf corresponding the targeted psiElement
+ */
+@Suppress("UNCHECKED_CAST")
+fun <A : PsiElement> IdeMetaPlugin.addLineMarkerProvider(
```
i-walker, line 21:
```suggestion
internal fun <A : PsiElement> IdeMetaPlugin.addLineMarkerProvider(
```

danimontoya: 👍

i-walker: Excuse that. 
We'll open a refactoring task to unify all those variants.

PR #644, opened 2020-05-27
Provide a Gradle task for running the Intellij Plugin Verifier

rachelcarmena:
## Description

It provides a new Gradle task for running the Intellij Plugin Verifier

```
./gradlew runPluginVerifier
```
That task comes from https://github.com/FWDekker/intellij-randomness

## How to check this pull request

- **Intellij Plugin Verification** check should create the same result as before.


rachelcarmena:
Thanks @danimontoya :tada:

PR #647, opened 2020-05-28
[WIP] proof-of-concept to let find usages show implicit method calls of the proof system

jansorg:
arrow-find-usages

@raulraja 

**This is a proof of concept to show that it's possible**

- It only shows usages in the currently opened files
- It matching methods by name, this isn't good enough for a full implementation

Implementation:
- Takes a similar approach as the Scala plugin
- A methods reference extension adds results to the usage search of a particular method

Missing pieces:
- The extension has to locate all `PsiElements` in the project (or sub-scope), which are method calls to the queried method. I don't know enough about the proof system, but I assume that this data is not yet available. What we need is a mapping `implicitly called method name --> PsiElement where the proof system adds the call`.   
- This mapping has to provide results for all .kt files of the project, which are modified by the proof system
For example, for 
```
val myUnion1 : arrow.Union2<Int,String> = 0
```
the PsiElement of `0` has to be returned, because the proof system adds an implicit call of `first` to it
   - This sounds very similar to the Quote system cache, but I couldn't find the equivalent piece for the proof system

### DIFF
--- a/idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/ArrowImplicitMethodsReferencesSearch.kt
+++ b/idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/ArrowImplicitMethodsReferencesSearch.kt
@@ -1,0 +1,90 @@
+package arrow.meta.ide.plugins.proofs
+
+import arrow.meta.ide.dsl.application.ApplicationSyntax
+import arrow.meta.ide.plugins.proofs.markers.participatingTypes
+import arrow.meta.plugins.proofs.phases.coerceProof
+import com.intellij.openapi.application.QueryExecutorBase
+import com.intellij.openapi.fileEditor.FileEditorManager
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiManager
+import com.intellij.psi.PsiRecursiveElementVisitor
+import com.intellij.psi.PsiReference
+import com.intellij.psi.search.searches.MethodReferencesSearch
+import com.intellij.util.IncorrectOperationException
+import com.intellij.util.Processor
+import org.jetbrains.kotlin.asJava.elements.KtLightMethod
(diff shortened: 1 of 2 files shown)


raulraja:
perfect!

i-walker:
Nice 👍 
I do have a question.
Is it simpler for IntelliJ, if we implement this with an Index, as you mentioned in earlier meeting, regarding Quotes and, which classes should we target for that? 
In the sense that those usages get picked up automatically, without us implementing a search algorithm.

jansorg:
> Is it simpler for IntelliJ, if we implement this with an Index, as you mentioned in earlier meeting, regarding Quotes and, which classes should we target for that?

For an index we'd
1. apply the proof
2. find out which methods are called implicitly
3. store these methods in an index, e.g. a list of `{methodname:String, offset:Int}` for each file
4. use the data of the index to quickly locate where a method with a certain name is called
5. map the offset to a psi element in a file and tell the usage search about it

@i-walker 
Using an index is only possible if a proof is only using data of the current file. 
Is this the case?

If there are proofs, which reference more than the current file, then we can't use an index.

(See the red box on https://www.jetbrains.org/intellij/sdk/docs/basics/indexing_and_psi_stubs/file_based_indexes.html for details on the requirements of an index)

> In the sense that those usages get picked up automatically, without us implementing a search algorithm.

This isn't possible, unfortunately. IntelliJ is only picking those elements automatically, which have proper references pointing to the searched element. This isn't possible due to technical limitations of the integration with the Kotlin plugin.

Therefore the Arrrow plugin has to search for usages on its own. Iterating through all files of a project when find usages is called won't work on larger project.

So, either
1. do caching as in the quote system, i.e. apply transformations for all files at startup and maintain the cache when files change
2. or use an index, this would require much less memory (as far as I can tell). But it's unclear if this is possible (see above)

(1) or (2) have to provide a reverse mapping from name to elements so that find usages is able to locate these elements.

i-walker:
Thanks so much @jansorg  💚 
Our efforts in the IDE are currently discontinued

PR #648, opened 2020-05-28
Meta extension for KtCallExpression.

mattmoore:
WIP: Do not merge.

Adding scope for call expressions as this will make pattern matching desugaring transformations easier to handle (on separate branch pattern-matching, will rely on this PR once it's ready to merge).

Still need to verify the KtCallExpression attribute semantics, so putting this into WIP status.


raulraja:
@mattmoore, thanks for this, we are closing some outdated and old PRs since the layout has merged considerably and the code base is about to change more. Feel free to bring it back if it's still relevant at any time. Thank you!

PR #651, opened 2020-05-29
Pattern matching

mattmoore:
Plugin to support pattern matching.

KEEP for adding this to the compiler: https://github.com/Kotlin/KEEP/pull/213

After discussion, it was decided that implementing this as part of the Arrow Meta plugin system would be ideal for building a working prototype. Perhaps down the road this could be implemented in the compiler but it would probably take some time to get approved.

There are tentatively two variations to research:
- Position-based pattern matching, similar to Scala (with wildcards and other patterns), e.g.: `Person(firstName, _)`
- Field-based pattern matching, where the parameter name can be specified (thereby avoiding problems that come from position-dependent parameters), e.g.: `Person(firstName == "Matt")`

Fixes https://github.com/arrow-kt/arrow/issues/1470


compiler-plugin/src/main/kotlin/arrow/meta/plugins/patternMatching/Desugar.kt
```diff
@@ -0,0 +1,19 @@
...
+  get() =
+    firstChild.nextSibling
+      .firstChild.nextSibling
+      .text.replace("_", "person.firstName")
+
+val BindingTrace.desugar
+  get() =
+    entriesFollowing(analysisPatternRules).forEach { entry ->
+      // TODO Write replacement entry
+      val parentCaseExpression = entry.key.parent.parent.parent.parent.parent.parent
```
raulraja, line 17: this can be done with findParent or similar

compiler-plugin/src/test/kotlin/arrow/meta/plugins/patternMatching/PatternMatchingTests.kt
```diff
@@ -0,0 +1,44 @@
...
+
+  @Test
+  fun `rewrite constructor pattern match expression`() {
+    val code =
+      """$personSource
+         fun case(arg: Any): Person = TODO("Deal with later...")
+         
+         @Suppress("UNRESOLVED_REFERENCE")
+         val result = when (person) {
+           case(Person(_, "Moore")) -> "Matched"
```
nomisRev, line 32: 😍

compiler-plugin/src/main/kotlin/arrow/meta/plugins/patternMatching/Desugar.kt
mattmoore: Is there a build-in method for this? Haven't pushed it yet, but didn't find one so just made a recursive function to search upwards.

raulraja:
@mattmoore, thanks for this, we are closing some outdated and old PRs since the layout has changed considerably and the code base is about to change more. Feel free to bring it back if it's still relevant at any time. Thank you!

PR #661, opened 2020-06-06
Enable all Ide tests

### LINKED ISSUE #659: [BUG] Meta extensions are not deregistered
Branch ja-fix-quotetests
See PR #656 

1. checkout that branch
1. Run all tests of the idea-plugin module
1. Tests will fail

Running a test in isolation doesn't fail. Running all tests makes tests fail. The reason is that the [Quotesystem's extensions](https://github.com/arrow-kt/arrow-meta/blob/6b2511c9234b0797a1484fdd3fc326d100d0e114/idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/quotes/system/QuoteSystem.kt#L44) contain extensions of disposed projects.

1. Each test creates a project.
1. For each project the extensions are registered. It seems that they're kept in a global map
1. A project is disposed when a test terminates. But the extension, which references it, is still registered
1. As soon as this extension is used during another test run, this breaks the tests, because the dispose state is checked

Scope: This isn't just a problem for tests, as far as I understand. The same will happen when you open a project in an ide, work with it and close it. When you now open another project, the extensions referencing the old project should still be around.

See the disposed state of the selected project:
Screenshot_012

i-walker:
To further align the lifecycle of quotes with the Ide and Cli, we introduce in the CompilerContext a mutable list with all defined quotes and as with #545 the lifecycle of those definitions are automatically maintained and disposed accordingly. 

For simplicity, the `QuoteSystemService` reduces it's members to the core function, that applies quotes to a given file.
Hijacking and manipulating this service results solely in Ide related changes, but the Cli is unaffected from that. 
This gives plugin authors the ability to fine-tune quote manipulations at will for their on Ide features.

Thanks, @jansorg for pointing out the Bugs!
resolves #659

### DIFF
--- a/compiler-plugin/src/main/kotlin/arrow/meta/Meta.kt
+++ b/compiler-plugin/src/main/kotlin/arrow/meta/Meta.kt
@@ -26,7 +26,7 @@ data class Plugin<A>(
  * val Meta.helloWorld: CliPlugin get() =
  *   "Hello World" {
  *     meta(
- *       namedFunction({ name == "helloWorld" }) { c ->  // <-- namedFunction(...) {...}
+ *       namedFunction(this, { name == "helloWorld" }) { c ->  // <-- namedFunction(...) {...}
  *         Transform.replace(
  *           replacing = c,
  *           newDeclaration = """|fun helloWorld(): Unit =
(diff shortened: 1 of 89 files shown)


raulraja:
Thanks @i-walker @jansorg !

PR #665, opened 2020-06-10
Make ide stable with version 2019.3.5

danimontoya:
The IDE got more stable with the fix done in #692 .
So now this PR approaches some other fixes/features also improving stability and a more friendly UI:
- Remove some logging (excessive logging creates laggy behaviour)
- Add `staticDescription` that refers to the little description of the applicable inspections
- Fix issue with typealias for Unions and Tuples foldingBuilders
- Remove duplicate refinement line marker
- Fix line marker for proofs, pointing to the descriptor instead of comments or annotation
- Revisit text displayed for given, coercion and extension (inspections and annotations)
- Replace coercion line markers with annotations and styled like implicits in Scala language
- Replace added given line markers for injections with annotations (to be aligned with the coercions and have better scalability)
- Add renderer to display nicely the function information (maybe this renderers aren't needed with the new versions of IDEA, since most of the information gathered is now provided by default.. like showing KDocs and some other information)
- Add `fix` within the annotator to let the users navigate to the function definition
- Add null at the beginning of the folding instead of at the end for increasing visibility in case of nullabillity


i-walker:
Please don't forget to merge master :D to enable the ide tests and quotes

PR #666, opened 2020-06-10
Add Proof references

i-walker:
DuplicatedFunReferences is one example how to register PsiReferences , where
`pattern` denotes, which PsiElement is populated with the PsiReferences in the function 
`references`.
For the one taking this task over. I left a bunch of comments, where and how to get started. Please also review this PR #647 to get an intuition for the detection part, even though the PR matches on strings. 
One may note that populating PsiReferences that already are registered to an Element, are being collapsed.


i-walker:
RIP Ide 💚

PR #683, opened 2020-06-20
Add proof resolution

i-walker:
Implemented resolution strategies:
**In general**
* Internal orphan override
  - [X] Internal orphans take precedence over publicly available proofs.
  - [X] Internal orphans are not permitted to be published publicly, to preserve coherence in the ecosystem.
* Ambiguity
  - [x] Disallow ambiguity for internal orphans
  - [x] Disallow ambiguity for public proofs
  - [X] If the compiler runs into ambiguities, as Jars in the Jvm can be instrumented it shall force the user to define an internal orphan.
* Resolution scope
  - [X] ~~Package scoped resolution of proof instances.~~ The former is not needed for coherent resolution. User can declare their proofs as top-level extension functions in the case of `Extension Proofs`, same goes for Given and Refinement Proofs - Descriptors need to be at the top-level.
  - [X] Add compiler error's for skipped instances, as they break coherence. Opt the user to declare an internal proof to restore coherence.
* Ownership
  - [X] Ownership rules e.g.: Given proofs can only be declared public if the User owns the type.

**Refinement Proofs**

- [x] For a Refinement from a target type `A` to it's refined zero-cost abstraction `A'`. The companion object of `A'` has to implement this interface `Refined<A, A'>`, which corresponds to `A -> A'`.

**Extension Proofs**

**Given Proofs**
- [X] Semi-inductive resolution rules.
- [x] Call-site resolution for missing or existing Given proofs

Further Improvements:
- Predicate Errors are being reported through Diagnostics rather than the MessageCollector
- Task logs are being printed whether for failed/ canceled or successful Gradle tasks for git repo's in the Ide
-  The parameter `check` in `storageComponent` is being registered. For convenience, there is `declarationChecker` function.

This PR uses Diagnostic Error's, which will be picked up, by the Ide to supply QuickFixes, compile-time errors, or other features.
In addition, to prevent duplication of identifying targets.
Utilizing those Diagnostic Error's in the Ide is still WIP and will be added in further PR's.


i-walker:
- Regarding published internal proofs: the message is only preliminary.
 
Screenshot 2020-06-22 at 16 48 16

compiler-plugin/src/main/java/arrow/meta/diagnostic/MetaErrors.java
```diff
@@ -0,0 +1,49 @@
...
+    DiagnosticFactory1<KtElement, String> RefinementValidationError = DiagnosticFactory1.create(ERROR);
+
+    DiagnosticFactory2<KtObjectDeclaration, KotlinType, KotlinType> IncorrectRefinement = DiagnosticFactory2.create(ERROR, onIdentifyingElement);
+
+    DiagnosticFactory2<KtObjectDeclaration, KotlinType, KotlinType> TooManyRefinements = DiagnosticFactory2.create(ERROR, onIdentifyingElement);
+    /**
+     * needed to prevent NPE in
+     * org.jetbrains.kotlin.diagnostics.rendering.DefaultErrorMessages#getRendererForDiagnostic(org.jetbrains.kotlin.diagnostics.Diagnostic)
+     */
+    @SuppressWarnings("UnusedDeclaration")
```
raulraja, line 43: :O

compiler-plugin/src/test/kotlin/arrow/meta/plugins/proofs/ResolutionTests.kt
```diff
@@ -0,0 +1,397 @@
...
+      source = """
+        @Given
+        internal val x: (Int) -> String = { i -> i.toString() }
+        val result = given<(Int) -> String>().invoke(5)
+      """) {
+      "result".source.evalsTo("5")
+    }
+  }
+
+  @Ignore // TODO: Add compiler Error after fixing resolution
```
i-walker, line 278: Here it's the same as in the comment above

compiler-plugin/src/main/kotlin/arrow/meta/plugins/proofs/phases/ir/ProofsIrCodegen.kt
```diff
@@ -59,16 +72,15 @@ class ProofsIrCodegen(
       }
     }
   }
-
+  //TODO handle ambiguity and orphan selection
```
raulraja, line 75: if ambiguity and orphan selection has been done in the resolution phase then in IR is already guaranteed to not be ambiguous and we can remove this TODO
```suggestion
```

compiler-plugin/src/test/kotlin/arrow/meta/plugins/proofs/ResolutionTests.kt
```diff
@@ -0,0 +1,397 @@
...
+    resolutionTest(
+      """
+      @Coercion
+      fun String.toInt16(): Int? =
+        toIntOrNull(16)
+        
+      val x: Int? = "30"
+      """) {
+      failsWith {
+        it.contains("This ExtensionProof test.toInt16: kotlin.String -> kotlin.Int? violates ownership rules, because public Proofs over 3rd party Types break coherence over the kotlin ecosystem. One way to solve this is to declare the Proof as an internal orphan.")
```
raulraja, line 25: I think we can drop the whole breaks coherence and provide a link instead to the docs of meta where this is explained. Otherwise, we should just simply render the signature of the function with the `internal` keyword for users to just replace it with the suggestion or copy and paste the suggestion from their terminal

raulraja (approved):
Changes look great!

compiler-plugin/src/main/java/arrow/meta/diagnostic/MetaErrors.java
i-walker: I know, It's a weird hack :P 
But I found the same code in the compiler

compiler-plugin/src/test/kotlin/arrow/meta/plugins/proofs/ResolutionTests.kt
i-walker: Sure, we can leave a TODO when the link to the Docs exists. 
I'll add an Issue with Arrow Error messages, just like we have with https://github.com/arrow-kt/arrow-meta/issues/131#issuecomment-555540571

compiler-plugin/src/main/java/arrow/meta/diagnostic/MetaDefaultErrorMessages.java
```diff
@@ -0,0 +1,39 @@
+package arrow.meta.diagnostic;
```
rachelcarmena, line 1: Awesome @i-walker :raised_hands: Please, what's the reason of **Java** source for several classes in `compiler-plugin`? I know that Java sources were added in `idea-plugin` because of a copy-paste from other repository (see `README.txt` in `idea-plugin` sources). Is it the same case ? If yes, could we add that repository? Thank you so much!!

rachelcarmena (approved):
I've just added a comment :raised_hands:

compiler-plugin/src/main/java/arrow/meta/diagnostic/MetaDefaultErrorMessages.java
i-walker: The Java sources are necessary for the Error Message classes. They have a rather special initializer, which I got only working in Java, but tried to replicate in Kotlin with no success. 
The compiler does the same for `org.jetbrains.kotlin.diagnostics.Errors`.
To resolve that, one would have to change a private member of the DiagnosticFactory, so the NPE doesn't occur at Runtime loading the class. 
It's a bug-fix, I can contribute if that is desired.

rachelcarmena: Great, thanks for the info @i-walker :raised_hands: It would be awesome if you could add a README file like Joachim made in `idea-plugin` to add the link of those sources for future updates.

i-walker:
Closing notes:

* The current Ir Error that for instance: 
```
interface Box<T> {
  val value: T
}

@Given
val a: Box<Int> = object : Box<Int> {
  override val value: Int = 8
}

@Given
val b: Box<String> = object : Box<String> {
  override val value: String = "Mellow"
}

val wrongBox: (@Given() Box<String>) -> String = {
  it.value
}
val result: String = wrongBox() // Typemismatch resulting in Int, but String was expected.
```
This will cause 2 error's first the wrong instance will be injected in `wrongBox`, which leads to the Typemismatch in `result`. 

The error will be handled in a subsequent PR, regarding `chooseSpecific` to inject the right instance.

* The Ide will be updated to 2020.2.2 stable for Arrow migration

PR #702, opened 2020-07-06
Fix tuples code folding

danimontoya:
- Use the package from prelude instead of arrow core
- Add truncate mechanism for tuples
- Add tuples operators up until 22


idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/folding/tuples.kt
```diff
@@ -23,20 +25,27 @@ fun KtTypeReference.tupleTypeMatches(): Boolean =
     }
 
 private fun KotlinType?.isTypeMatching() =
```
i-walker, line 27: nit:
Maybe it's worth to set a parameter here with the FqName we want to target.

danimontoya: thanks @i-walker ! updated! 👍

PR #707, opened 2020-07-08
Pattern matching - support sealed classes

ShikaSD:
Converts `case(Number.One(false)) -> ` into `is Number.One && it.component1() == false`


ShikaSD:
I just have merged master, so those ones related to your changes, not mine :)

rachelcarmena:
> I just have merged master, so those ones related to your changes, not mine :)

Oh, I see... However, I think this branch shouldn't be updated with `master` branch but just `pattern-matching`, the target branch. I mean, I'm thinking about the review moment by @mattmoore. 

Thanks for the info @ShikaSD !!

ShikaSD:
No worries :)
I think we can upstream base branch, then those two will even out each other

PR #711, opened 2020-07-13
Testing adding local inspections through the plugin.xml

danimontoya:
Adding the local inspections through the plugin.xml is possible to set them as `enabledByDefault = true`

If we want to provide to users a DSL to add their localInspections, I'm not sure if its possible with the current API intellij provides.. since this field is not possible to be overridden.

Question: Maybe we can keep the DSL and find an alternative to enable the localInspections by default (when corresponding) in another phase?


danimontoya:
This draft isn't needed since coercion local inspections are refactored using annotations: #717

PR #712, opened 2020-07-14
Add/Update messages for refined line markers 

danimontoya:
- Add message when the user is hovering the mouse over the refined line marker annotation
- Update line marker message for refined predicates (remove quotes)


rachelcarmena (approved):
Thanks @danimontoya :raised_hands:

PR #706, opened 2020-07-07
Upgrade to 1.4.10

rachelcarmena:
## Tasks

- [x] Configuration for 1.4.0-rc
- [x] Use of related Kotlin IDE Plugin
- [x] Remove the need of `arrow-annotations` or `arrow-core-data` libraries (moved classes to prelude #713)
- [x] Solve compilation errors
- [x] Upgrade testing library
- [x] Upgrade source code according to [compatibility list](https://youtrack.jetbrains.com/issues/KT?q=Tag:%20language-committee-approved%20Target%20versions:%201.4-M3,%201.4-M2,%201.4-M1,%201.4.0)
- [x] Create a branch for the playground (used in an Arrow Meta IDE test)
- [x] Solve errors with doc validation

## Notes

Kotlin 1.4.x is not backward-compatible with 1.3.x so it's necessary to review the [compatibility list](https://youtrack.jetbrains.com/issues/KT?q=Tag:%20language-committee-approved%20Target%20versions:%201.4-M3,%201.4-M2,%201.4-M1,%201.4.0) (1.4-M3 is a pre-release version so a different compatibility list like https://kotlinlang.org/docs/reference/compatibility-guide-13.html is not available).

For instance, I was reviewing the failure with KindTest and two Kind22 types are considered different now and it impacts in this evaluation from `arrow.meta.plugins.proofs.phases.ir.ProofsIrCodegen`:

```
  private fun CompilerContext.insertExtensionSyntaxCall(expression: IrCall) {
    val valueType = ...
    val targetType = ...
    if (targetType != null && valueType != null && targetType != valueType && !baseLineTypeChecker.isSubtypeOf(valueType, targetType)) {
     ...
```

* **Update**: https://kotlinlang.org/docs/reference/compatibility-guide-14.html

## Comment about latest DEV version

I tested the latest DEV version as well and the compilation fixes can be found here: https://github.com/arrow-kt/arrow-meta/blob/activate-checks/.github/workflows/sandbox/1.4.20-dev.diff However, it's not possible to check the IDE plugin because of an incompatibility with Kotlin IDE plugin:

> Class 'org.jetbrains.kotlin....' is compiled by a pre-release version of Kotlin and cannot be loaded by this version of the compiler

PR #720, opened 2020-07-21
Configuration: activate daily checks for 1.4-x

rachelcarmena:


rachelcarmena:
Closing and re-opening #706 because 1.4.x is not backward-compatible with 1.3.x and it's necessary to adapt the source code (beyond solving compilation issues).

PR #725, opened 2020-08-06
Configuration: specific classpath for Ank validation

rachelcarmena:
## Reason

Current errors when validating the documentation for `1.4.0-rc`. There is no clue to solve them.

If all the dependencies from compile classpath and runtime classpath (295 dependencies) are added then Ank raises a lot of errors from conflicts between them so the classpath for Ank validation includes:

* The whole runtime classpath: 31 dependencies; some of them with specific Kotlin version for Ank
* Specific dependencies from Intellij IDEA + Kotlin IDEA Plugin

## Changes

The classpath for Ank validation will include:

* Specific dependencies with the current Kotlin version: 24 dependencies
* Specific dependencies from Intellij IDEA + Kotlin IDEA Plugin

## Advantages

That control about dependencies for Ank validation is useful for upgrades because just the expected dependencies are considered.

### DIFF
--- a/docs/build.gradle
+++ b/docs/build.gradle
@@ -17,6 +17,14 @@ apply plugin: 'ank-gradle-plugin'
 dependencies {
     compile project(':idea-plugin')
     compile "junit:junit:4.13"
+    runtimeOnly "org.jetbrains.kotlin:kotlin-compiler:$KOTLIN_VERSION"
+    runtimeOnly "org.jetbrains.kotlin:kotlin-scripting-compiler:$KOTLIN_VERSION"
+    runtimeOnly "org.jetbrains.kotlin:kotlin-scripting-compiler-impl:$KOTLIN_VERSION"
+    runtimeOnly "org.jetbrains.kotlin:kotlin-script-util:$KOTLIN_VERSION"
+    runtimeOnly "org.jetbrains.kotlin:kotlin-reflect:$KOTLIN_VERSION"
+    runtimeOnly "org.jetbrains.kotlin:kotlin-scripting-jvm:$KOTLIN_VERSION"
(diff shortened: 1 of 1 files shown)


docs/build.gradle
```diff
@@ -17,6 +17,14 @@ apply plugin: 'ank-gradle-plugin'
...
     compile project(':idea-plugin')
     compile "junit:junit:4.13"
+    runtimeOnly "org.jetbrains.kotlin:kotlin-compiler:$KOTLIN_VERSION"
+    runtimeOnly "org.jetbrains.kotlin:kotlin-scripting-compiler:$KOTLIN_VERSION"
+    runtimeOnly "org.jetbrains.kotlin:kotlin-scripting-compiler-impl:$KOTLIN_VERSION"
+    runtimeOnly "org.jetbrains.kotlin:kotlin-script-util:$KOTLIN_VERSION"
+    runtimeOnly "org.jetbrains.kotlin:kotlin-reflect:$KOTLIN_VERSION"
+    runtimeOnly "org.jetbrains.kotlin:kotlin-scripting-jvm:$KOTLIN_VERSION"
+    runtimeOnly "org.jetbrains.kotlin:kotlin-scripting-common:$KOTLIN_VERSION"
+    runtimeOnly "org.jetbrains.kotlin:kotlin-script-runtime:$KOTLIN_VERSION"
```
rachelcarmena, line 27: It's not necessary to add these dependencies to make it work now because current Ank and Meta are using the same Kotlin version. However, it would be necessary to avoid problems in other situations.

rachelcarmena:
Thanks for the review @i-walker, @danimontoya :raised_hands:

Issue #663, opened 2020-06-06
populate IntegrationSyntax


i-walker:
There are a couple of missing extension to complement cli-ide integration as proposed in #652 
The steps are similar for each extension X in the bucketlist below:
1) Create an extension in `arrow.meta.ide.phases.integration` with the same name as X [here](https://github.com/arrow-kt/arrow-meta/blob/9a8529614c78fd33652b03c28110bedfc7434019/idea-plugin/src/main/kotlin/arrow/meta/ide/phases/integration/integration.kt#L7-L6) are a couple of examples 
```kotlin
package arrow.meta.ide.phases.integration
//...

interface X : ExtensionPhase {
  fun `meaningfull function name`(project : Project): arrow.meta.<...>.X
}
```

2) Add an extension function that creates X in `arrow.meta.ide.dsl.integration.IntegrationSyntax` [examples](https://github.com/arrow-kt/arrow-meta/blob/c26453a5b69b81c66ac3c940590506529aefd025/idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/integration/IntegrationSyntax.kt#L18)

3) Register X in `IdeInternalRegistry` [examples](https://github.com/arrow-kt/arrow-meta/blob/882ea6cf86cea51be305c722feb4f913d08f08f7/idea-plugin/src/main/kotlin/arrow/meta/ide/internal/registry/IdeInternalRegistry.kt#L91)

Feel free to ask in this issue, if there are any blockers or questions 😄 

Bucketlist:
- [ ] arrow.meta.phases.resolve.synthetics.SyntheticScopeProvider
- [ ] arrow.meta.phases.codegen.ir.IRGeneration
- [ ] arrow.meta.phases.resolve.DeclarationAttributeAlterer
- [ ] arrow.meta.phases.codegen.asm.Codegen
- [ ] arrow.meta.phases.codegen.asm.ClassBuilder
- [ ] arrow.meta.phases.analysis.AnalysisHandler
- [ ] arrow.meta.phases.config.StorageComponentContainer
- [ ] arrow.meta.phases.analysis.PreprocessedVirtualFileFactory
- [ ] arrow.meta.phases.analysis.ExtraImports
- [ ] arrow.meta.phases.config.Config
- [ ] arrow.meta.phases.analysis.CollectAdditionalSources

Starmel:
I take it

i-walker:
resolved in #727 
Thanks to @Starmel

PR #727, opened 2020-08-11
#663 Populate IntegrationSyntax

Starmel:
Added phases for Ide from #663 issue.
Added suffix `Ide` - `IdeSyntheticScopeProvider` to simplify imports 

- [X]  arrow.meta.phases.resolve.synthetics.SyntheticScopeProvider
- [X]  arrow.meta.phases.codegen.ir.IRGeneration
- [X]  arrow.meta.phases.resolve.DeclarationAttributeAlterer
- [X]  arrow.meta.phases.codegen.asm.Codegen
- [X]  arrow.meta.phases.codegen.asm.ClassBuilder
- [X]  arrow.meta.phases.analysis.AnalysisHandler
- [X]  arrow.meta.phases.config.StorageComponentContainer
- [X]  arrow.meta.phases.analysis.PreprocessedVirtualFileFactory
- [X]  arrow.meta.phases.analysis.ExtraImports
- [X]  arrow.meta.phases.config.Config
- [X]  arrow.meta.phases.analysis.CollectAdditionalSources


i-walker:
Hi, @Starmel thanks for contributing. 
There are some unwritten conventions we have in the Ide and some more general ones in Arrow: 
- All imports are explicit
- The prefix `Ide` in the PR here to each of the Phases may add confusion because those phases don't exist in the Ide. Instead they ley as a bridge to the compiler from the Ide API's. If you have more questions on that you can ping me on Slack in the #arrow-meta channel. 

So feel free to name them the same, the FQName will be the determining factor, whether users are targeting the compiler or Ide.

Other than that looks very good :D 👍

Starmel:
@i-walker, thanks for the answer!

For me, the practice of writing full imports for types with the same name is not the most common and can be overwhelming. If the goal for these classes is to be a bridge between IDE and the core of the library, then I think it would be quite normal to call it `IdeBridgeSyntheticScopeProvider`.

What do you think about this? If the naming strategy has already been adopted, then I am ready to correct the naming, as it was before.

If they need to be named the same, then can I use import aliases to make it easier to read inside the library? Something like that:

```kotlin
import arrow.meta.phases.resolve.synthetics.SyntheticResolver as SyntheticResolverGeneralPhase
import com.intellij.openapi.project.Project

interface SyntheticResolver : ExtensionPhase {
  fun syntheticResolver(project: Project): SyntheticResolverGeneralPhase?
}
```

```kotlin
import arrow.meta.phases.resolve.synthetics.SyntheticResolver as SyntheticResolverGeneralPhase

interface IntegrationSyntax {

  fun MetaIde.syntheticResolver(
    f: (Project) -> SyntheticResolverGeneralPhase?
  ): ExtensionPhase =
    object : SyntheticResolver {
      override fun syntheticResolver(project: Project): SyntheticResolverGeneralPhase? = f(project)
    }
}

**Comment by i-walker** [comment]
Using the FqName as in the original example is not an issue :) 

They would collide with other naming convention, we prefixed with Ide, that's why it's not preferred.

**Comment by Starmel** [comment]
Updated :)

File: idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/integration/IntegrationSyntax.kt [hunk 1]
```diff
@@ -16,20 +27,96 @@ import org.jetbrains.kotlin.types.KotlinType
...
+
   fun MetaIde.syntheticResolver(
-    f: (Project) -> arrow.meta.phases.resolve.synthetics.SyntheticResolver?
+    f: (Project) -> SyntheticResolver?
   ): ExtensionPhase =
-    object : SyntheticResolver {
-      override fun syntheticResolver(project: Project): arrow.meta.phases.resolve.synthetics.SyntheticResolver? =
-        f(project)
+    object : arrow.meta.ide.phases.integration.SyntheticResolver {
+      override fun syntheticResolver(project: Project): SyntheticResolver? = f(project)
```
i-walker, line 35:
```suggestion
      override fun syntheticResolver(project: Project): SyntheticResolver? =
         f(project)
```
Nit: does this format look good to you?

idea-plugin/src/main/kotlin/arrow/meta/ide/phases/integration/integration.kt
```diff
@@ -1,14 +1,56 @@
...
-import arrow.meta.phases.resolve.synthetics.SyntheticResolver
 import com.intellij.openapi.project.Project
 
 interface SyntheticResolver : ExtensionPhase {
-  fun syntheticResolver(project: Project): SyntheticResolver?
+  fun syntheticResolver(project: Project): arrow.meta.phases.resolve.synthetics.SyntheticResolver?
 }
 
 interface PackageProvider : ExtensionPhase {
-  fun packageFragmentProvider(project: Project): PackageProvider?
```
i-walker, line 13: I think Kotlin can resolve that for instance, `SyntheticResolver` is meant to be `arrow.meta.phases.resolve.synthetics.SyntheticResolver`. 
But if it didn't work that's fine here, too. 

:D looks good

Starmel: If I understand correctly, it will look better, but it may not be at all obvious that a different type is returned here, and not the one that was just declared.

idea-plugin/src/main/kotlin/arrow/meta/ide/dsl/integration/IntegrationSyntax.kt
Starmel: ¯\\_(ツ)_/¯ 

think, since the same operation is performed here in the entire file in the body of the method, you can not transfer

PR #731, opened 2020-08-22
Fixes GivenTests

i-walker:
Resolves the GivenProof resolution, which caused the failing GivenTests.

Prior to 1.3.71 there was a member `descriptor`, which would return the resolved call at the call-site. It has been removed ever since in this [PR](https://github.com/JetBrains/kotlin/commit/ed79ab68abb0969c725e2c0b7d377866f41ac070#diff-5d9c42c2e76ac014d97644460546e5af). 
This PR resolves the Kotlintype for type parameters so that calculating the upper bound resolves with the concrete type at call-site and not the generic type parameter, which is found in the ValueparameterDescriptor.

### DIFF
--- a/compiler-plugin/src/main/kotlin/arrow/meta/log/Log.kt
+++ b/compiler-plugin/src/main/kotlin/arrow/meta/log/Log.kt
@@ -10,9 +10,9 @@ sealed class Log {
 operator fun <A> Log.invoke(
   tag: A.() -> String,
   f: () -> A
-): A = f()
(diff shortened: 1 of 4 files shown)


rachelcarmena:
Awesome, fixed!! All the IDE Plugin tests are passing now because the pending test was using Given (Gradle test with typeproofs playground) :raised_hands: Merging because target branch is related to #706 . Thank you so much!!!!!

PR #735, opened 2020-08-28
Configuration: use of the new publication workflow

rachelcarmena:

PR #736, opened 2020-08-28
Revert "Configuration: use of the new publication workflow"

rachelcarmena:
Reverts arrow-kt/arrow-meta#735


rachelcarmena:
It reverts the last change and adds the old publish configuration to make it work because the migration implies more changes.

PR #740, opened 2020-09-07
Remove maven plugin and improve dependencies

rachelcarmena:
- The same plugin will be used to publish into local repository and an external repository
- The new configurations can be used
- Use of the same publication configuration than Arrow now
- Remove the use of the old configurations
- The publication of the shadowed jar is added automatically now (no extra conf needed for it)
- Upgrade of Gradle
- Add a warning about the versions of Intellij IDEA
- Reproduce a real private repository for IDEA plugins
- Check Kotlin version in use with the version used by Arrow Meta Gradle Plugin


idea-plugin/build.gradle
```diff
@@ -22,9 +22,10 @@ dependencies {
   implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$KOTLIN_VERSION"
   implementation "org.assertj:assertj-core:$ASSERTJ_VERSION"
   implementation "org.celtric.kotlin:kotlin-html:$KOTLIN_HTML_VERSION"
-  api project(path: ':compiler-plugin', configuration: 'shadow')
+  compileOnly project(path: ':compiler-plugin', configuration: 'shadow')
```
rachelcarmena, line 25: `shadow` plugin doesn't add `classifier` to the dependency in the POM file with `maven-publish` plugin so that's the reason why the dependency is added as `compileOnly` in order not to appear in the wrong way

PR #743, opened 2020-09-22
Fix publish and run IDE

rachelcarmena:
This PR continues https://github.com/arrow-kt/arrow-meta/pull/740 in order to use the same artifacts that are published into repositories for local tests.


danimontoya (approved):
Thanks @rachelcarmena the error about `InternalRegistry` is gone!

raulraja (approved):

i-walker (approved):

PR #744, opened 2020-09-22
IDE plugin: remove additional libraries

rachelcarmena:
The libraries that are added to fix the errors that are raised by the Intellij Plugin Verifier impact on Arrow Meta IDE Plugin (errors about incompatibilities and missing classes)


danimontoya (approved):
thanks @rachelcarmena !! 🙌

PR #752, opened 2020-09-23
Replace testing framework

rachelcarmena:
It replaces KotlinTest by JUnit because nothing related to KotlinTest is being used, just annotations and engine from JUnit.


i-walker (approved):
🙌  Thanks a lot @rachelcarmena !

danimontoya (approved):
Thanks @rachelcarmena :love47: 🚀

PR #769, opened 2020-10-13
Fix some minor errors in IDE

i-walker:


compiler-plugin/src/main/kotlin/arrow/meta/phases/codegen/ir/IrUtils.kt
```diff
@@ -211,6 +212,21 @@ val IrCall.unsubstitutedDescriptor: FunctionDescriptor
 val IrCall.substitutedValueParameters: List<Pair<ValueParameterDescriptor, KotlinType>>
   get() = unsubstitutedDescriptor.substitutedValueParameters(this)
 
+
+val IrTypeParametersContainer.allTypeParameters: List<IrTypeParameter>
+  get() = if (this is IrConstructor)
```
danimontoya, line 217: should we use `safeAs` instead?

compiler-plugin/src/test/kotlin/arrow/meta/plugins/proofs/ResolutionTests.kt
```diff
@@ -51,8 +52,8 @@ class ResolutionTests {
       @Coercion
       internal fun String.toInt16(): Int? =
         toIntOrNull(16)
-        
-      val x: Int? = "30"
+  
+      val s = "30"
```
danimontoya, line 56: I see this test is disabled.. but renaming `x` to `s` wont work since below is checking for the `x`

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/folding/tuples.kt
```diff
@@ -22,14 +22,14 @@ val IdeMetaPlugin.codeFoldingOnTuples: ExtensionPhase
 
 fun KtTypeReference.tupleTypeMatches(): Boolean =
   getType().isTypeMatching() &&
-    strictParents().all { psiElement ->
+    parents.all { psiElement ->
```
danimontoya, line 25: Did you check if this passed the tests?

danimontoya (approved):
Thank youu @i-walker 🙌

compiler-plugin/src/main/kotlin/arrow/meta/phases/codegen/ir/IrUtils.kt
i-walker: that won't necessary change the logic

idea-plugin/src/main/kotlin/arrow/meta/ide/plugins/proofs/folding/tuples.kt
i-walker: yes locally

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.6.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.5.1-all.zip
```
rachelcarmena, line 3: @danimontoya this version is for Arrow Meta. It's not involved in this case.

There are 2 versions that should be the same:
* That version in that file in `arrow-typeproofs` repository
* `BASE_GRADLE_VERSION` in this repository which is used to run the test in `arrow-typeproofs` repository. that's the reason why they must be the same version.

rachelcarmena, line 3: @danimontoya this version is used by Arrow Meta. It's not involved in this case.

There are 2 versions that should have the same value:
* The version in that file in `arrow-typeproofs` repository
* `BASE_GRADLE_VERSION` in this repository which is used to run the test on `arrow-typeproofs` repository. That's the reason why they must be the same version.

idea-plugin/src/test/kotlin/arrow/meta/ide/plugins/proofs/repo/GradleTypeproofsTest.kt
```diff
@@ -17,8 +17,8 @@ class GradleTypeproofsTest : GradleTestSetUp() {
       val arrowProjectDir = myProjectRoot.findChild("arrow-typeproofs")
       myProjectRoot = arrowProjectDir
 
-      // val checkoutResult = gitCheckout(myProject, "<other-branch>", myProjectRoot)
-      // assertTrue(checkoutResult.success())
+       val checkoutResult = gitCheckout(myProject, "is-fix-ide-errors", myProjectRoot)
```
rachelcarmena, line 20: This branch should exist in `arrow-typeproofs` repository.

The test download that repository in the previous statements.

gradle/wrapper/gradle-wrapper.properties
danimontoya: I thought so.. I did it because just to give another try. Tried already with BASE_GRADLE_VERSION and gradle version from typeproofs being the same (6.5.1) and didn't work
I will revert that next! Thanks!! 😄

idea-plugin/src/test/java/org/jetbrains/plugins/gradle/importing/GradleImportingTestCase.java
```diff
@@ -71,7 +71,8 @@
 @RunWith(value = Parameterized.class)
 public abstract class GradleImportingTestCase extends ExternalSystemImportingTestCase {
     // patched for Arrow Meta:
-    public static final String BASE_GRADLE_VERSION = "5.6.4";
+    public static final String BASE_GRADLE_VERSION = "6.5.1";
```
i-walker, line 74: we should either stick with an Array or one version but not both

rachelcarmena (approved):
:tada:

PR #777, opened 2020-10-19
compiler-plugin: fix paths when transforming sources

rachelcarmena:
## Change

Remove manipulation of paths because it removes the first file separator for absolute paths.

## Reason

It's not possible to use absolute paths when transforming sources now, just relative paths: https://github.com/arrow-kt/arrow-meta/blob/master/compiler-plugin/src/test/kotlin/arrow/meta/quotes/transform/plugins/TransformNewSourcePlugin.kt#L109


raulraja (approved):

raulraja:
Thanks @rachelcarmena !

danimontoya (approved):
thanks @rachelcarmena !! 🙌

PR #778, opened 2020-10-21
[Check-Draft] Test with same gradle version of type-proofs

danimontoya:


danimontoya:
Closing draft since tests are passing now in #769

PR #779, opened 2020-10-23
Fix #591 - Path for generated source code

### LINKED ISSUE #591: [BUG] Wrong path for generated files

rachelcarmena:
## Proposal

* Absolute paths for `filePath` are not considered
* Two directories for generated sources:
  * Base directory (optional): provided via plugin option - it's provided automatically with Arrow Meta Gradle Plugin
  * Specific directory for a file (optional): provided via `filePath` in `file` - considered as a relative path

## Changes

* When generating source code:
   * Add default directory when `filePath` is not provided
   * Consider the base directory provided in `generatedSrcOutputDir` plugin option
* Remove the support of absolute paths (previous patch)
* Remove dead code
* Remove testing considerations from production code

## Pending changes

- [x] Fix [this comparison](https://github.com/arrow-kt/arrow-meta/compare/fix-591?expand=1#diff-f6b376f1922ccc8723495b30a42028ae7440d96b37f716f3dcc9f1d72f6f6918R390) because it's just related with testing
- [x] Use of plugin options during testing
- [x] Add tests with the use of plugin options
- [x] Arrow Meta Gradle Plugin - add the plugin option with the buildDir of the project by default
- [x] Previous review

## Screenshots

From an example that will be added in `arrow-meta-examples`:

Screenshott

## Notes

* Example about how to provide the plugin option (it will be documented in `arrow-meta-examples`)

```
compileKotlin {
    kotlinOptions {
        jvmTarget = "$JVM_TARGET_VERSION"
        freeCompilerArgs = [
                "-Xplugin=${project.rootDir}/create-plugin/build/libs/create-plugin-all.jar"
                ,"-P", "plugin:arrow.meta.plugin.compiler:generatedSrcOutputDir=${buildDir}"
        ]
    }
}
```


rachelcarmena:
Ready for review!! :raised_hands:

raulraja (approved):
Awesome, thank you @rachelcarmena !🙌🙌🙌

compiler-plugin/src/main/kotlin/arrow/meta/quotes/Quote.kt
```diff
@@ -386,9 +387,10 @@ fun ArrayList<KtFile>.replaceFiles(file: KtFile, newFile: List<KtFile>) {
 
 fun CompilerContext.changeSource(file: KtFile, newSource: String, rootFile: KtFile, sourcePath: String? = null): KtFile {
   var virtualFile = rootFile.virtualFile
-  if (file.name != DEFAULT_META_FILE_NAME) {
-    val path = sourcePath ?: System.getProperty("arrow.meta.generated.source.output", DEFAULT_SOURCE_PATH)
-    val directory = Paths.get(path).toFile()
+  if (sourcePath != null) {
+    val baseDir = configuration?.get(ArrowMetaConfigurationKeys.GENERATED_SRC_OUTPUT_DIR, listOf(DEFAULT_BASE_DIR.toString()))?.get(0)
+    val path = Paths.get(baseDir, sourcePath)
```
danimontoya, line 392: `baseDir` here is nullable, not sure if it can cause any problems.. maybe we could wrap it with `?.let`

danimontoya (approved):
Looks good to me! only a minor nit! thank you @rachelcarmena 🙌

compiler-plugin/src/main/kotlin/arrow/meta/quotes/Quote.kt
rachelcarmena: Good catch!! Thank you so much @danimontoya :raised_hands:

i-walker (approved):
Thanks. @rachelcarmena 🙌

rachelcarmena:
Thank you so much @raulraja @danimontoya @i-walker :raised_hands: :raised_hands: :raised_hands:

Issue #749, opened 2020-09-22
Kinds Test Error


i-walker:
Since this [PR](https://github.com/JetBrains/kotlin/commit/ed79ab68abb0969c725e2c0b7d377866f41ac070#diff-5d9c42c2e76ac014d97644460546e5af) descriptors at runtime provide solely an unsubstituted version of their type, receiver, and dispatch information, thus making the current approach [to resolve the value and target Type](https://github.com/arrow-kt/arrow-meta/blob/80df7896f5c16e3bee81788930e87aa08c453160/compiler-plugin/src/main/kotlin/arrow/meta/plugins/proofs/phases/ir/ProofsIrCodegen.kt#L167-L173) incorrect, as a type, receiver, and dispatch substitution is needed. 

To substitute type information one may use `org.jetbrains.kotlin.ir.types.IrTypeSubstitutor`.

PR #780, opened 2020-10-28
IrTypeparameter resolution

i-walker:
addresses #749, as in adding new members and utilities to resolve IrTypeParameters through the IrTree and recursively walking the parent hierarchy for potentially unsubstituted type parameters. 

Conceptually, it enables the functionality found in `IrTypeSubstituter`, which is specialized for `IrValueParameters` for `IrMemberAccessExpressions`, by breaking down its API and behavioral aspects. 

In a further PR, this can be used to refine the `targetType`, when extension proofs are injected, that have missing type information. But the technical debt persists to improve the detection of that very `targetType` consistently for all expressions.  

An example is here - given that List has an Functor instance:

```
call-site: 
listOf(2).fmap{ it * 2 }
// in this case the call-site is broken down by each "call expression"
```

expression | unsubstituted type | missing type information | resolved Type
--- | --- | --- | ---
listOf(2) | `List<T>` | `T` | `Int`
fmap { it * 2} | `B from fmap` | `B` |  `Int`

To further resolve `fmap` the remaining Type parameters in the parent hierarchy need to be resolved of `Functor` and `Functor.Ops`, which is:

expression | unsubstituted type | missing type information | resolved Type
--- | --- | --- | ---
fmap { it * 2}  | `Functor.Ops<F, A>` | `F`, `A` |  `F` -> `ListOf<Int>`, `A` -> `Int`
fmap { it * 2}  | `Functor<F>` | `F`|  `ListOf<Int>`

implying that `ListOf<A>` is an typealias for `Kind<`List(_)`, A>` and established as an isomorphism to `List<T>`

This is necessary, because the expected value- and target type for an extensionproof injection at the aforementioned call-site would be: 
- valueType: `List<Int>` or worst case: `List<T>` <- detection works
- targetType: `Functor<F, A>` <- which can now be refined to `Functor<ListOf<Int>, Int>`
But currently the target Type is not resolved to `Functor<F, A>`


compiler-plugin/src/main/kotlin/arrow/meta/phases/codegen/ir/IrUtils.kt
```diff
@@ -218,13 +232,44 @@ val IrTypeParametersContainer.allTypeParameters: List<IrTypeParameter>
   else
     typeParameters
 
-fun IrMemberAccessExpression.getTypeSubstitutionMap(container: IrTypeParametersContainer): Map<IrTypeParameter, IrType> =
-  container.allTypeParameters.withIndex().associate {
-    it.value to getTypeArgument(it.index)!!
-  }
+fun IrTypeParametersContainer.getTypeSubstitutionMap(expression: IrMemberAccessExpression): Map<IrTypeParameter, Either<KotlinType?, TypeProjection?>> =
```
i-walker, line 235: Here I was also keen on return Map<IrTypeParameter, Either<IrSimpleType, IrTypeArgument>>, depending on how we use it in other places, besides Extensions. 
But I can change it to one or the other, it's not fixed.

danimontoya (approved):
thank you @i-walker looks good! 🙌

raulraja (approved):
👏 thanks @i-walker !

PR #784, opened 2020-11-15
Optics meta plugin 

raulraja:
Port of the optics processor originally by @nomisRev from arrow-core to arrow-meta. One down to remove kapt.


compiler-plugin/src/test/kotlin/arrow/meta/plugins/optics/DSLTests.kt
```diff
@@ -0,0 +1,21 @@
...
+  fun `DSL is generated for complex model`() {
+    """
+      |$imports
+      |$dslModel
+      |val john = Employee("Audrey Tang",
+      |       Company("Arrow",
+      |               Address("Functional city",
+      |                       Street(42, "lambda street"))))
+      |val modify = Employee.company.address.street.name.modify(john, String::toUpperCase)
+      |val r = modify.company?.address?.street?.name
```
nomisRev, line 17:
```suggestion
      |val r = modify.company.address.street.name
```

These are not nullable in `dslModel`, we can make them `nullable` since that should also be supported. Those should then be generated as `Optional`, and still, be composable.

nomisRev (approved):
Looks good to me! Nice work, thanks @raulraja for taking care of this 😘

compiler-plugin/src/test/kotlin/arrow/meta/plugins/optics/DSLTests.kt
raulraja: In this case, I'm accessing an instance of `john` which is poorly named as `modify`

rachelcarmena (approved):
:tada:

PR #795, opened 2021-03-17
Fix #794 Set Java env

### LINKED ISSUE #794: [BUG] compiler-plugin java class files are on version 55
**Describe the bug**
It looks like the latest snapshot published was compiled with JDK 11 and all the Kotlin source files got compiled for compatibility with Java 1.8, but the *Java* source files wound up being Java 11 format. 

```
❯ javap -classpath `pwd`/compiler-plugin-1.4.10-SNAPSHOT.jar -verbose arrow/meta/diagnostic/MetaDefaultErrorMessages  | grep major
  major version: 55
```

I know the actions explicitly set JAVA_HOME to Java 8, but perhaps the latest snapshot was published without that. The other potential fix would just be to update targetCompatibility for java in compiler-plugin/build.gradle

```
java {
    targetCompatibility = JavaVersion.VERSION_1_8
}
```

rachelcarmena:
Reason: the default JDK for GH environments has changed.

### DIFF
--- a/.github/workflows/build_artifacts.yml
+++ b/.github/workflows/build_artifacts.yml
@@ -11,13 +11,15 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Set env
+        run: |
+          echo "$JAVA_HOME_8_X64/bin" >> $GITHUB_PATH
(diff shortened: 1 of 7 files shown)


raulraja (approved):
thanks @rachelcarmena !

PR #797, opened 2021-03-30
try fixing casting error

i-walker:
There is a type-inference issue when requesting 3rd Party overwritten service - which is then casted to their service interface here `MetaIde`
registering the coherently resolved Plugin, which is either the one Upstream or the one overwriting the Service.
TODO: If this is fixed Upstream we can change that temporary fix.


raulraja (approved):
thanks @i-walker !

rachelcarmena (approved):

i-walker:
RIP Ide 💚

PR #801, opened 2021-04-17
remove ide plugin

i-walker:


raulraja (approved):
Thank you @i-walker and RIP IDE plugin for now 🙏 . Let's wait for @rachelcarmena before merging, and I think she is in the middle of adapting to 1.5. Thanks!

danimontoya (approved):
thanks @i-walker ! Indeed RIP ide plugins!

.github/workflows/publish_artifacts.yml
```diff
@@ -1,120 +0,0 @@
-name: Publish Artifacts
```
rachelcarmena, line 1: Thanks @i-walker !! This file must exist. I'll add a new commit in a few minutes!

i-walker: Thanks @rachelcarmena 😄

rachelcarmena (commented):
@i-walker :raised_hands: 

Please, take a look at the changes I added:

- Recovering artifacts publication
- Removing shadow jar creation
- Improving compiler plugin dependencies
- Removing Arrow Meta IDE plugin installation from Gradle Plugin
- Adding transitive deps from compiler plugin in the Gradle Plugin instead of explicit deps.
- Removing the configuration of doc validation because there aren't compile deps from Intellij to add in runtime.
- Update documentation (sidebar, setup, ...)
- Other small configuration changes

Please, I think these files could be updated as well:

* compiler-plugin/src/main/kotlin/arrow/meta/dsl/platform/Platform.kt
* compiler-plugin/src/main/kotlin/arrow/meta/dsl/resolve/ResolveSyntax.kt
* compiler-plugin/src/main/kotlin/arrow/meta/internal/registry/InternalRegistry.kt
* compiler-plugin/src/main/kotlin/arrow/meta/internal/kastree/ast/psi/Parser.kt

and this file could be removed, right?

* compiler-plugin/src/main/resources/META-INF/services/org.jetbrains.kotlin.diagnostics.rendering.DefaultErrorMessages$Extension

Thanks!!! :raised_hands:

gradle-plugin/build.gradle
```diff
@@ -21,7 +21,6 @@ dependencies {
   implementation "org.jetbrains.kotlin:kotlin-gradle-plugin-api:$KOTLIN_VERSION"
   implementation "org.jetbrains.kotlin:kotlin-gradle-plugin:$KOTLIN_VERSION"
   implementation "io.github.classgraph:classgraph:$CLASS_GRAPH_VERSION"
-  implementation "xerces:xercesImpl:$XERCES_VERSION"
```
i-walker, line 24: Just a question for what did we use this @rachelcarmena?

gradle-plugin/src/main/kotlin/arrow/meta/plugin/gradle/ArrowGradlePlugin.kt
```diff
@@ -40,21 +38,6 @@ class ArrowGradlePlugin : Plugin<Project> {
         )
       }
     }
-    project.tasks.register("install-idea-plugin", InstallIdeaPlugin::class.java) {
```
i-walker, line 43: 🙌🏾

i-walker:
Thanks a lot @rachelcarmena  🙌🏾 I am going to push some additional changes from what you suggested.

i-walker:
On some of the changes I haven't changed/ updated:
- compiler-plugin/src/main/kotlin/arrow/meta/dsl/resolve/ResolveSyntax.kt
IMO those are fine bc `main` still depends on the Descriptor API's in 1.4.10. 
- compiler-plugin/src/main/kotlin/arrow/meta/internal/registry/InternalRegistry.kt
Haven't found something in InternalRegistry that would relate to the Ide. 
- compiler-plugin/src/main/kotlin/arrow/meta/dsl/platform/Platform.kt
Here and all usages of the `ide` function are required, so the compiler plugin interacts well with the IDE base APIs for compiler plugins. Unless we don't want to open projects in the IDE built with Meta.
- compiler-plugin/src/main/kotlin/arrow/meta/internal/kastree/ast/psi/Parser.kt
So true, we haven't used this part of the Parser at all 😄 
- compiler-plugin/src/main/resources/META-INF/services/org.jetbrains.kotlin.diagnostics.rendering.DefaultErrorMessages$Extension
This one is also needed so it registers our Diagnostics in the compiler. It existed in the Ide because that has a symmetric extension/ api. 

Thanks for the feedback review @rachelcarmena , when I oversaw something let me know 🙌🏾

gradle-plugin/build.gradle
rachelcarmena: To parse the XML file of the private IDEA plugins repository: https://github.com/arrow-kt/arrow-meta/blob/main/gradle-plugin/src/main/kotlin/arrow/meta/plugin/gradle/InstallIdeaPlugin.kt#L50

rachelcarmena (approved):
Awesome, thanks @i-walker !!! :tada:

PR #812, opened 2021-05-08
CU-jfaprn Support composition of remote and local refined types predicates.

raulraja:
In order to evaluate at compile-time constant expressions from predicates that are derived from remote jars + local user declarations, we need to account for that in the codegen of auxiliary functions that are evaluated at compile time by the IrInterpreter.
For example consider the following declarations:
```kotlin
value class PositiveIntEven private constructor(val value: Int) {
  companion object : Refined<Int, PositiveIntEven>(::PositiveIntEven, PositiveInt, Even)
}

value class Eighteen private constructor(val value: Int) {
  companion object
    : Refined<Int, Eighteen>(::Eighteen, {
    ensure((it == 18) to "expected 18")
  })
}

value class PositiveIntEven18 private constructor(val value: Int) {
  companion object : Refined<Int, PositiveIntEven18>(::PositiveIntEven18, PositiveIntEven, Eighteen)
}
```
`PositiveIntEven`(local) == PositiveInt(remote) + Even(remote)
`PositiveIntEven18`(local) == PositiveIntEven (local) + Eighteen (local)

This matters because remote and local predicates need to be addressed in a different way by the `IrInterpreter`. While remote predicates can be loaded and evaluated with reflection, local predicates are interpreted.


raulraja:
Looks like I need to update my branch based on the latest naming changes.

rachelcarmena (approved):
Awesome!! :rocket:

PR #817, opened 2021-05-21
Update multiplatform configuration

rachelcarmena:
- Add `js` target
- Update dependencies to make it work for both targets (IDE included)
- Small fix in tests to make them run for `js` target as well
- Remove `kotlin.stdlib.default.dependency=false` because of https://youtrack.jetbrains.com/issue/KT-46847
- Fix the application of Kotlin JVM plugin
- Extract and relocate code to be able to customize `clean` task

### DIFF
--- a/arrow-refined-types/build.gradle
+++ b/arrow-refined-types/build.gradle
@@ -12,12 +12,9 @@ kotlin {
         commonTest {
             dependencies {
                 compileOnly "org.jetbrains.kotlin:kotlin-stdlib:$KOTLIN_VERSION"
-                implementation "io.kotest:kotest-runner-junit5:$KOTEST_VERSION"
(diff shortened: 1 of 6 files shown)


arrow-refined-types/build.gradle
```diff
@@ -43,6 +40,10 @@ kotlin {
             }
         }
     }
+
+    js {
+        nodejs()
```
raulraja, line 45: should we enable the browser too or it does not matter?

raulraja (approved):
thanks @rachelcarmena !

arrow-refined-types/build.gradle
rachelcarmena: Of course, done, thanks!! :raised_hands:

PR #819, opened 2021-06-01
Update MPP configuration

rachelcarmena:
To recover the use of `kotlin.stdlib.default.dependency` property


raulraja (approved):
thanks @rachelcarmena !

PR #827, opened 2021-07-02
@Context meta annotations

raulraja:
Removes all proofs except for `@Context` injection.
To implement what we have currently in Arrow a library or user would declare

```kotlin
import arrow.Context
    
@Context
@Retention(AnnotationRetention.RUNTIME)
@Target(
      AnnotationTarget.CLASS,
      AnnotationTarget.FUNCTION,
      AnnotationTarget.PROPERTY,
      AnnotationTarget.VALUE_PARAMETER
 )
 annotation class Given

 inline fun <A> given(@Given identity: A): A = identity
```

Context val, fun, object and class are accepted as providers.

for any function that declares `@Given` or any other contextual annotation args such as
```kotlin
import arrow.Given

fun <A> given(@Given ev: A): A
```
an internal corresponding representation is generated for frontend resolution and usable without passing the arguments
```kotlin
@CompileTime
internal fun <A> given(ev: A = TODO("Compile time replaced"), unit: Unit = Unit): A
```
that can be used without args

```kotlin
given<Type>()
```
Once all `@Context` requirements are resolved and replaced in the local module the internal `@CompileTime` functions are removed in the IR phase and their calls replaced by the injected values.

Includes new features in the validation of call sites checking for ambiguous providers, automatic derivation, and other improvements to the previous state of the proofs plugin.


arrow-meta/src/main/kotlin/arrow/meta/plugins/proofs/phases/ProofsExt.kt
```diff
@@ -0,0 +1,63 @@
...
+  }.toSet()
+
+fun DeclarationDescriptor.asProof(): Sequence<Proof> =
+  when (this) {
+    is PropertyDescriptor -> asProof()
+    is ClassConstructorDescriptor -> containingDeclaration.asProof()
+    is FunctionDescriptor -> asProof()
+    is ClassDescriptor -> asProof()
+    is FakeCallableDescriptorForObject -> classDescriptor.asProof()
+    else -> TODO("asProof: Unsupported proof declaration type: $this")
```
nomisRev, line 28: Can we provide more information for the user here?
Ideally we provide the code location, line-number with file.

Same for other in this file, but can be moved to ticket/seperate PR.

proofs-plugin/src/main/java/arrow/meta/diagnostic/MetaDefaultErrorMessages.java
```diff
@@ -27,13 +25,12 @@ public DiagnosticFactoryToRendererMap getMap() {
         );
         MAP.put(AmbiguousProof,
                 "This {0} has following conflicting proof/s: {1}.\nPlease disambiguate resolution, by either declaring only one internal orphan / public proof over the desired type/s or remove conflicting proofs from the project.", RenderProof, RenderProofs);
+        MAP.put(AmbiguousProofForSupertype,
+                "Found ambiguous proofs for type {0}. Proofs : {2}", RenderTypes, RenderProof, RenderProofs);
         MAP.put(OwnershipViolatedProof,
                 "This {0} violates ownership rules, because public Proofs over 3rd party Types break coherence over the kotlin ecosystem. One way to solve this is to declare the Proof as an internal orphan.", RenderProof);
         MAP.put(UnresolvedGivenProof, "This GivenProof on the type {0} can't be semi-inductively resolved. Please verify that all parameters have default value or that other injected given values have a corresponding proof.", RenderTypes);
+        MAP.put(CycleOnGivenProof, "This GivenProof on the type {0} has cyclic dependencies: {1}. Please verify that proofs don't depend on each other for resolution.", RenderTypes, RenderProofs);
```
nomisRev, line 33: These messages are not autoshown in the IDEA, are they?

proofs-plugin/src/main/kotlin/arrow/meta/plugins/proofs/phases/quotes/GenerateGivenSupportingFunctions.kt
```diff
@@ -3,88 +3,136 @@ package arrow.meta.plugins.proofs.phases.quotes
...
+private fun List<ValueParameterDescriptor>.renderAsArguments(): String =
+  joinToString { it.name.asString() }
 
-private fun KtTypeParameter.containsGivenConstrain(): Boolean =
-  extendsBound?.annotationEntries?.any { it.text.matches(givenAnnotation) } ?: false
\ No newline at end of file
+private tailrec fun ModuleDescriptor.declarationsWithGivenArguments(
+  acc: List<DeclarationDescriptor> = emptyList(),
+  packages: List<FqName> = listOf(FqName.ROOT),
+  skipPacks: Set<FqName> = setOf(
```
nomisRev, line 107: This is an optmisation technique? Since these packages can never contain any of our annotations?

proofs-plugin/src/main/kotlin/arrow/meta/plugins/proofs/phases/resolve/cache/ProofsCache.kt
```diff
@@ -35,24 +31,42 @@ fun ModuleDescriptor.initializeProofCache(ctx: CompilerContext): List<Proof> =
     emptyList()
   }
 
-private tailrec fun ModuleDescriptor.computeModuleProofs(
-  acc: Sequence<Proof>,
-  packages: List<FqName>
-): Sequence<Proof> =
+private val skipPackages = setOf(
```
nomisRev, line 34: This is defined twice.

proofs-plugin/src/test/kotlin/arrow/meta/plugins/proofs/GivenTest.kt
```diff
@@ -34,7 +34,7 @@ class GivenTest {
     givenTest(
       source = """
         @Given internal val x = "yes!"
-        fun id(evidence: String = given()): String =
+        fun id(@Given evidence: String): String =
```
nomisRev, line 37: 🔥

nomisRev (approved):
Awesome work! Can't wait to take this for a spin!!

proofs-plugin/src/main/java/arrow/meta/diagnostic/MetaDefaultErrorMessages.java
raulraja: I believe they will not show up immediately as other analysis errors do but will be reported as errors with the right code location when a full build is performed. Same as codegen and other plugins that have no IDEA support yet until FIR compiler plugins are a thing.

proofs-plugin/src/main/kotlin/arrow/meta/plugins/proofs/phases/quotes/GenerateGivenSupportingFunctions.kt
raulraja: Yes, there is currently no automatic way in the compiler to scan the classpath for annotated descriptors, and this ensures we don't recurse down paths where we know there are no proofs or we explicitly want to ignore any proofs people are trying to project over core packages

arrow-meta/src/main/kotlin/arrow/meta/plugins/proofs/phases/ProofsExt.kt
raulraja: I think at some point, this got refactored to become a `Sequence<Proof>`, and the `else` case should have instead returned an empty sequence since the cases above are all the kind of declarations we currently support as proofs. I'll change it to return an emptySequence

PR #829, opened 2021-07-11
Add toEither toValidated to refined types

lidonis:
The goal is to be able to work with arrow type seamlessly


raulraja (approved):
thanks @lidonis !

PR #832, opened 2021-07-16
Fix generation path in given internal supported functions

raulraja:
Co-authored with @JavierSegoviaCordoba


nomisRev (approved):
LGTM 👏 👏

i-walker (approved):
🙌🏾

JavierSegoviaCordoba:
The path generated is invalid in windows because it takes the drive, and the suffix is `build`, i.e. `build/C:\someFolder`. Then `:` is invalid.

I fixed it with this one, but probably this should be reviewed in the future.

```
fun CompilerContext.getOrCreateBaseDirectory(parentPathFile: File?): File {
  val baseDir = configuration?.get(ArrowMetaConfigurationKeys.GENERATED_SRC_OUTPUT_DIR, listOf(DefaultElementScope.DEFAULT_BASE_DIR.toString()))?.get(0)
    ?: DefaultElementScope.DEFAULT_BASE_DIR.toString()
  val parentPath = Paths.get(parentPathFile?.path ?: "")
  val parentPathWithoutRoot = parentPath.root.relativize(parentPath)
  val path = Paths.get(baseDir, parentPathWithoutRoot.toString())
  val directory = path.toFile()
  directory.mkdirs()
  return directory
}
```

raulraja:
Thanks @JavierSegoviaCordoba , I thinks this is a lot better than what we had 👏 .

PR #833, opened 2021-07-16
Upgrade to Arrow 0.13.2, and update optics plugin

nomisRev:


i-walker:
let me fix the proof resolution test. It fails, bc Eq doesn't exist anymore. I'll replace it with a different example

i-walker (approved):
Thanks @nomisRev 👍🏾

nomisRev:
Thanks for fixing the resolution test @i-walker ❤️

PR #834, opened 2021-08-06
Change the workflow to generate and publish documentation site

franciscodr:
We want to publish the Arrow Meta documentation in the same location as other Arrow libraries like core or fx. To do that, we need to update the workflows that generate the documentation to indicate the new folder and base URL, that is, `doc/meta`.

After this change, the URL for the Arrow Meta documentation will be https://arrow-kt.io/docs/meta instead of https://meta.arrow-kt.io.


raulraja (approved):
Thanks @franciscodr !

PR #836, opened 2021-08-18
Remove tuples and synthetic from prelude

raulraja:


nomisRev (approved):
Thanks for the cleaning up!

PR #837, opened 2021-08-18
POC implementation of checking via SMT

serras:
This is the first step towards an Arrow Meta plug-in which checks pre- and post-conditions of functions using SMT a SMT solver. Examples are available as part of the [testsuite](https://github.com/arrow-kt/arrow-meta/blob/rr-liquid-draft/refined-types-plugin/src/test/kotlin/arrow/meta/plugins/liquid/LiquidDataflowTests.kt).

### Current set of features

1. Several ways to describe the pre- and post-conditions, including a [DSL](https://github.com/arrow-kt/arrow-meta/blob/rr-liquid-draft/arrow-refined-types/src/commonMain/kotlin/arrow/refinement/RefinementDSL.kt) and attaching [laws](https://github.com/arrow-kt/arrow-meta/blob/rr-liquid-draft/arrow-refined-laws-std/src/main/kotlin/Test.kt) to existing functions. The information is gathered [in this way](https://github.com/arrow-kt/arrow-meta/blob/rr-liquid-draft/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintCollection.kt).
2. Those conditions turn into Java annotations for easier handling in later stages. This is done [in this way](https://github.com/arrow-kt/arrow-meta/blob/rr-liquid-draft/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/ir/ConstraintsAnnotations.kt), obtaining information like in [this example](https://github.com/arrow-kt/arrow-meta/blob/rr-liquid-draft/refined-types-plugin/src/test/kotlin/arrow/meta/plugins/liquid/LiquidDataflowTests.kt#L96).
3. These conditions are given as source to the [checking phase](https://github.com/arrow-kt/arrow-meta/blob/rr-liquid-draft/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintChecks.kt). We perform different types of checks:
    * Whether the pre-conditions in a function are consistent (so a precondition `x > 0 && x < 0` is rejected),
    * Whether post-conditions hold at the end of the function,
    * Whether the arguments to each call satisfy the pre-conditions of that call (rejecting things like potentially dividing by zero),
    * Warning about unreachable code given the information known at each point.

### Some technical points

1. Communication with the SMT solver is handled using the [`java-smt`](https://github.com/sosy-lab/java-smt) library. This library downloads the required dependencies automatically.
    - Our goal is to use the more powerful Z3, but we are having problems with the initialization. @raulraja is working on those, in the meanwhile we use the frictionless SMTInterpol.
2. The checking phase uses a [custom implementation of sequences](https://github.com/arrow-kt/arrow-meta/blob/rr-liquid-draft/refined-types-plugin/src/main/kotlin/arrow/meta/continuations/ContSeq.kt) by @nomisRev.
    - The main feature which sets it aside from Flow is the ability to `abort` a computation. This is used, for example, when a `return` statement is found, since at that point we need to stop checking anything else in the function.


raulraja (approved):
🚢

PR #841, opened 2021-08-18
Small improvements ContSeq

nomisRev:
Some improvemenets based on usage.
The `andThen` function seems less useful as I originally thought. Sorry @serras 😅

Some unrelated improvements. I turned `ContSeq` into a `sealed class` since it doesn't need to be open inheritance, and using a class allows for making the functions instance methods whilst still allowing for `inline`.
This results in requiring type over anonymous implementations but this is all local to the file.


raulraja (approved):
thanks @nomisRev !

refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintChecks.kt
```diff
@@ -303,53 +300,53 @@ private fun SolverState.checkCallArguments(
...
-      mayInteger != null ->
-        solver.ints {
-          equal(
-            solver.makeIntegerObjectVariable(associatedVarName),
-            makeNumber(mayInteger)
-          )
+    associatedVarName: String,
+    expression: KtConstantExpression
+): ContSeq<Unit> = cont {
+    solver.formulae {
```
nomisRev, line 306: What does this line do? I see we get a new `FormulaManager` here but we never seem to manipulate it. Does it create a new scope of some kind?

raulraja: It's just syntax for https://sosy-lab.github.io/java-smt/api/org/sosy_lab/java_smt/api/FormulaManager.html but it may not be in use in this scope.

nomisRev: Yes, it's not being used in this scope.

nomisRev: I was wondering if we could enhance the DSL to make these snippets a bit nicer with less nesting.

I.e. now we do `solver.boolean { not(booleanFormula) }` I was wondering if we couldn't add a `fun BooleanFormula.not(): BooleanFormula` inside `SolverState` since that seems to be the receiver everywhere.

That way we could eminate all the nesting everywhere, but if the nested is needed to create scopes than that's a different story.

serras: The problem is that you need either a `FormulaManager` or one of the `BooleanFormulaManager`, `NumeralFormulaManager`... to create new formulae. So it's not possible simply do `BooleanFormula.not()`, which would be the nicest DSL.

plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintChecks.kt
```diff
@@ -116,16 +110,15 @@ private fun SolverState.checkDeclarationWithBody(
...
-    val inconsistentPreconditions =
-      checkPreconditionsInconsistencies(constraints, context, declaration)
-    ensure<Unit>(!inconsistentPreconditions)
-    goOn()
-  }.flatMap {
-    checkExpressionConstraints(resultVarName, body, context, ReturnPoints(resultVarName, emptyMap()))
-  }.andThenSideEffect {
-    checkPostConditionsImplication(constraints, context, declaration)
-  }
+    continuationBracket.map {
```
serras, line 113: Shouldn't this be `flatMap`? Otherwise the call to `ensure` won't influence the flow.

serras: Got it! The `map` expects a `ContSyntax`, not a function.

refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintChecks.kt
nomisRev: You can hide that way in `SolverState`, no? This is what I had in mind.

```
class SolverState(...) {
   fun BooleanFormula.not(): BooleanFormula =
     booleans { not(this@not) }
}
```

Since it's defined inside `SolverState` it has access to `booleans { }`, right?

And since all the functions defined in `ConstraintChecker.kt` are defined as extensions on `SolverState` you would have that syntax available inside the method bodies.
It's similar to what soon would be solve-able with multiple context receivers.

It's just a suggestion because the nesting seems very annoying if it's not needed for wrapping code in scopes.

nomisRev:
@serras feel free to merge if happy with the changes 👍

PR #843, opened 2021-08-19
Support for mutable variables

serras:
This PR enables the analysis to work on `var` declarations and assignments. This is done by assigning fresh names to each new assignment of a variable; at a conceptual level the following code:

```kotlin
var x = 1
x = 2
return x
```

is "translated" into (but of course no translation takes place):

```kotlin
val x1 = 1
val x2 = 2
return x2
```

### DIFF
--- a/plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintCollection.kt
+++ b/plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintCollection.kt
@@ -311,6 +311,19 @@ private fun ResolvedCall<out CallableDescriptor>.preOrPostCall(): Boolean =
   preCall() || postCall()
 
 /**
+ * returns true if [this] resolved call is calling [arrow.refinement.invariant]
+ */
+internal fun ResolvedCall<out CallableDescriptor>.invariantCall(): Boolean =
+  resultingDescriptor.fqNameSafe == FqName("arrow.refinement.invariant")
+
+/**
+ * returns true if [this] resolved call is calling
+ * [arrow.refinement.pre] or [arrow.refinement.post] or [arrow.refinement.invariant]
+ */
+private fun ResolvedCall<out CallableDescriptor>.preOrPostOrInvariantCall(): Boolean =
+  preCall() || postCall() || invariantCall()
+
+/**
  * Translates a [resolvedCall] into an smt [BooleanFormula]
  * Ex.
  * ```kotlin
@@ -403,7 +416,7 @@ private fun Solver.argsFormulae(
   val results = arrayListOf<Formula>()
   val visitor = expressionRecursiveVisitor {
     val resolvedCall = it.getResolvedCall(bindingContext)
-    if (resolvedCall != null && !resolvedCall.preOrPostCall()) { // not match on the parent call
+    if (resolvedCall != null && !resolvedCall.preOrPostOrInvariantCall()) { // not match on the parent call
       val args = argsFormulae(resolvedCall, bindingContext)
       val descriptor = resolvedCall.resultingDescriptor
       val expressionFormula = formulaWithArgs(descriptor, args, resolvedCall)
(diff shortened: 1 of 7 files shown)


serras:
There's some work missing on respecting the invariants. For example, you can declare:

```kotlin
var x = invariant("greater than 0", { it > 0 }) { 1 }
x = 2
```

and then this is checked on every assignment to `x`.

Why is this important? Once we introduce loops, we cannot easily track which is the "latest" state of a variable. However, if the programmer declares an invariant, we can ensure that it holds. In that way we can make progress with it.

47degdev:
Task linked: [CU-1cruqbb Handle `var` correctly](https://app.clickup.com/t/1cruqbb)

nomisRev:
> Make KtLint happy

Sorry 😅 You can install a pre-commit hook through Gradle using ``addKtlintCheckGitPreCommitHook`.
I've also thought to make a CI script to auto-format and make a `ktlint formatted` commit, since we squash on merge it's not really an issue.

serras:
Ready for review. You can see how attaching invariants works in [this example](https://github.com/arrow-kt/arrow-meta/pull/843/files#diff-d8893c41325132d95cc89c5aceedc10903e6dd9ea55a0c7fc4413d35bfb1d0ebR90); feel free to propose a more idiomatic syntax.

plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintCollection.kt
```diff
@@ -470,15 +483,21 @@ private fun Solver.expressionToFormulae(
...
 
 internal fun Solver.isResultReference(ex: KtElement, bindingContext: BindingContext): Boolean {
-  val maybePostCall =
+  val maybeParentCall =
     ex.getParentResolvedCall(bindingContext)?.call?.callElement.getParentResolvedCall(bindingContext)
-  return if (maybePostCall != null && maybePostCall.postCall()) {
-    val expArg = maybePostCall.valueArguments.entries.toList()[1].value as? ExpressionValueArgument
+  return when {
+    maybeParentCall?.postCall() == true
+      -> maybeParentCall.valueArguments.entries.toList()[1].value
```
raulraja, line 490: I found a couple of cases where we are failing with an array index out of bounds as I was encoding the laws. Regardless of this instance here we need to do a general pass on all these cases and actually check we can access the args in the collections if those elements are there. I presume it was failing because I was referring to fields like `size` and we don't support that yet.

plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/SolverInteraction.kt
```diff
@@ -130,3 +130,16 @@ internal fun SolverState.checkConditionsInconsistencies(
       MetaErrors.InconsistentConditions.on(expression.psiOrParent, unsatCore)
     )
   }
+
+internal fun SolverState.checkInvariant(
+  formula: BooleanFormula?,
```
raulraja, line 135: We should also do a pass in another PR to ensure we are not taking nullable values as arguments as it's easier to delegate the responsibility at the call site to only invoke the function if there is an actual formula. It should simplify all functions where we return defaults based on nullable arguments.

raulraja (approved):
Thanks, @serras! 🙏

plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintCollection.kt
serras: I've replaced this code with one that looks for the `predicate` argument, instead of hardcoding the position. In fact, this was already wrong, since we provide different overloads of `post` with different argument order.

plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/SolverInteraction.kt
serras: Done

PR #845, opened 2021-08-23
Handle 'return' correctly

serras:
This PR fixes the handling of `return` nodes, which was incorrect until now, since it `abort`ed the whole computation from that point on. One important consequence was that post-conditions were *not* checked when appearing after a `return`, like in the following snippet:

```kotlin
fun f(x: Int): Int {
  return 2.post("greater than 0") { it > 0 }
}
```

In a technical side, this PR changes the signature of the checking functions from `ContSeq<Unit>` to `ContSeq<Return>`. This additional piece of information tells whether we simply got to the end of the block (`NoReturn`) or we were explicit (`ExplicitReturn`). In the latter case we also record an explicit label if it appears. The core of the handling is in:
- `checkReturnExpression`, where we handle the `ExplicitReturn` case and continue if it was "ours",
- `checkBlockExpression`, where we need to *not* continue with the next of the block if there was an explicit return in the current expression.

Unfortunately, there's some amount of accidental complexity in `checkCallArguments` and `checkSimpleConditional`. The reason is that Kotlin allows `return`s to appear everywhere, also within an argument or in a condition.

```kotlin
val f(x: Int): Int = f(return 2) // <- why not?
```

So there's some complicated handling to ensure that if when handling an argument a `return` is called, we stop the handling of the rest of arguments (we cannot just keep checking all of them, since we could obtain constraint which are not true). Same for conditions. **Help is more than welcome to refactor that terrible piece of code.**

### DIFF
--- a/plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintChecks.kt
+++ b/plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintChecks.kt
@@ -401,32 +401,32 @@ private fun SolverState.checkCallArguments(
   //   and in that case we stop the check of any more arguments
   //   (stopping there is important, since introducing additional constraints from
   //    other arguments may not be right in the general case)
-  fun <A> go(expressions: List<Triple<String, A, KtExpression?>>): ContSeq<Either<ExplicitReturn, List<Pair<String, String>>>> =
-    when (expressions.size) {
-      0 -> cont { emptyList<Pair<String, String>>().right() }
-      else -> {
-        val (name, _, expr) = expressions[0]
-        val argUniqueName =
-          if (expr != null && solver.isResultReference(expr, data.context.trace.bindingContext)) {
-            RESULT_VAR_NAME
-          } else {
-            names.newName(name)
-          }
-        checkExpressionConstraints(argUniqueName, expr, data).flatMap { returnInfo ->
-          when (returnInfo) {
-            is ExplicitReturn -> cont { returnInfo.left() }
-            else ->
-              go(expressions.drop(1)).flatMap {
-                it.fold(
-                  { r -> cont { r.left() } }, // if we found a return, do it
-                  { restArgs -> cont { (listOf(name to argUniqueName) + restArgs).right() } }
-                )
+  fun <A> acc(
+    upToNow: ContSeq<Either<ExplicitReturn, List<Pair<String, String>>>>,
+    current: Triple<String, A, KtExpression?>
+  ): ContSeq<Either<ExplicitReturn, List<Pair<String, String>>>> =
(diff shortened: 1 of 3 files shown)


47degdev:
Task linked: [CU-1cymjpx Bug: check post-conditions after return](https://app.clickup.com/t/1cymjpx)

serras:
For some reason CI is failing, but I cannot reproduce it locally, `gradle :arrow-refined-types:allTests` is successful.

plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintChecks.kt
```diff
@@ -350,17 +392,42 @@ private fun SolverState.checkCallExpression(
...
+): ContSeq<Either<ExplicitReturn, List<Pair<String, String>>>> {
+  // why is this so complicated?
+  //   in theory we just need to run checkExpressionConstraints over each argument
+  //   (in fact, the original implementation just did that, and then called .sequence())
+  //   alas, Kotlin allows arguments to include 'return' (yes, really!)
+  //   so we need to check after each step whether a ExplicitReturn has been generated
+  //   and in that case we stop the check of any more arguments
+  //   (stopping there is important, since introducing additional constraints from
+  //    other arguments may not be right in the general case)
+  fun <A> go(expressions: List<Triple<String, A, KtExpression?>>): ContSeq<Either<ExplicitReturn, List<Pair<String, String>>>> =
```
nomisRev, line 404: These can typically also be implemented using `scan`, or `runningFold`. There is a large set of these functions scan/reduce/fold but they all have different names than you'd find them in Haskell.

[runningFold](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/running-fold.html)
[scan](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/scan.html)
[collection ops](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections)

serras: Thanks for the pointers. At the end, it turned out that a simple `fold` was enough, with the caveat that now if an argument sets an `ExplicitReturn`, it's still threaded through the rest of the arguments (although unchanged). I don't think this will have any performance implication, since argument lists are short and the explicit return case almost never happens.

serras: [Here's the diff](https://github.com/arrow-kt/arrow-meta/pull/845/commits/bdb168184f10254af06ecf4a6b0e238b1e67f631) in case you want to have a closer look.

raulraja (approved):
Thanks @serras !

PR #846, opened 2021-08-24
[SMT plug-in] Initial implementation of fields

serras:
The goal is to be able to reference fields or properties in the pre- and post-conditions. For example, `.get` may only be called when the first argument is smaller than the list size.

The plan is outlined in [this document](https://docs.google.com/document/d/1gyg4vyUEDk0AKTiurPCUt-WT96SZQSlKzjGIVZHQt5o/edit?usp=sharing):
- [X] Introduce dependencies in `@Pre` and `@Post`
- [ ] Bring all fields as `DeclarationConstraint`s
- [ ] Introduce field definitions in the solver


47degdev:
Task linked: [CU-yjrh0t Fields](https://app.clickup.com/t/yjrh0t)

PR #848, opened 2021-08-27
Upgrade java_smt to 3.10.0

raulraja:


serras (approved):

raulraja:
@serras I have also pushed this branch that shows how to switch to Z3 if we ever want to https://github.com/arrow-kt/arrow-meta/compare/rr-z3-turnkey-java-smt-3-10-0

PR #849, opened 2021-08-30
Scoping fixes

serras:
This PR fixes two problems with the previous naïve approach to variables:
1. Shadowing was not taken into account, so the system would get confused if you declare, for example, a `val x` within the body of a conditional.
    - This has been solved by keeping an additional `CurrentVarInfo` in which the "top-most" declaration of a name appears earlier in the list. This maps each name to its "SMT name".
    - To keep the names nice for the non-shadowing case, the "SMT name" is initialized to the one declared by the user, and only further re-declaration create a new name. In particular, the parameter names receive the same "SMT names" as they have.
2. For mutable variables always the "last" assignment was chosen. However, this is **unsound** in general; for mutable variables **only the invariant** is the correct information to be taken into account.
    - As a consequence, one of the test now fails. I've introduced new tests to check that the behavior is the right one.

Additionally, I've added support for `this` and `this@label`. It was an easy one-liner after this refactoring.

### DIFF
--- a/plugins/liquid/refined-types-plugin/src/main/java/arrow/meta/plugins/liquid/errors/MetaDefaultErrorMessages.java
+++ b/plugins/liquid/refined-types-plugin/src/main/java/arrow/meta/plugins/liquid/errors/MetaDefaultErrorMessages.java
@@ -35,6 +35,9 @@ public class MetaDefaultErrorMessages implements DefaultErrorMessages.Extension 
         MAP.put(InconsistentConditions,
                 "unreachable code due to conflicting conditions: {0}",
                 FormulaRendererKt.RenderFormula);
+        MAP.put(InconsistentInvariants,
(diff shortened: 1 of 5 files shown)


47degdev:
Task linked: [CU-yycurg Implement real scoping](https://app.clickup.com/t/yycurg)

47degdev:
Task linked: [CU-1cykzm7 Check only top-level nodes](https://app.clickup.com/t/1cykzm7)

serras:
@nomisRev @raulraja sorry folks, I accidentally pushed the fix for another problem in this same branch :/

raulraja (approved):
Thanks, @serras. Looks great! I have another branch where we save the users' messages in the annotations and adopt the same convention of `require` where the predicate is strict, and the msg is lazy based on @nomisRev suggestions. Once this is merged, I will send a PR for that.

PR #851, opened 2021-09-01
Description of analysis errors

serras:
This PR adds documentation about what each kind of error that may arise during analysis means.


47degdev:
Task linked: [CU-10hpvvh Document information available for error messages](https://app.clickup.com/t/10hpvvh)

raulraja:
@serras, thanks for the excellent doc!. I have taken your doc here and included it in the kdocs in #852, so we have it as a reference as we shape each error template. As a side effect of being part of the kdocs this will show up in the API docs for the plugin that we can use and link where more examples are explaining what each case does. Once we have support in Ank to run compiler plugins, we can maybe have the example you added type-checked by the plugin rendering the right error messages. /cc @nomisRev

serras:
Nice idea, this keeps everything together :)

I'm closing this PR thus, in favor of #852.

PR #852, opened 2021-09-01
CU-yyb9hg `formula.dumpKotlinLike()` in preparation for better error messages

raulraja:
Transforms formulae to kotlin like strings in preparation for better error message reporting


franciscodr:
Task linked: [CU-yyb9hg Better error messages](https://app.clickup.com/t/yyb9hg)

serras:
Everything here looks good to me. I think there are a couple of cases missing, related to the "special uninterpreted functions" we use:
- any usage of `int`, `bool`, or `decimal` should be dropped. Their definition is in `Solver`, so you should be able to compare the name with those `FunctionDeclaration`,
- any usage of `field` must be changed to a field access. Note that the names look a bit like `field(kotlin.List.size, x)`, so you might need to do a bit of string manipulation.
    - In this respect, I'm using the `FqName` of the field/method as the name of the field, but if we prefer to use something like `class#field` to make our life easier, this can be changed without any issue.

PR #853, opened 2021-09-02
Special treatment for control flow functions

serras:
This PR adds special treatment in the analysis for the functions `also`, `apply`, `let`, and `run`, when they are called with a lambda expression afterwords. The reason is that we can get a much better analysis than resorting to a future higher-order function analysis.

Technically, the implementation is easy. For example, when we find code like:

```
thing.let { it + 1 }
```

what the analysis does is:
- analyze the argument `thing` as usual,
- create a new context in which `it` is equal to `thing`,
- analyze the body of the function, in which `it` now points to the right value,
- state that the return value of the whole computation is the return value of the block.

Since there are essentially four combinations:
1. whether `thing` goes in the block as `this`, or as `it`/given lambda argument,
2. whether we use the return value of the block, or we use `thing` itself as return value,

I've created a small data structure to hold that information, namely `ControlFlowFn`.

In addition, I've refactored `checkCallExpression`, because it was going wild.


47degdev:
Task linked: [CU-pzeh92 Handle basic higher-order functions from `Any`](https://app.clickup.com/t/pzeh92)

raulraja (approved):
Discussed with @serras in slack regarding making this more generic down the road but this is already great and covers these very important functions.

PR #854, opened 2021-09-02
[Analysis] Nullability (?. and ?:)

serras:
This PR adds support for:

- [X] Safe function calls `x?.fn(...)`
- [X] Checks for nullability `x == null`, `x != null`
- [x] The special operator `?:`

This is done by keeping an additional `null` uninterpreted function, which tells us whether we are in a definitely-null (if true), definitely-not-null (if false), or unknown situation.

### DIFF
--- a/plugins/liquid/refined-types-plugin/src/test/kotlin/arrow/meta/plugins/liquid/LiquidDataflowTests.kt
+++ b/plugins/liquid/refined-types-plugin/src/test/kotlin/arrow/meta/plugins/liquid/LiquidDataflowTests.kt
@@ -314,6 +314,29 @@ class LiquidDataflowTests {
   }
 
   @Test
+  fun `safe indexing`() {
+    """
+      ${imports()}
+      
+      @Law
+      fun <A> List<A>.safeGet(index: Int): A {
+        pre( index >= 0 ) { "index non-negative" }
+        pre( index < size ) { "index smaller than size" }
+        return get(index)
+      }
+      
+      @Law
+      fun <A> emptyListIsEmpty(): List<A> =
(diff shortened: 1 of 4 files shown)


47degdev:
Task linked: [CU-pzegv4 Handle safe calls `?.` and the Elvis operator `?:`](https://app.clickup.com/t/pzegv4)

plugins/liquid/refined-types-plugin/src/test/kotlin/arrow/meta/plugins/liquid/LiquidDataflowTests.kt
```diff
@@ -338,6 +361,51 @@ class LiquidDataflowTests {
...
+  fun `nullability, 1`() {
+    """
+      ${imports()}
+      fun nully1(x: Int?): Int? {
+        val y = x?.let { 1 }
+        return y.post({ it > 0 }) { "greater than 0" }
+      }
+      """(
+      // it seems that Kotlin's warning overrides ours
+      withPlugin = { failsWith { it.contains("`nully1` fails to satisfy the post-condition") } },
```
raulraja, line 376: I suppose this is the case related to the warnings that you mentioned in our call. I created https://app.clickup.com/t/1e0yfab so we can look into it after this PR

raulraja (approved):
Looking great @serras , thanks!

plugins/liquid/refined-types-plugin/src/test/kotlin/arrow/meta/plugins/liquid/LiquidDataflowTests.kt
serras: Yes, this is one of them, the other one is `bad predicates`.

PR #855, opened 2021-09-06
[Analysis] Loops

serras:
This PR handles `for`, `while`, and `do` loops. The general idea is the following: every time we find one looping construct, we begin two "branches",
1. On the one hand we check the body of the loop, which will hold whenever the condition is true. So we assert the condition, and then check the body. The only possible outside effect is if we return, so other than that we abort the execution.
2. Then we check whatever is after the loop, in which case we know that the condition is false. So we assert the negation of the condition, and ask the analysis to keep going.

You might be wondering: "mutable variables! in that case (1) is not really independent of (2)!" For this to be true we had to ensure that changing the value of a mutable variable does not change the analysis. And that holds, because for mutable variables we only keep as much information as its `invariant` block says, and that information is... well.. invariant throughout the analysis.

### DIFF
--- a/plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintChecks.kt
+++ b/plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintChecks.kt
@@ -25,14 +25,18 @@ import org.jetbrains.kotlin.psi.KtConstantExpression
 import org.jetbrains.kotlin.psi.KtDeclaration
 import org.jetbrains.kotlin.psi.KtDeclarationWithBody
 import org.jetbrains.kotlin.psi.KtDeclarationWithInitializer
+import org.jetbrains.kotlin.psi.KtDoWhileExpression
 import org.jetbrains.kotlin.psi.KtElement
 import org.jetbrains.kotlin.psi.KtExpression
+import org.jetbrains.kotlin.psi.KtForExpression
 import org.jetbrains.kotlin.psi.KtIfExpression
 import org.jetbrains.kotlin.psi.KtLabeledExpression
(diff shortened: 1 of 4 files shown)


47degdev:
Task linked: [CU-pzehv3 Loops](https://app.clickup.com/t/pzehv3)

plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintChecks.kt
```diff
@@ -908,6 +935,116 @@ private fun <A> SolverState.yesNo(conditionVars: List<Pair<A, String>>): List<Pa
   return go(conditionVars, emptyList())
 }
 
+enum class LoopPlace {
```
raulraja, line 938: Not for this PR, but when it's a good time, I plan to do a new reorganization of functions and files and packages to avoid this file and the one for gathering constraints to grow much bigger. Perhaps separating it into feature files like  `Loops,` etc... or whatever makes the most sense.

raulraja (approved):
thanks, @serras ! Very nice way to solve this issue 👏

raulraja:
Now, I'm wondering what full support for mutable vars without relying just on `invariant` looks like.

serras:
> Now, I'm wondering what full support for mutable vars without relying just on `invariant` looks like.

People have tried many different things. One possibility is to translate to [Single Static Assignment form](https://en.wikipedia.org/wiki/Static_single_assignment_form), in which case you need to introduce something called "phi nodes" for when multiple execution paths may lead to the same mutable change.

My idea for this implementation comes from the [`while` rule for Hoare triples](https://en.wikipedia.org/wiki/Hoare_logic#While_rule), which manages to only talk about the Boolean condition in the `while`.

PR #856, opened 2021-09-06
[Analysis] Annotated expressions and 'run' without receiver

serras:
Just a couple of expressions which were missing, but require just a few lines of code.

### DIFF
--- a/plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintChecks.kt
+++ b/plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintChecks.kt
@@ -18,6 +18,7 @@ import org.jetbrains.kotlin.codegen.kotlinType
 import org.jetbrains.kotlin.descriptors.CallableDescriptor
 import org.jetbrains.kotlin.descriptors.DeclarationDescriptor
 import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.KtAnnotatedExpression
 import org.jetbrains.kotlin.psi.KtBinaryExpression
 import org.jetbrains.kotlin.psi.KtBlockExpression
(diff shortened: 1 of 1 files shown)


plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintChecks.kt
```diff
@@ -491,29 +496,42 @@ data class ControlFlowFn(
...
-        val argumentName = blockElement.valueParameters.getOrNull(0)?.name ?: "it"
-        ControlFlowFn(thisElement, bodyElement, argumentName, ControlFlowFn.ReturnBehavior.RETURNS_ARGUMENT)
+  return if (blockElement != null && bodyElement != null) {
+    if (thisElement != null) {
+      when (resolvedCall.resultingDescriptor.fqNameSafe) {
+        FqName("kotlin.also") -> {
+          val argumentName = blockElement.valueParameters.getOrNull(0)?.name ?: "it"
+          ControlFlowFn(thisElement, bodyElement, argumentName, ControlFlowFn.ReturnBehavior.RETURNS_ARGUMENT)
+        }
+        FqName("kotlin.apply") ->
```
raulraja, line 515: Do we have some kind of a plan to support lambda args and be able to define these special cases also as Laws?

raulraja (approved):
thanks @serras !

plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/ConstraintChecks.kt
serras: Even if we did so, I think the language won't be expressive enough to implement something like `also`, where you want the function to be executed, yet the result is equal to the receiver. The closest I could think of is analyzing the inlined body, which could work quite well in these cases.

PR #857, opened 2021-09-06
CU-yyb9hg Error messages psi links

raulraja:
Links constrains to their counterpart psi elements and displays error messages like

For call validation where constraints are in a jar
```txt
error found: /var/folders/3g/zgtzwm_90zx5n8d3s1685b1w0000gn/T/Kotlin-Compilation17789789074855678202/sources/Source.kt: (23, 27): pre-condition `index smaller than size` is not satisfied in `get(0)`
  -> unsatisfiable constraint: `(0 < this0.size)`
  -> `0` bound to param `index` in `kotlin.collections.List.get`   
```

For call validation when the constraints are in the local module
```txt
error found: /var/folders/3g/zgtzwm_90zx5n8d3s1685b1w0000gn/T/Kotlin-Compilation16553951783276572048/sources/Source.kt: (16, 20): pre-condition `x is 42` is not satisfied in `bar(1)`
  -> unsatisfiable constraint: `(1 == 42)`
  -> `1` bound to param `x` in `test.bar` at /var/folders/3g/zgtzwm_90zx5n8d3s1685b1w0000gn/T/Kotlin-Compilation16553951783276572048/sources/Source.kt: (11, 15):          `

```

PR #858, opened 2021-09-06
Code cleanup to package things based on our model

raulraja:
Breaks down types and functions into our semantic model in terms of collection, checks, etc.
This PR is currently targeting #857 and should be reviewed after that one.


serras (approved):
Thanks for taking the time for  cleaning this up! 💙

PR #860, opened 2021-09-09
[Analysis] Exceptions (try, catch, finally)

serras:
This PR adds support for exception handling. Note that the support itself is very _rough_, since we have no detailed information about which exceptions may be thrown, so we follow _every_ possibility. But it's sound.


47degdev:
Task linked: [CU-11tuxc8 Exceptions](https://app.clickup.com/t/11tuxc8)

PR #861, opened 2021-09-10
[Analysis] Subtyping

serras:
This PR changes the analysis of declarations to go well with subtyping:
- check that the Liskov Substitution Principle is obeyed in overridden members
- look for constraints in overridden members if they cannot be found in the current one

### DIFF
--- a/plugins/liquid/refined-types-plugin/src/test/kotlin/arrow/meta/plugins/liquid/LiquidDataflowTests.kt
+++ b/plugins/liquid/refined-types-plugin/src/test/kotlin/arrow/meta/plugins/liquid/LiquidDataflowTests.kt
@@ -460,6 +460,65 @@ class LiquidDataflowTests {
       withoutPlugin = { compiles }
     )
   }
+
+  @Test
+  fun `subtyping, pre- and post-conditions are checked in overridden functions`() {
+    """
+      ${imports()}
+      open class A() {
+        open fun f(): Int = 2.post({ it > 0 }) { "greater than 0" }
+      }
+      
+      class B(): A() {
+        override fun f(): Int = 0
+      }
+      """(
+      withPlugin = { failsWith { it.contains("`f` fails to satisfy the post-condition") } },
+      withoutPlugin = { compiles }
+    )
+  }
+
+  @Test
+  fun `subtyping, Liskov Substitution Principle, 1`() {
+    """
+      ${imports()}
+      open class A() {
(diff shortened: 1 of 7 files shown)


47degdev:
Task linked: [CU-1258cnw Check parent conditions when checking a body](https://app.clickup.com/t/1258cnw)

47degdev:
Task linked: [CU-1258cf7 Modify search of constraints to consider the "nearest parent"](https://app.clickup.com/t/1258cf7)

plugins/liquid/refined-types-plugin/src/test/kotlin/arrow/meta/plugins/liquid/LiquidDataflowTests.kt
```diff
@@ -460,6 +460,65 @@ class LiquidDataflowTests {
...
+        override fun f(): Int = 0
+      }
+      """(
+      withPlugin = { failsWith { it.contains("`f` fails to satisfy the post-condition") } },
+      withoutPlugin = { compiles }
+    )
+  }
+
+  @Test
+  fun `subtyping, Liskov Substitution Principle, 1`() {
```
raulraja, line 482: Since this is related to subtyping, we should perhaps include tests for `object`, `interface`, `fun interface`, and anonymous inline implementations. Another question is generic and type-bound restrictions in the function declarations. For example, `fun <A: Comparable> A....` the Liskov substitution principle should also apply in these cases since the target functions can be refined by predicates carried in the type bounds.

raulraja (approved):
Looks great, thanks @serras !

plugins/liquid/refined-types-plugin/src/test/kotlin/arrow/meta/plugins/liquid/LiquidDataflowTests.kt
serras: One question there is: how should be attach laws to things without a body, such as members in an `interface`?

raulraja: For cases where they actually have a body, it would be the same, interfaces in Kotlin allow implemented functions.
For cases like interfaces or abstract classes where there is nobody all I can think at the moment is using the annotations directly or allowing a special kind of synthetic body that is just used for constraint declaration and that the compiler plugin removes to ensure the member remains abstract.

PR #863, opened 2021-09-13
[Analysis] Constructors

serras:
This PR's goal is to handle constructors by:
- [x] attaching the correct pre- and post-conditions to the constructors themselves
- [x] checking them as part of the body of a class / object
- [x] ensuring that subtyping is obbeyed
- [x] introduce type invariants when introducing a new SMT variable

### DIFF
--- a/plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/check/Declarations.kt
+++ b/plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/check/Declarations.kt
@@ -53,16 +53,16 @@ internal fun <A> SolverState.checkTopLevel(
   context: DeclarationCheckerContext,
   descriptor: DeclarationDescriptor,
   declaration: KtDeclaration,
+  resultSmtName: String?,
   bodyCheck: (checkPost: () -> Unit) -> ContSeq<A>
 ): ContSeq<A> {
   // bring the constraints in (if there are any)
   val constraints =
     constraintsFromSolverState(descriptor)?.let {
-      when (declaration) {
-        // special case for constructors: instead of $result we use this
-        is KtConstructor<*> ->
-          solver.renameDeclarationConstraints(it, mapOf(RESULT_VAR_NAME to "this"))
-        else -> it
+      if (resultSmtName != null) {
+        solver.renameDeclarationConstraints(it, mapOf(RESULT_VAR_NAME to resultSmtName))
+      } else {
+        it
       }
     }
   // perform the checks
@@ -88,7 +88,7 @@ internal fun <A> SolverState.checkTopLevel(
   }.flatMap {
     // check the body
     bodyCheck {
-      // and finally check the postconditions
+      // and finally check the post-conditions
       checkPostConditionsImplication(constraints, context, declaration)
     }
   }
@@ -99,7 +99,7 @@ internal fun SolverState.checkTopLevelDeclarationWithBody(
   descriptor: DeclarationDescriptor,
   declaration: KtDeclaration
 ): ContSeq<Return> =
-  checkTopLevel(context, descriptor, declaration) { checkPost ->
+  checkTopLevel(context, descriptor, declaration, null) { checkPost ->
     // only check body when we are not in a @Law
     doOnlyWhen(!descriptor.hasLawAnnotation(), NoReturn) {
       val result = solver.makeObjectVariable(RESULT_VAR_NAME)
@@ -115,13 +115,12 @@ internal fun SolverState.checkPrimaryConstructor(
   descriptor: DeclarationDescriptor,
   declaration: KtPrimaryConstructor
 ): ContSeq<Return> =
-  checkTopLevel(context, descriptor, declaration) { checkPost ->
+  checkTopLevel(context, descriptor, declaration, THIS_VAR_NAME) { checkPost ->
     val klass = declaration.getContainingClassOrObject()
-    val thisRef = solver.makeObjectVariable("this")
-    val data = CheckData(context, ReturnPoints.new(declaration, thisRef), initializeVarInfo(declaration))
+    val data = CheckData(context, ReturnPoints.new(declaration, solver.thisVariable), initializeVarInfo(declaration))
     ContSeq.unit.flatMap {
       // introduce 'val' and 'var' from the constructor
-      introduceImplicitProperties(thisRef, klass)
+      introduceImplicitProperties(klass)
     }.flatMap {
       // call the superclass constructors
       // (this will ultimately check the Liskov for classes)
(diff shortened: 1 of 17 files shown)


serras:
The current implementation needs a post-processing pass, in which references to fields in `pre` are changed to references to the actual parameters. Also post-conditions which mention `this` have to be changed to reference `$result` instead (which is the actual result of the constructor).

plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/check/Declarations.kt
```diff
@@ -36,34 +38,39 @@ import org.jetbrains.kotlin.resolve.checkers.DeclarationCheckerContext
...
-        descriptor.allParameters.forEach { param ->
-          if (!param.type.isMarkedNullable) {
-            val paramName = param.name.asString()
-            val notNullFormula = solver.isNotNull(solver.makeObjectVariable(paramName))
-            addConstraint(NamedConstraint("$paramName is not null", notNullFormula))
-          }
+  // bring the constraints in (if there are any)
+  val constraints = constraintsFromSolverState(descriptor)
+  // choose a good name for the result
+  // should we change it for 'val' declarations?
```
raulraja, line 49: the current `$result` name is safe because it's not a kotlin identifier, but we can change it to whatever makes sense

plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/collect/ConstraintCollection.kt
```diff
@@ -185,30 +192,135 @@ internal fun CompilerContext.collectDeclarationsConstraints(
...
+        if (call.preCall()) {
+          val rewritten = rewritePrecondition(solverState, context, call, formula.formula)
+          preConstraints.add(NamedConstraint(formula.msg, rewritten))
+        }
+        // in constructors 'require' has a double duty
+        if (call.postCall() || call.requireCall()) {
+          val rewritten = rewritePostcondition(solverState, formula.formula)
+          postConstraints.add(NamedConstraint(formula.msg, rewritten))
+        }
       }
```
raulraja, line 215: We should probably move these implementations inside the when branches to separate functions that hint at what is done on each case, but we can do that if you want in another cleanup PR

plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/collect/ConstraintCollection.kt
```diff
@@ -185,30 +192,135 @@ internal fun CompilerContext.collectDeclarationsConstraints(
...
+ * namely the pre- and post-conditions of init blocks
+ * should be added to their own list
+ */
+internal fun <A : KtConstructor<A>> KtConstructor<A>.constraintsFromConstructor(
+  solverState: SolverState,
+  context: DeclarationCheckerContext
+): List<Pair<ResolvedCall<*>, NamedConstraint>> =
+  this.containingClassOrObject?.let { klass ->
+    (klass.getAnonymousInitializers() + listOf(this)).flatMap {
+      it.constraintsFromFunctionLike(solverState, context)
```
raulraja, line 257: 👍 this is what I meant previously at observing multiple `init`

plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/collect/ConstraintCollection.kt
```diff
@@ -217,7 +329,10 @@ internal fun KtDeclaration.constraints(
 private fun KtElement.constraintsDSLElements(): Set<PsiElement> {
   val results = hashSetOf<PsiElement>()
   val visitor = callExpressionRecursiveVisitor {
-    if (it.calleeExpression?.text == "pre" || it.calleeExpression?.text == "post") {
+    if (it.calleeExpression?.text == "pre" ||
+      it.calleeExpression?.text == "post" ||
+      it.calleeExpression?.text == "require"
```
raulraja, line 334: this partially addresses https://app.clickup.com/t/p5eyun at least the `require` part. The other part is related to contracts; we can perhaps create a separate issue and look into it after our first release.

plugins/liquid/refined-types-plugin/src/test/kotlin/arrow/meta/plugins/liquid/LiquidDataflowTests.kt
```diff
@@ -568,6 +568,45 @@ class LiquidDataflowTests {
...
     )
   }
+
+  @Test
+  fun `class with two init blocks`() {
+    """
+      ${imports()}
+      data class A(val n: Int) {
+        init {
+          pre(n > 0) { "n must be positive" }
```
raulraja, line 578: How do we fail when multiple init blocks contain contradicting constraints? For example:

```kotlin
init {
  pre(x > 0) { "x > 0" }
}
init {
  pre(x < 0) { "x < 0" }
}
```

raulraja (approved):
Thanks, @serras looks great!

plugins/liquid/refined-types-plugin/src/test/kotlin/arrow/meta/plugins/liquid/LiquidDataflowTests.kt
serras: This is handled by the inconsistency check for pre-conditions, in the same way that it's handled for the other functions (be aware that this is not yet implemented, but that's the idea).

plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/collect/ConstraintCollection.kt
serras: Done!

serras:
My idea is to handle the introduction of information associated with a type as part of this PR. Hence I'm still marking it in progress.

plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/collect/ConstraintCollection.kt
serras: Indeed. We might want to separate this in two different tasks.

serras:
@raulraja ready for review! I've tried to cover different cases for tests, but feel free to suggest or add more.

This PR does not handle enum entries, which I expect to handle in a forecoming PR. But for this we need to discuss a bit what should be the relation between the parent and the child pre/postconditions.

PR #864, opened 2021-09-13
Fix gradle snippet for refined types plugin

kyusu:
When playing around with arrow-meta I've noticed that the gradle setup snippet for the refined types compiler plugin is actually showing the setup for the proofs plugin.


raulraja (approved):
Thanks @kyusu !

PR #865, opened 2021-09-17
[Analysis] Enumeration entries

serras:
This PR refactors and uses parts of the primary constructor check to check "enumeration entries."

### DIFF
--- a/plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/check/Expressions.kt
+++ b/plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/check/Expressions.kt
@@ -93,6 +93,7 @@ import org.jetbrains.kotlin.types.isNullable
 import org.jetbrains.kotlin.types.typeUtil.isBoolean
 import org.jetbrains.kotlin.types.typeUtil.isSubtypeOf
 import org.sosy_lab.java_smt.api.BooleanFormula
+import java.math.BigInteger
 
 // 2.2: expressions
 // ----------------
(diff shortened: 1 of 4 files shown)


plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/check/Expressions.kt
```diff
@@ -696,6 +697,13 @@ private fun SolverState.checkConstantExpression(
   NoReturn
 }
 
+private fun String.toBigIntegerSeveralAttempts(): BigInteger? =
```
raulraja, line 700: LOL what's this xD?

raulraja (approved):
Looks great @serras! exciting to see new declarations getting support.

plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/check/Expressions.kt
serras: Me after figuring out that after 30 minutes debugging it was only a matter of parsing an integer correctly xD

I've changed it to `asIntegerLiteral` and made it follow Kotlin's grammar.

serras: But maybe it should be replaced with https://github.com/JetBrains/kotlin/blob/master/compiler/psi/src/org/jetbrains/kotlin/parsing/ParseUtils.kt altogether

PR #866, opened 2021-09-20
[Analysis] Several fixes required for better null analysis

serras:


serras:
@raulraja @nomisRev I am not that happy with the big post-conditions you sometimes need to write (see https://github.com/arrow-kt/arrow-meta/pull/866/files#diff-d8893c41325132d95cc89c5aceedc10903e6dd9ea55a0c7fc4413d35bfb1d0eb), so I am open to having some sort of special construct in our refinement DSL.

plugins/liquid/refined-types-plugin/src/test/kotlin/arrow/meta/plugins/liquid/LiquidDataflowTests.kt
```diff
@@ -403,23 +403,56 @@ class LiquidDataflowTests {
...
-  fun `nullability, 1`() {
+  fun `nullability, predicate with null, 1`() {
     """
       ${imports()}
-      fun nully1(x: Int?): Int? {
+      fun nully1a(x: Int?): Int? {
         val y = x?.let { 1 }
-        return y.post({ it > 0 }) { "greater than 0" }
+        return y.post({ 
+          if (x == null) { it == null }
```
raulraja, line 412: We can potentially simplify this introducing something like `nullable(x) { it > 0 }` which performs 
```kotlin
if (x == null) { it == null } 
else { (it != null) && (it > 0) } 
```
Still most users will understand what it is now as acceptable IMO

plugins/liquid/refined-types-plugin/src/main/kotlin/arrow/meta/plugins/liquid/phases/analysis/solver/collect/ConstraintCollection.kt
```diff
@@ -572,12 +579,36 @@ internal fun Solver.expressionToFormula(
       ex.operationToken.toString() == "EXCLEQ" &&
       ex.right is KtConstantExpression && ex.right?.text == "null" ->
       ex.left?.let { expressionToFormula(it, bindingContext) as? ObjectFormula }?.let { isNotNull(it) }
+    ex is KtBinaryExpression &&
+      ex.operationToken.toString() == "ANDAND" ->
```
raulraja, line 583: We should probably extract big bodies within branches to their own functions but I can take care of those on a clean up after I finish isolating away the kotlin compiler apis

raulraja (approved):
Looks great, thanks @serras !

PR #867, opened 2021-09-22
[Analysis] Always use topmost declaration as field name

serras:
This PR fixes a bug in how we refer to fields. Before this change, it was the case that `List.size` was declared as a different field than `Collection.size`, but this is wrong! (note that there could be stronger post-conditions in the former than in the latter, but this is different from the field declaration itself.) The solution is two-fold:

1. When creating a reference to a field, use the `topmostDescriptor` instead of simply the descriptor.
2. When creating the unique numbers for the fields, assign to all those which override another field the same number.

As a result the function `introduceFieldNamesInSolver` has become a bit more complicated. Essentially, we iterate over the constraints and declarations keeping a couple of mutable sets, which we later use to introduce the right information.


serras:
@raulraja do you think this can be merged into `main` before your refactoring?

raulraja (approved):
Looks great thanks @serras !

PR #868, opened 2021-09-23
Update to Kotlin 1.5.31, Arrow 1.0, Kotest 5, and Gradle 6.9.1

serras:
This PR updates the version of those dependencies, and then makes everything compile again.

The main problem is that in version 1.5.30 the `IrInterpreter` within the Kotlin compiler changed dramatically. Instead of bringing it again (since we may remove it if we deprecate the Refined Types plug-in), I've brought the missing pieces from Kotlin 1.5.20's source code.


nomisRev:
🧙‍♂️

serras:
I've also updated Kotest and Gradle to their latest versions

libs/arrow-meta/src/main/kotlin/arrow/meta/phases/codegen/ir/IrUtils.kt
```diff
@@ -289,3 +284,17 @@ fun IrReturnTarget.returnType(irBuiltIns: IrBuiltIns) =
     is IrReturnableBlock -> type
     else -> error("Unknown ReturnTarget: $this")
   }
+
+// copied from org.jetbrains.kotlin.ir.interpreter
+fun IrClass.internalName(): String {
```
nomisRev, line 289: We could delegate to the original impl if it was internal by bypassing the internal modifier.

Since this is a jvm project it will always be public in the binary. That way we could avoid code duplication but also prevent relying on the hack annotation everywhere.

(Need to look up the annotations)

serras: That is really out of my knowledge. Do you have a pointer or an example where this is done? Because using the original implementation is clearly better than keeping this one in sync with the compiler.

nomisRev: `@file:Suppress("UNUSED", "INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")` you need to add this to the file, which will allow you to import and call the `internal` function from the Kotlin Compiler.

It needs to be on the file because otherwise the `import xxxx` will not resolve.

nomisRev: Example usage:
 
 - https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-debug/src/DebugProbes.kt#L5
 
 
 They're accessing `DebugProbesImpl` from the core module.

serras: Done!

raulraja (approved):
Thanks @serras !

PR #872, opened 2021-09-29
First step toward abstracting compiler APIs in Arrow Analysis

raulraja:
This PR abstract away the Kotlin compiler dependencies from the Arrow Analysis runtime.
It follows an ADT-like approach with an interpreter function `model` that provides the Kotlin specific impl for an AST based on the PSI model.
This PR is the first step to enable the possibility to run Arrow Analysis over other compiler plugins for targets such as Java and Scala.


serras (approved):
One step further to support other languages!

PR #873, opened 2021-09-29
[Analysis] Extra example of invariant checking

serras:
This is an interesting example which shows that we can use this plug-in in the same way we could use the Refined one.


raulraja (approved):
This is great; as soon as we have support for Strings, we can drop the other implementation based on the IR Interpreter

PR #876, opened 2021-09-30
Remove Refined Types plug-in and `IrInterpreter`

serras:
This PR removes the Refined Types plug-in, and the `IrInterpreter` code supporting it in the main `arrow-meta` library. Maintaining the latter has proven to be cumbersome, because it is essentially a copy from the main compiler code, which often changes.

This PR does not address the change of the name `refined-types-plugin` to `analysis-plugin`, which will come in another PR, once the current Analysis PRs are merged.


nomisRev:
> +56 −10,222
🤯

raulraja (approved):
Thanks @serras ! Very happy to get rid of this in favor of the new analysis.

PR #877, opened 2021-09-30
[Analysis] Kotlin 1.6.0-M1 support

serras:
This PR makes the necessary changes for Arrow Meta to compile both in 1.5.31 and 1.6.0-M1. Note that it's a PR against #876 because it was useless to update `IrInterpreter` to remove it aftwerwards.

In addition, the GitHub Action is now configured to check against the two Kotlin versions, and also against Java targets 1.9 and 11.


nomisRev (approved):
Really nice!! 👏 👏 👏

raulraja (approved):
thanks @serras !

serras:
@raulraja @nomisRev  it seems that the "old" `build_artifacts` will remain as "waiting" until everything is merged in `main`. However, I cannot merge because I have no permissions, could you any of you do so?

nomisRev:
@serras it's green 🥳
I noticed when changing Github Actions sometimes it tries to look for an old version of the job as a `required` step. This means we have to force merge (admin).

We can also temporarily remove the `build_artifacts` requirement.

nomisRev:
Just saw your message. I will remove `build_artifacts` for now as a requirement, because all subsequent PRs will have the same issue afaik.

We have also removed this requirement in Arrow for the same reason atm. I cannot re-add new requirements there :/

PR #878, opened 2021-10-01
[Analysis] Reorganize

serras:
This PR has two goals:
- Moved from `refined` and `liquid` terminology towards `analysis`,
- Separate the common parts of `analysis` from the Kotlin-specific ones


raulraja (approved):
thanks @serras !

PR #880, opened 2021-10-04
[Analysis] Initial law work

serras:
This PR paves the way to introducing the laws of the Kotlin standard library. Two concrete goals:
- separate the `laws` library from the `types` library, and make the former been treated by our plug-in,
- ensure that a first example works correctly.

At this point, this entails a lot of bug-fixing in the `commons` project, especially with respect to handling the constraints from annotations.


plugins/analysis/common/src/main/kotlin/arrow/meta/plugins/analysis/phases/analysis/solver/ast/context/Type.kt
```diff
@@ -20,4 +20,8 @@ interface Type {
   fun isShort(): Boolean
   fun isUnsignedNumberType(): Boolean
   fun isChar(): Boolean
+
+  fun isEqualTo(other: Type): Boolean =
```
raulraja, line 24: If this is meant to compare strict equality and not bounds an alternative impl is to leave this abstract and use the underlying kotlin apis for `KotlinType.equals` which delegate to the StrictEqualityTypeChecker
```kotlin
final override fun equals(other: Any?): Boolean {
        if (this === other) return true
        if (other !is KotlinType) return false

        return isMarkedNullable == other.isMarkedNullable && StrictEqualityTypeChecker.strictEqualTypes(unwrap(), other.unwrap())
    }
```

plugins/analysis/common/src/main/kotlin/arrow/meta/plugins/analysis/phases/analysis/solver/collect/ConstraintCollection.kt
```diff
@@ -323,24 +325,74 @@ private fun SolverState.addConstraints(
...
-        val fn = subjectFqName.name.split(".").lastOrNull()
-        descriptor.module.getPackage(pck)?.getMemberScope()?.getContributedDescriptors { it == fn }?.firstOrNull()
+    val name = lawSubject.argumentValueAsString("fqName")
+    val result = name?.let {
+      descriptor.module.obtainDeclaration(FqName(it), descriptor)
+    }
+    result.also {
+      if (it == null) {
+        if (name == null)
+          throw Exception(ErrorMessages.Parsing.subjectWithoutName(descriptor.fqNameSafe.name))
```
raulraja, line 350: should we report these instead of throwing an exception?

plugins/analysis/kotlin-plugin/src/main/kotlin/arrow/meta/plugins/analysis/phases/ir/ConstraintsAnnotations.kt
```diff
@@ -95,8 +96,20 @@ private fun IrUtils.lawSubjectAnnotationFromClassId(classId: ClassId, descriptor
...
+
+private fun DeclarationDescriptor.getLawName(): String {
+  val containing = containingDeclaration
+  return when {
+    this is org.jetbrains.kotlin.descriptors.PackageFragmentDescriptor ||
+      this is org.jetbrains.kotlin.descriptors.PackageViewDescriptor ||
+      this is org.jetbrains.kotlin.descriptors.ModuleDescriptor ||
+      containing == null
+    -> fqNameSafe.asString()
+    else -> "${containing.getLawName()}/${name.asString()}"
```
raulraja, line 134: If this is calling itself we may be able to make this function `tailrec`

plugins/analysis/kotlin-plugin/src/test/kotlin/arrow/meta/plugins/analysis/AnalysisTests.kt
```diff
@@ -321,6 +321,21 @@ class AnalysisTests {
     )
   }
 
+  @Test
+  fun `pre-conditions for subjects`() {
+    """
+      ${imports()}
+      @Pre(messages = ["not zero divisor"], formulae = ["(not (= (int other) 0))"], dependencies = [])
+      @Subject(fqName = "kotlin/Int/div")
```
raulraja, line 329: I see we are using `kotlin/Int/div`. Shouldn't this use the fqName version that uses "kotlin.Int.div" for the name?

serras: I had to change this logic, because the fq-name was ambiguous. For example, you could not tell upfront whether something like `kotlin.collections.arrayOf` was package `kotlin` class `collections`, package `kotlin.collections`... So I've decided to change it to `module/class1/class2/.../member`, which makes it clear for example that `kotlin.collections/arrayOf` is a function within the module, but `kotlin/Int/div` is a function within a class.

Note that I might not have found the good way to obtain a descriptor given the fq-name. If there's one in the Kotlin compiler, I think that would be the good one to use.

plugins/analysis/common/src/main/kotlin/arrow/meta/plugins/analysis/phases/analysis/solver/collect/ConstraintCollection.kt
serras: There are two reasons:
- at some point these should go, they are really cases which should not happen, unless somebody tinkers with the annotations;
- there is no particular `PsiElement` to which they can be hung, so it was not clear how to turn those into messages.

plugins/analysis/common/src/main/kotlin/arrow/meta/plugins/analysis/phases/analysis/solver/ast/context/Type.kt
serras: Done!

plugins/analysis/kotlin-plugin/src/main/kotlin/arrow/meta/plugins/analysis/phases/ir/ConstraintsAnnotations.kt
serras: Done in a new commit

serras:
Yay! The build fails, but that's actually *great*, because that means that the example is working! That pre-condition comes from `analysis-laws`, and it's used in a different project :)

Captura de pantalla 2021-10-04 a las 21 52 31

serras:
@raulraja this is now ready for review, I've added the two additional changes we've discussed:
- fixing the way constraints are searched for (not only fully-qualified name, but also compatible type),
- adding a new `Laws` interface which can be extended to declare a bunch of laws.

PR #882, opened 2021-10-08
[Analysis] Lambdas + fix scopes

serras:
This PR does two (interrelated) things:
- adds support for lambda expressions and local functions: we cannot check laws on those yet, but at least we check that the body is correct (for example `listOf(1, 0).map { x -> 1 / x }` is correctly flagged as potentially problematic);
- fixes a long-standing problem with scopes: we were handling scoping information by means of a mutable piece of data; it turns out this was not behaving correctly. The solution has been to thread this information manually as a piece of `StateAfter` data. The main combinator is called `inScope`, which generates a new inner scope (for example, for local functions, but also `if` blocks, `try` blocks, and so on).


serras:
There's one missing functionality I am not sure how to handle, which is treating blocks which use the `it` variable. AFAIK, the compiler does not reflect in the tree whether `{ 1 / it }` is a block or a lambda expression with one parameter. Any ideas?

PR #884, opened 2021-10-11
[Analysis] 'trustMe' + small fixes for constraint collection

serras:
This PR has a couple of additions for better handling constraints and laws:
- you can mark a call with `trustMe` and then the pre-conditions won't be checked (but of course, it may fail at run-time),
- you can use a Boolean property directly as a pre/postcondition. For example, in `Result`, you can now say `post { it.isSuccess }` instead of the longer `post { it.isSuccess == true }`.


serras:
>  If we want to expose `trustMe` as a public API for users to bypass analysis perhaps we should name it `unsafeCall` or similar, something that hurts more to see. 

Done, we have `unsafeCall` now.

> Perhaps we should also consider a block version of it.

I've called it `unsafeBlock`, but I'm happy to change the name.

Issue #893, opened 2021-10-27
[BUG] @annotations as ScopedList type have wrong separator


paranoidmonoid:
**Describe the bug**
`@annotations` in `ClassOrObjectScope` class are of `ScopedList` type. The default separator for `ScopedList` is a comma and no custom separator is passed. This results in the following: if there's more than one annotation, they become concantenated with comma when casted to string.

**To Reproduce**
See branch reproducer: https://github.com/paranoidmonoid/kotlin-serial-name/tree/reproducer-commas
Run Main
In console output you will see the following (which doesn't show the full issue)
`Exception in thread "main" kotlinx.serialization.SerializationException: Serializer for class 'Cat' is not found.
Mark the class as @Serializable or provide the serializer explicitly.`

To get more precise logs, in "serial name plugin" add something like `error(newDeclaration)` below `newDeclaration`
Note, that annotations in logs are separated with commas:
```
java.lang.IllegalStateException: @Serializable, @SerializeWithStrategy(SerializationStrategy.KebabCase) data class Cat(@SerialName("nickname") val name: String, @kotlinx.serialization.SerialName("age") val age: Int, @kotlinx.serialization.SerialName("owner-name") val ownerName: String, @kotlinx.serialization.SerialName("breed") val breed: String) {}
```

**Expected behavior**
A custom separator (whitespace) will be passed like for `declarations` in `ClassOrObjectScope`

**Screenshots**

**Environment (please complete the following information):**
- KOTLIN_VERSION=1.5.31
- JVM_TARGET_VERSION=1.8
- ARROW_VERSION=1.0.0
- ARROW_META_VERSION=1.5.0-RC-SNAPSHOT
- Kotlin: 212-1.5.31-release-546-IJ4638.7

serras:
Closed by #915

PR #915, opened 2021-11-09
[WIP] Add whitespace as separator for annotations

paranoidmonoid:
Fix for https://github.com/arrow-kt/arrow-meta/issues/893

Add a whitespace as a separator for `@annotations` in `ClassOrObject`

### DIFF
--- a/libs/arrow-meta/src/main/kotlin/arrow/meta/quotes/classorobject/ClassOrObject.kt
+++ b/libs/arrow-meta/src/main/kotlin/arrow/meta/quotes/classorobject/ClassOrObject.kt
@@ -29,7 +29,7 @@ import org.jetbrains.kotlin.psi.psiUtil.modalityModifierType
 open class ClassOrObjectScope<out T : KtClassOrObject>(
   override val value: T,
   override val descriptor: ClassDescriptor?,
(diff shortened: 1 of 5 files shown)


paranoidmonoid:
@raulraja Do you have the guidelines about testing the changes? Also, i noticed that the same issue is potentially in `File`, `ModifierList` and `AnnotatedExpression`.

Should I include them in this PR, too?

raulraja:
@paranoidmonoid thanks for the PR! . These kinds of issues can be tested similar to how it is done in https://github.com/arrow-kt/arrow-meta/blob/main/libs/arrow-meta/src/test/kotlin/arrow/meta/quotes/scope/templates/IfExpressionTest.kt

You create a plugin that does the transformation that you want to test and a compiler test where the plugin is applied and you assert things like if it compiles, the output contains certain words or a program after processed produces a value.

serras:
@paranoidmonoid thanks for sending the fix in! I've taken the liberty of merging with the latest `main` and running the formatter; otherwise the build fails on a preliminary step and tests don't really run (btw, you can run the formatting yourself by using `gradle spotlessKotlinApply`).

serras:
I've taken a look at this and it turned out that more annotations had this problem, so thanks for catching this for us, @paranoidmonoid. Once the CI passes, I'll merge :)

serras (approved):

PR #918, opened 2021-11-15
Allow different extension names for Meta plug-ins

serras:
This is needed to apply several compiler plug-ins over the same project.


raulraja (approved):
looks good, thanks @serras !

PR #919, opened 2021-11-16
Move to Kotlin 1.6.0

serras:


nomisRev (approved):
🥳  Is there an easy way to keep track of the latest versions of Kotlin?

I.e. is there an `1.6.1-eap` available? And is it worth building against that now instead of `1.5.31`?

serras:
@nomisRev I don't know of an easy way... In any case, we are a bit stuck with `1.5.31` as the "main" target because `kotlin-compile-testing` only works with that one.

PR #921, opened 2021-11-17
[Analysis] Add more fallbacks to Kotlin analysis interpreter

serras:
Sometimes the Kotlin compiler surprises us with more members of the `Descriptor` hiearchy. For most of them we do not really want to do something special, so I've reworked the `model` call to take care of those with a fallback implementation.

This has bitten me in particular while using the Analysis plug-in in an external project:

```
e: kotlin.NotImplementedError: An operation is not implemented: Missing impl for org.jetbrains.kotlin.resolve.calls.util.FakeCallableDescriptorForObject@d1bb6a5 (org.jetbrains.kotlin.resolve.calls.util.FakeCallableDescriptorForObject)
        at arrow.meta.plugins.analysis.phases.analysis.solver.ast.kotlin.ast.KotlinInterpreterKt.model(KotlinInterpreter.kt:46)
```


plugins/analysis/kotlin-plugin/src/main/kotlin/arrow/meta/plugins/analysis/phases/analysis/solver/ast/kotlin/ast/KotlinInterpreter.kt
```diff
@@ -42,7 +42,11 @@ fun <
     is ReceiverParameterDescriptor -> KotlinReceiverParameterDescriptor(this).repr()
     is LocalVariableDescriptor -> KotlinLocalVariableDescriptor(this).repr()
     is PackageFragmentDescriptor -> KotlinPackageFragmentDescriptor(this).repr()
+    // fallback cases: we sometimes find unknown descriptors
```
raulraja, line 45: 👍

raulraja (approved):
thanks @serras !

PR #923, opened 2021-11-17
[Analysis] Improvements to formula dump

serras:
This PR changes part of the logic for showing a formula back to a user, and fixes some corner cases:
- better handling of negation: whereas before we had `(! x > 0)` now we proper show `!(x > 0)`,
- special cases for negating equality or comparisons in formulae,
- better handling of parentheses,
- the `null` special function is now shown as in `x == null`


raulraja (approved):
thanks @serras !

PR #925, opened 2021-11-18
[Analysis] Generate annotations in Java plug-in

serras:


plugins/analysis/java-example/build.gradle.kts
```diff
@@ -0,0 +1,29 @@
...
+  annotationProcessor(projects.arrowAnalysisJavaPlugin)
+}
+
+tasks.withType<JavaCompile> {
+  javaCompiler.set(javaToolchains.compilerFor {
+    languageVersion.set(JavaLanguageVersion.of(11))
+  })
+  options.compilerArgs.addAll(listOf(
+    "-parameters",
+    "-Xplugin:ArrowAnalysisJavaPlugin"
```
raulraja, line 27: We may have to also have this in the Gradle plugin in addition to allowing this kind of usage. the Gradle plugin can apply these settings automatically over Java projects.

raulraja (approved):
Amazing to see the progress in Java, thanks @serras !

plugins/analysis/java-example/build.gradle.kts
serras: Indeed, that is the next step. It needs some further investigation, since we cannot reuse the Arrow Meta common implementation, as it targets Kotlin projects.

Issue #926, opened 2021-11-19
[BUG] Broken Arrow Meta Gradle plug-ins


serras:
It seems that I've somehow broken the Gradle plug-ins for Arrow Meta in one of the PRs. Right now whaetever plug-in you try to apply, you are always welcome with:

```
e: java.lang.IllegalStateException: Generated sources output dir is not found: directory to locate sources
        at arrow.meta.phases.analysis.DefaultElementScopeKt.getOrCreateBaseDirectory(DefaultElementScope.kt:495)
        at arrow.meta.plugins.analysis.phases.PhasesKt$analysisPhases$1.invoke(Phases.kt:38)
        at arrow.meta.plugins.analysis.phases.PhasesKt$analysisPhases$1.invoke(Phases.kt:31)
        at arrow.meta.dsl.analysis.AnalysisSyntax$analysis$1.doAnalysis(AnalysisSyntax.kt:89)
```

Alas, I cannot find the source of the problem. My guess is that it's related to the introduction of different names for plug-ins, which was required to be able to apply more than one in the same project.

raulraja:
@serras I'm looking at this today. Thanks.

serras:
Closed by #927

PR #927, opened 2021-11-22
Use different MetaCliProcessors

serras:
I think this solves #926. The idea is that each plug-in should provide its own local configuration (which you can still override in a global fashion by setting the `arrow.meta.generated.src.folder` property). At least the plug-ins can now be applied without problems.


serras:
This builds but there are issues with the tests, investigating now...

raulraja (approved):
Thanks @serras ! this makes perfect sense that each plugin has its own

PR #931, opened 2021-11-25
Use different versions for library and plug-ins

serras:
This separates the versioning scheme so that the library and the plug-ins can evolve separately:
- The main Meta library follows the Kotlin versioning scheme in a loose way, so that people can easily see which compiler versions are compatible. Hence the current version is `1.6.0-SNAPSHOT`.
- The plug-ins have been set to `2.0-SNAPSHOT`.


raulraja (approved):
thanks @serras !
The only issue I see with the new version is that there have been already snapshots for larger versions in the case of the plugins. A system that observes snapshots won't resolve them as new or ever resolve them until we do a release. Maybe we should change the artifact ids before the final release so there is no confusion as it has happened with arrow-meta vs arrow-meta in arrow still kapt based.

serras:
What about bumping the version of all plugins to 2.0? That would also take care of the problem, but we wouldn’t need to change every id (I am a bit worried about documentation and example projects).

serras:
@raulraja I've change the hardcoded versions in plug-ins to refer to a compile-time generated file instead, so now the only source of truth is the top-level `gradle.properties` file. I've also bumped the version of all the plug-ins to 2.0, so no more problems for people using previous versions.

PR #947, opened 2021-12-03
[Analysis] Fix problems with fields

serras:
I ran into a bug while working on the back-end example; it turned out that the numbering for fields was not correctly initialized at all spots. This was the result of one of the optimizations we applied, of not introducing field names until required; however some places were missing. I've added the check now to `addConstraint`, which is the single point of constraint addition in the system.


serras:
The PR is a big bigger than expected because to query the fields we need access to the context, so it has to be passed from any single place where it's used.

raulraja (approved):
Thanks @serras !

PR #954, opened 2021-12-09
Optics plug-in using KSP

serras:
Translate the usage of Meta to KSP to generate optics. The Gradle plug-in has been removed, since when using KSP + Multiplatform you should use `kspJvm` and things like that, and also using KSP directly is just simpler than a plug-in.

```kotlin
plugins {
    id("com.google.devtools.ksp") version "1.6.0-1.0.1"
    kotlin("jvm") version "1.6.0"
}

dependencies {
    ksp("io.arrow-kt:arrow-optics-ksp:2.0-SNAPSHOT")
}
```


nomisRev:
That was fast, hope my estimation was not too far off 😄
Awesome @serras 👏 👏 👏

nomisRev (approved):

PR #956, opened 2021-12-10
Enable class `kind` to be `open`.

vnermolaev:
Value `kind` of `ClassDeclaration` can be not only `data` and `sealed` kinds, but also `open`. This PR adds supports for this.


vnermolaev:
I wanted to add a test, but apparently, I am doing something wrong.
The test is below

`libs/arrow-meta/src/test/kotlin/arrow/meta/quotes/ClassDeclarationKind.kt`
```kotlin
package arrow.meta.quotes.scope.plugins

import arrow.meta.CliPlugin
import arrow.meta.Meta
import arrow.meta.invoke
import arrow.meta.phases.CompilerContext
import arrow.meta.quotes.Transform
import arrow.meta.quotes.classBody
import arrow.meta.quotes.classDeclaration

open class ClassDeclarationKindPlugin : Meta {
  override fun intercept(ctx: CompilerContext): List<CliPlugin> = listOf(classDeclarationKind)
}

private val Meta.classDeclarationKind: CliPlugin
  get() = "Class Declaration Kind Plugin" {
    meta(
      classDeclaration(this, { true }) { (c, _) ->
        Transform.replace(
          replacing = c,
          newDeclaration =
          """|$`@annotations` $kind $name $`(typeParameters)` $`(params)` : $supertypes {
             |  $body
             |}""".`class`
        )
      }
    )
  }
```

`libs/arrow-meta/src/test/kotlin/arrow/meta/quotes/scope/templates/ClassDeclarationKindTest.kt`
```kotlin
package arrow.meta.quotes.scope.templates

import arrow.meta.plugin.testing.Code
import arrow.meta.plugin.testing.CompilerTest
import arrow.meta.plugin.testing.CompilerTest.Companion.source
import arrow.meta.plugin.testing.assertThis
import arrow.meta.quotes.scope.plugins.ClassBodyPlugin
import arrow.meta.quotes.scope.plugins.ClassDeclarationKindPlugin
import org.junit.jupiter.api.Test

class ClassDeclarationKindTest {

  companion object {
    private val openClassTest =
      """
      | //metadebug
      |
      | open class OpenClass {
      |   public val x = 0
      | }
      """.source
  }

  @Test
  fun `Validate class body scope properties`() {
    validate(openClassTest)
  }

  private fun validate(source: Code.Source) {
    assertThis(
      CompilerTest(
        config = { listOf(addMetaPlugins(ClassDeclarationKindPlugin())) },
        code = { source },
        assert = { quoteOutputMatches(source) }
      )
    )
  }
}
```

This test works, but if I change the plugin to add extra thing to the body of the processed class, as such,
```kotlin
"""|$`@annotations` $kind $name $`(typeParameters)` $`(params)` : $supertypes {
     |  $body
     |  val o: Int = 1
     |}""".`class`
```
I observe no difference in the processed source and the test above still passes.

To verify this behaviour one can put a breakpoint [here](https://github.com/arrow-kt/arrow-meta/blob/d11b1e17fec8db1890f967a515b42a7897310416/libs/meta-test/src/main/kotlin/arrow/meta/plugin/testing/CompilationAssertions.kt#L228) and observe that there is no difference between `actualSourceWithoutCommands` and `expectedSourceWithoutCommands`.

serras:
Indeed, it seems that the transformation is never applied. The tests check that the source coincides with the input, so they won't catch his problem. @raulraja do you have an idea about why this is not working?

raulraja:
@serras, I have no idea what is going on but also not sure if worth exploring. @vnermolaev, with the upcoming stability in FIR, the quotes model based on PSI + Descriptors have to change completely. We will have to remove the quote and template system to provide a solution based on this new model that will become standard in the Kotlin compiler.

vnermolaev:
@serras  thanks for looking into this. I would like to make two points.

- @raulraja All right, I trust your judgement, but please be aware that (I suspect) no test in `libs/arrow-meta/src/test/kotlin/arrow/meta/quotes/scope/templates` works, because the respective plugins use `identity()` call which is a templated string. In fact, replacing `identity` with `emptyBody` in, say, `BinaryExpressionPlugin` does not result in a change of the target test class.

- This PR targets the bug regarding skipping the `open` modifier. It works in the wild :) Please consider it separately of the testing behavior.

I am really looking forward to the stable version. I use Arrow Meta heavily in my project.

serras:
@vnermolaev I agree with your judgment, let's merge this once the checks pass.

serras (approved):

PR #957, opened 2021-12-10
[Analysis] Two small things

serras:
1. Support for string templates
2. Change the level of "unreachable code" to a warning in SARIF, to align it with the Kotlin and Java plug-ins

These are both required for the `backend-example` to work well


raulraja (approved):
thanks @serras!, I'm sure this improves the backend example 👏

PR #962, opened 2021-12-15
[Analysis] Docs about wrappers

serras:
This should only be merged after @franciscodr had merged the solution to the problem with docs publication.

### DIFF
--- a/docs/docs/_data/analysis/sidebar.yml
+++ b/docs/docs/_data/analysis/sidebar.yml
@@ -14,6 +14,9 @@ options:
     - title: Types and invariants
       url: /analysis/types
 
+    - title: Wrappers
+      url: /analysis/wrappers
+
     - title: Fields
       url: /analysis/fields
 
--- a/docs/docs/analysis/wrappers/README.md
+++ b/docs/docs/analysis/wrappers/README.md
@@ -1,7 +1,7 @@
 ---
 layout: docs-analysis
 title: Analysis - Wrappers
-permalink: /analysis/wrappers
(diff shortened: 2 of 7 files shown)


docs/docs/analysis/wrappers/README.md
```diff
@@ -0,0 +1,89 @@
...
+  return (x + 1).post({ it > 0 }) { "positive" }
+}
+
+fun List<Int>.example() = map { increment(it) }
+```
+
+This will not work, as the analysis cannot guarantee that the value passed to `increment` inside `map` is indeed positive. Alas, there is no way to define a pre-condition which talks about _all the elements_ in the list (technically, that would involve quantifiers, and this is not supported by the logic in Λrrow Analysis.) The solution is to use the corresponding wrapper type:
+
+```kotlin
+fun List<Positive>.example() = map { increment(it.value) }
```
nomisRev, line 70: Nice!

nomisRev (approved):
LGTM! Great doc 👏 👏
I like the removal of `Arrow` in the features 😁

docs/docs/_data/analysis/sidebar.yml
```diff
@@ -14,6 +14,9 @@ options:
     - title: Types and invariants
       url: /analysis/types
 
+    - title: Wrappers
```
nomisRev, line 17: Could it make sense to call this `Types` or `Classes`?
Since we also have "Fields". "Wrappers" is often related to a negative connotatiion.

PR #963, opened 2021-12-15
Fixing permalinks for Arrow Analysis

franciscodr:
Jekyll is able to serve files when the URI doesn't provide the suffix `.html` but AWS S3 doesn't seem to support that functionality so the site for Arrow Analysis is broken.

This pull request fixes that issue by doing a couple of changes:
#### It changes the tree structure of the documentation, moving from:
```
analysis-quickstart.md
analysis-conditions.md
analysis-control.md
analysis-fields.md
analysis-java.md
analysis-laws.md
analysis-mutability.md
analysis-sarif.md
analysis-types.md
```
to
```
analysis/README.md
analysis/conditions/README.md
analysis/control/README.md
analysis/fields/README.md
analysis/java/README.md
analysis/laws/README.md
analysis/mutability/README.md
analysis/sarif/README.md
analysis/types/README.md
```
This change will generate an `index.html` file in each folder which will allow S3 to serve those files when it receives a request like `GET /analysis/conditions`

#### Additionally, we need to add an ending slash to the permalink to let Jekyll know it has to create the folder with the `index.html` file inside


raulraja (approved):
thanks @franciscodr !

PR #967, opened 2021-12-22
[Analysis] Better analysis for empty lists

serras:
This fixes the problem with the following snippet:

```kotlin
fun <A> List<A>.allEqual = all { it == first() }
```

This would flag the call to `first()` as problematic, since we need to have a non-empty list. However, if the list if empty, `all` will never call the function, so we are good!

This is done by adding a new kind of annotation to explain in which cases the arguments are not evaluated, based on some property of the receiver:

```kotlin
@Law
inline fun <E> List<E>.allLaw(predicate: (x: E) -> Boolean): Boolean {
  doNotLookAtArgumentsWhen(isEmpty()) { "empty lists have no elements" }
  return all(predicate)
}
```

### DIFF
--- a/gradle/projects.libs.versions.toml
+++ b/gradle/projects.libs.versions.toml
@@ -8,6 +8,7 @@ dokka = "1.6.0"
 intellijOpenApi = "7.0.3"
 javaAssist = "3.28.0-GA"
 junit = "5.8.2"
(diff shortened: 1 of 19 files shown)


raulraja (approved):
thanks @serras !

PR #968, opened 2022-01-10
[Analysis] Empty mapOf and isEmpty laws

bloderxd:
This PR implements 2 laws to the internals, that explicits axioms to `mapOf<K, V>` and `list.isEmpty()`. Before this PR pre conditions like this:

```kotlin
fun <E> List<E>.x(): Int {
  pre(isEmpty()) { "list is not empty" }
  return 1
}
val l = listOf<Int>().x()
```

and

```kotlin
fun <K, V> Map<K, V>.x(): Int {
  pre(isEmpty()) { "map is not empty" }
  return 1
}
val l = mapOf<Int, Int>().x()
```

breaks because there're no axioms for them.


serras (approved):
Thanks for adding these! I still find the way the compiler handles the `vararg` with the empty case a bit weird, but it's good to see that we can just add both cases and it works :)

PR #969, opened 2022-01-12
Remove Proofs plug-in and quote system

serras:
This PR removes the Proofs plug-in, which is now deprecated. Since it was the only user of the quote system in Arrow Meta, this has been removed too. The documentation has been updated to reflect this new situation; the old files haven't been removed since they may be linked from somewhere else, but there's now a warning about being deprecated.

As part of this PR we have also a few changes to the Meta site, to better showcase what can be done and link to the API docs correctly.


raulraja (approved):

raulraja:
We can probably remove the kastree internal impl the quotes system was using in Transform.NewSources and Transform.Replace

serras:
> We can probably remove the kastree internal impl the quotes system was using in Transform.NewSources and Transform.Replace

Done!

PR #990, opened 2022-01-26
[Analysis] Handle finally

serras:


plugins/analysis/common/src/main/kotlin/arrow/meta/plugins/analysis/phases/analysis/solver/check/Expressions.kt
```diff
@@ -292,12 +296,26 @@ private fun SolverState.checkBlockExpression(
...
+        }
+      }
+    }
     expressions.size == 1 &&
       implicitReturnFromLast -> // this is the last element, so it's the return value of the
       // expression
       checkExpressionConstraints(associatedVarName, expressions[0], data)
     else -> {
-      val first = expressions[0]
+      val first = expressions.first()
```
nomisRev, line 318: Not sure why you changed this, but I've started preferring `first` over `[0]` so I like this change 😁 
I rarely ever find myself using indexed access, `first`, `tail`, `last`, `map`, etc cover all my use-cases.

nomisRev (approved):

plugins/analysis/common/src/main/kotlin/arrow/meta/plugins/analysis/phases/analysis/solver/check/Expressions.kt
serras: I didn't think that much, I just say `[0]` and thought "ugly!". I guess I'm slowly internalizing good Kotlin idioms :)

PR #992, opened 2022-01-26
[Analysis] Docs on disabling checks

serras:


docs/docs/analysis/conditions/README.md
```diff
@@ -57,17 +57,26 @@ e: `wat` has inconsistent pre-conditions: (x < 0), (x > 0)
...
 
-val wrong = unsafeCall(increment(-2))
-```
+val wrong1 = unsafeCall(increment(-2))
+val wrong2 = unsafeBlock { increment(increment(-2)) }
 
-This case usually arises for data which comes from external input. However, we still encourage you to handle the possibility of failure, instead of blindly disabling the checks. An exception may be hiding behind that `unsafeCall`...
+@Suppress("UnsatCallPre")
+fun wrong3(): Int = increment(emptyList().first())
+```
```
nomisRev, line 79: Not a blocker right now, but I think having the examples shown under the explenation might improve clarity.
And using the same example ever time, to show the different behavior.

If I understood correctly.

```kotlin
val wrong = unsafeCall(increment(increment(-2)))
                                        ^ ignored
                                                            ^ failed to satisfy -2 => 0
```

```kotlin
val wrong = unsafeBlock  { increment(increment(-2)) }
                                          ^ ignored
```

```kotlin
fun wrong3() {
  increment(increment(-2))
  ...
}
// Whole fun body is ignored
// @Supress("UnsatCallPre") <~> `fun wrong3() = unsafeBlock { }`
```

nomisRev (approved):

docs/docs/analysis/conditions/README.md
serras: I've copy-pasted your examples in the docs, they make thing much clearer

PR #993, opened 2022-01-27
[Analysis] Missing nodes for Kotlin plug-in

serras:
This should cover all the remaining nodes in Kotlin's AST.


raulraja (approved):
Thanks @serras !

PR #1004, opened 2022-02-04
Fix interpretation for throw statement

gimlet2:
I've tried to add arrow meta analyzes with my project which contained `throw <some exception>` code. And it was failing.
Please take a look on this PR. Is it right way to fix it?


serras:
@gimlet2 thanks for your help! Unfortunately this implementation is not right, `throw` is allowed by the common part of Arrow Analysis, so the wrapper element needs to subclass `ThrowExpression`. I've added a fix for this as part of #1005.

gimlet2:
Thank you @serras :)

PR #1007, opened 2022-02-09
[Analysis] Better support for vararg

### LINKED ISSUE #997: [BUG] Documentation Analysis Wrappers
**Describe the bug**
Section **[No higher-order support](https://arrow-kt.io/docs/meta/analysis/control/#no-higher-order-support)** states:

>  ... As a result, the following is accepted, since Λrrow Analysis tracks the length of the list through map, and can see that the call to first() is correct.

```kotlin
val ok = listOf(-1).map { absoluteValue(it) }.first()
```

I'm assuming the `absoluteValue` function is the one referenced from the section above

```kotlin
import arrow.analysis.post

fun absoluteValue(n: Int): Int = when {
  n < 0  -> -n
  n == 0 -> 0
  else   -> n
}.post({ it >= 0 }) { "result >= 0" }
``` 

Trying to compile this however results in:

```
e: .../example.kt: (55, 10): pre-condition `not empty` is not satisfied in `first()`
  -> unsatisfiable constraint: `(listOf(-1).map { absoluteValue(it) }.size) >= 1`
  -> <no local variable involved>
  -> in branch: 1 != null, listOf(-1) != null, listOf(-1).isEmpty, listOf(-1).map { absoluteValue(it) } != null
```

serras:
Fixes #997


raulraja (approved):
thanks @serras !

PR #1033, opened 2022-03-16
Update 'java-smt' to 3.12

serras:
Unfortunately, there's a bug in SMTInterpol in `java-smt` 3.12. This PR updates the dependency and adds works around the problem by catching the thrown exception and doing the right thing. This patch is forwards compatible, so even when they fix the bug the code can remain the same.

Closes #1027


plugins/analysis/common/src/main/kotlin/arrow/meta/plugins/analysis/smt/FormulaExtensions.kt
```diff
@@ -106,3 +123,39 @@ fun Solver.isFieldCall(f: Formula): Boolean {
...
+          val newFn =
+            FunctionDeclarationImpl.of(
+              fn.name,
+              fn.kind,
+              fn.argumentTypes.take(2),
+              fn.type,
+              fn.solverDeclaration
+            )
+          // this happens sometimes in SMTInterpol,
+          // so we try to fix it by using the function with each 2 arguments
```
raulraja, line 157: 👍 Thanks @serras !

raulraja (approved):

PR #1037, opened 2022-03-18
remove plugins

i-walker:
- proofs plugins is hosted here https://github.com/arrow-kt/arrow-proofs
- arrow-analysis plugin here https://github.com/arrow-kt/arrow-analysis

the proofs plugin is still under development


i-walker:
I've temporarily disabled publicaion, so mergig this doesn't affect upstream binaries

raulraja (approved):

raulraja:
Thanks @i-walker ! can we get Meta also configured with the gradle config plugin and ready to produce alphas? We are gonna need meta for changes around FIR in the proofs plugin.

PR #1038, opened 2022-03-20
Setup Arrow-Meta with arrow-gradle-config

i-walker:
- automatic semantic versioning
- binaryCompability tooling
- api files are being updated for each PR
- generate tag and generate alpha tag github workflows
- publishing alpha/beta/rc tags and final too
This PR will produce a 1.6.1-alpha.1 release. Since I pushed a 1.6.0 tag. Which will be cleaned up after


raulraja (approved):
Thanks @i-walker !, I'm also tagging @JavierSegoviaCordoba, he has more experience with this setup more than I do.

.github/workflows/generate-alpha-tag.yaml
```diff
@@ -0,0 +1,59 @@
+name: generate-alpha-tag
+
+on:
+  push:
+    branches:
+      - main
+
+env:
+  GRADLE_OPTS: -Dorg.gradle.daemon=false -Dorg.gradle.parallel=false -Dorg.gradle.jvmargs="-Xmx5g -XX:+UseParallelGC -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8"
```
JavierSegoviaCordoba, line 9: meta needs no daemon and no parallel to work too?

.gitignore
```diff
@@ -16,6 +16,8 @@ bin/
 gen/
 out/
 
+kotlin-js-store
```
JavierSegoviaCordoba, line 19: I think this should be in the repo and not ignored based on comments in Slack from JB engineers.

JavierSegoviaCordoba (approved):
Good job :) A few I leave a few doubts

.gitignore
i-walker: I am also in favor. Adding the kotlin-js-store hasn't been added consistently to all arrow projects. 
Maybe it makes sense to decide, whether we should have it in all projects or not
cc @raulraja , @nomisRev

nomisRev: > based on comments in Slack from JB engineers.

If it's officially recommended to keep this file in git then I think we should.

i-walker: Will add it :D and do the same for the proofs and analysis plugin

.github/workflows/generate-alpha-tag.yaml
i-walker: as in our chat, we could take a look if we can improve this in another PR

i-walker: as in our chat, we could take a look if we can improve this in another PR

PR #1048, opened 2022-04-05
Update org.jetbrains.kotlin to v1.6.20

JavierSegoviaCordoba:
| datasource | package                                   | from  | to     |
| ---------- | ----------------------------------------- | ----- | ------ |
| maven      | org.jetbrains.kotlin:kotlin-stdlib-jdk8   | 1.6.0 | 1.6.20 |
| maven      | org.jetbrains.kotlin:kotlin-stdlib-common | 1.6.0 | 1.6.20 |


raulraja (approved):

libs/arrow-meta/src/main/kotlin/arrow/meta/phases/analysis/diagnostic/PositionStrategies.kt
```diff
@@ -1,58 +0,0 @@
...
-import org.jetbrains.kotlin.diagnostics.PositioningStrategy
-import org.jetbrains.kotlin.diagnostics.hasSyntaxErrors
-import org.jetbrains.kotlin.diagnostics.markElement
-import org.jetbrains.kotlin.psi.KtAnnotationEntry
-import org.jetbrains.kotlin.psi.KtDeclaration
-import org.jetbrains.kotlin.psi.KtObjectDeclaration
-import org.jetbrains.kotlin.psi.psiUtil.textRangeWithoutComments
-import org.jetbrains.kotlin.resolve.BindingContext
-
-/**
```
raulraja, line 16: @i-walker we did not find references to any of the functions on this file. Tagging you here to verify we can get rid of this and this is not used somewhere else.

i-walker: Yes that is fine, these were some left overs from the ide implementation of proofs :)

PR #1071, opened 2023-04-29
Update to 1.8.21

serras:


libs/arrow-meta/src/main/kotlin/arrow/meta/dsl/analysis/AnalysisSyntax.kt
```diff
@@ -26,8 +26,8 @@ import org.jetbrains.kotlin.resolve.diagnostics.MutableDiagnosticsWithSuppressio
 /**
  * The Analysis phase determines if the parsed AST type checks and resolves properly. As part of
  * this phase, we have access to events happening before and after resolution. Before resolution, we
- * are given the chance to modify the compiler trees in the form of `KtElement` via the [Quote
```
raulraja, line 29: There is no longer a quote system, we removed it a while back.

raulraja (approved):
thanks @serras !

PR #1075, opened 2023-09-26
[WIP] Experiment - Update to Kotlin 1.9.20-Beta2

apatrida:
This PR is not meant to be merged, it is meant to be informational to see what the current state is for Arrow-Meta and latest Kotlin compilers.

Experiment to see the difference for upgrading to current Kotlin compiler from 1.9.20-Beta2.

Notes:

* Kotlin-Compiler-Testing will fail tests, but there is a branch that is close to working (https://github.com/tschuchortdev/kotlin-compile-testing/pull/388) and with one additional change to remove `useIR` setting in `KotlinCompilation.t` (`// args.useIR = useIR`) this branch works and tests pass again.

* `ClassBuilder` uses extension point that is K1 only, and not K2 compatible.  I marked it deprecated to leave the code, but it should be removed.  `ClassGeneration` is a new extension point.  `ClassBuilderInterceptorExtension` is the problematic class, which now has this deprecation eeror:  
	```
	@Deprecated(
	    "This extension is only supported in K1 and will not work properly in K2. " +
	            "Please migrate to org.jetbrains.kotlin.backend.common.extensions.IrGenerationExtension or " +
	            "org.jetbrains.kotlin.backend.jvm.extensions.ClassGeneratorExtension.",
	    level = DeprecationLevel.ERROR
	)
	```

* `IrUtils` becomes pretty suspect after this change since all of the utility functions based on descriptors access things like `pluginContext.symbols.externalSymbolTable.referenceSomething(this)` which moved one level deeper to ` pluginContext.symbols.externalSymbolTable.descriptorExtension.referenceSomething(this)` and produce deprecation warnings "Please use IR declaration properties and not its descriptor properties" which is referring to `IdSignature` instances which can be used in the original call `pluginContext.symbols.externalSymbolTable.referenceSomething(idSignature)`.  More info here:

	```
	/**
	 * This annotation is used on IR API elements which use front-end (FE) descriptors to obtain information, directly or indirectly.
	 *
	 * Descriptors are used in FE to represent main declaration properties and to refer declarations.
	 * Early IR versions were descriptor-based, so IR elements used descriptors to obtain some information about element properties.
	 * However, more correct and universal way is to store all necessary information inside IR elements themselves
	 * and do not use descriptors as some intermediate storage. It's planned to remove all descriptor usages from IR in future.
	 */
	@Target(
	    AnnotationTarget.CLASS,
	    AnnotationTarget.PROPERTY,
	    AnnotationTarget.VALUE_PARAMETER,
	    AnnotationTarget.FUNCTION,
	    AnnotationTarget.TYPEALIAS
	)
	@RequiresOptIn(message = "Please use IR declaration properties and not its descriptor properties", level = RequiresOptIn.Level.ERROR)
	annotation class ObsoleteDescriptorBasedAPI
	```
	and https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/IdSignature.kt
	
With this, all tests pass, but there is some ominous warnings there about `IrUtils` and the descriptors they were based upon.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -13,11 +13,6 @@ plugins {
 }
 
 allprojects {
-  repositories {
-    mavenCentral()
-    maven(url = "https://oss.sonatype.org/content/repositories/snapshots/")
-  }
-
   group = property("projects.group").toString()
 }
 
--- a/gradle/projects.libs.versions.toml
+++ b/gradle/projects.libs.versions.toml
@@ -7,12 +7,12 @@ dokka = "1.8.10"
 intellijOpenApi = "7.0.3"
 javaAssist = "3.29.2-GA"
 junit = "5.9.2"
-kotlin = "1.8.21"
-kotlinCompileTesting = "1.5.0"
+kotlin = "1.9.20-Beta2"
+kotlinCompileTesting = "1.5.1-SNAPSHOT"
 javaCompileTesting = "0.21.0"
 kotlinBinaryCompatibilityValidator = "0.13.0"
 detekt = "1.22.0"
-ksp = "1.8.21-1.0.11"
+ksp = "1.9.20-Beta2-1.0.13"
 
 [libraries]
 arrowCore = { module = "io.arrow-kt:arrow-core", version.ref = "arrow" }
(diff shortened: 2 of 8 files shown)


apatrida:
Quick test with this branch updating to Kotlin `1.9.30-dev-3330` (latest referenced from intellij-community K2 runtime configuration builds) shows this PR to work for that version as  well.

serras:
@apatrida thanks for your work! We would actually love to merge this, so we can keep Arrow Meta (and Arrow Analysis) working on the latest Kotlin versions. The `IrUtils` could maybe we reworked to use "IR declaration properties" (whatever they are) and keep this warning-free.

serras:
@apatrida would it be possible to allow pushes to your branch? There are a few changes I would like to add.

PR #1077, opened 2023-12-11
Upgrade to Kotlin 1.9.21

travisMiehm:
The PR upgrades the project's Kotlin version to 1.9.21 - the latest as of this change.

This change is heavily based off, #1075 , which seems to have been abandoned by the author. That PR had comments about some changes that seemed desirable, but I did not quite follow them. I am more than happy to push any additional changes, or to open the source branch to additional contributors upon request.

Most of the change comes from migrating from https://github.com/tschuchortdev/kotlin-compile-testing to https://github.com/ZacSweers/kotlin-compile-testing, as the former does not have an updated version for Kotlin 1.9.21. In the new dependency, `Result` was renamed `CompilationResult` and marked `@ExperimentalCompilerApi`.

Beyond that, various `@OptIn(ObsoleteDescriptorBasedAPI::class)` usages were added due to changes pulled in by Kotlin 1.9.21.


travisMiehm:
> Are all of those `ObsoleteDescriptorBasedAPI` necessary? I think they will be removed so we should avoid them.

There is probably an alternative, but I had trouble finding it. I am mostly kicking the can down the road here. If that's a blocker, I will close this and will need to revisit it later.

JavierSegoviaCordoba:
From my side, there is no problem if this is already working as it can be changed in a different PR or even when a new Kotlin version forces us to change it.

cc @raulraja @serras

Atternatt:
is there any update on this? Would be really nice to have support over `1.9.21`

